PR #20, opened 2024-08-01
feat(market): add liquidity to utilization getters

Rubilmax:


packages/blue-sdk/src/maths/AdaptiveCurveIrmLib.ts
```diff
@@ -145,4 +145,37 @@ export namespace AdaptiveCurveIrmLib {
...
+
+    const minBorrowRate = MathLib.wDivDown(rateAtTarget, CURVE_STEEPNESS);
+
+    return MathLib.max(
+      0n,
+      MathLib.mulDivDown(
+        TARGET_UTILIZATION,
+        borrowRate - minBorrowRate,
+        rateAtTarget - minBorrowRate,
+      ),
```
oumar-fall, line 178: How can this be negative? can either `borrowRate` or `rateAtTarget` be smaller than `minBorrowRate` ?

oumar-fall (approved):
I don't know the IRM enough to tell that the computations are exact but It looks good overall

packages/blue-sdk/src/maths/AdaptiveCurveIrmLib.ts
Rubilmax: `borrowRate` can yes, because it's a user input!

PR #21, opened 2024-08-02
fix(market): fix supply to utilization edge cases

Rubilmax:


packages/blue-sdk/src/market/MarketUtils.ts
```diff
@@ -137,26 +137,22 @@ export namespace MarketUtils {
...
       totalSupplyAssets: BigIntish;
       totalBorrowAssets: BigIntish;
     },
     utilization: BigIntish,
   ) {
     utilization = BigInt(utilization);
-    totalBorrowAssets = BigInt(totalBorrowAssets);
     if (utilization === 0n) {
-      if (totalBorrowAssets === 0n) return totalSupplyAssets;
+      if (getUtilization(market) === 0n) return 0n;
```
oumar-fall, line 148: This case depends a lot on the business case no?

oumar-fall (dismissed):

packages/blue-sdk/src/market/MarketUtils.ts
Rubilmax: Yes, so I updated jsdocs so it's clearer thanks

PR #30, opened 2024-08-07
[1.5.6] Feat/add migration addresses

oumar-fall:


packages/morpho-ts/src/time/time.ts
```diff
@@ -5,6 +5,10 @@ const toNumberish = (returnValue: bigint, initialValue: bigint | number) => {
 
 const UNITS = ["ms", "s", "min", "h", "d", "w", "mo", "y"] as const;
 export type TUnit = (typeof UNITS)[number];
+export type TPeriod = {
+  unit: TUnit;
+  duration: number;
+};
```
Rubilmax, line 11: imo we should either export all from `Time` and nowhere else or export all here but no type from `Time`

At first I would have said everything here but `Unit` is such a broad naming that I indeed prefer exporting only from `Time`

packages/blue-sdk/src/maths/MathLib.ts
```diff
@@ -190,8 +190,9 @@ export class MathLib {
...
    * @param compounding The compounding basis
    */
-  static rateToApy(rate: BigIntish, unit: Time.Unit) {
-    const factor = Time[unit].from.y(1n);
+  static rateToApy(rate: BigIntish, period: Time.PeriodLike) {
+    const { unit, duration } = Time.toPeriod(period);
+    const factor = Time[unit].from.y(1n) / BigInt(duration);
 
     return (
       (1 + Number(format.number.of(BigInt(rate), 18))) ** Number(factor) - 1
```
Rubilmax, line 198:
```suggestion
    const factor = Time[unit].from.y(1) / duration;

    return (
      (1 + Number(format.number.of(BigInt(rate), 18))) ** factor - 1
```

Allows to have more precise calculations

Rubilmax (changes requested):

packages/blue-sdk/src/maths/MathLib.ts
```diff
@@ -204,8 +203,9 @@ export class MathLib {
    * @param apr The yearly apr to convert (in WAD)
    * @param compounding The compounding basis
    */
-  static aprToApy(apr: BigIntish, compounding: Time.Unit) {
-    const factor = Time[compounding].from.y(1n);
+  static aprToApy(apr: BigIntish, compounding: Time.PeriodLike) {
+    const { unit, duration } = Time.toPeriod(compounding);
+    const factor = Time[unit].from.y(1n) / BigInt(duration);
 
     const rate = BigInt(apr) / factor;
```
Rubilmax, line 210:
```suggestion
    const rate = BigInt(apr) * BigInt(duration) / Time[unit].from.y(1n);
```

packages/blue-sdk/src/maths/MathLib.ts
```diff
@@ -204,8 +203,9 @@ export class MathLib {
    * @param apr The yearly apr to convert (in WAD)
```
Rubilmax, line 203: "yearly APR" is redundant: APR = Annual Percentage Rate

PR #47, opened 2024-08-20
feat(position): add position functions

Rubilmax:


packages/blue-sdk/src/position/Position.ts
```diff
@@ -167,6 +168,95 @@ export class AccrualPosition extends Position implements InputAccrualPosition {
...
+  ) {
+    const market = this.market.accrueInterest(timestamp);
+
+    this.collateral -= assets;
+
+    if (this.collateral < 0n)
+      throw new BlueErrors.InsufficientPosition(this.user, this.marketId);
+
+    if (!market.isHealthy(this))
+      throw new BlueErrors.InsufficientCollateral(this.user, this.marketId);
```
oumar-fall, line 223: I don't have way better but this can be confusing (the first one could also be seen as "insufficientCollateral")

packages/blue-sdk/src/position/Position.ts
```diff
@@ -167,6 +168,95 @@ export class AccrualPosition extends Position implements InputAccrualPosition {
...
+
+  public repay(assets: bigint, shares: bigint, timestamp?: bigint) {
+    let { market } = this;
+    ({ market, assets, shares } = market.repay(assets, shares, timestamp));
+
+    this.borrowShares -= shares;
+
+    if (this.borrowShares < 0n)
+      throw new BlueErrors.InsufficientPosition(this.user, this.marketId);
+
```
oumar-fall, line 252: shouldn't we technically check for health? (same for supply collat) ?

oumar-fall (approved):

packages/blue-sdk/src/position/Position.ts
Rubilmax: No because it's not done onchain and the reason is it can only increase the position's health!

packages/blue-sdk/src/position/Position.ts
Rubilmax: Indeed, but I don't have other suggestions neither 😅

packages/blue-sdk/src/position/Position.ts
oumar-fall: indeed, Idk why but I thought that the check `require(isHealthy())` was made after each operation 🤔

PR #62, opened 2024-09-06
fix(market): always create instance when market accrues interest

Rubilmax:


packages/blue-sdk/src/market/Market.ts
```diff
@@ -200,7 +200,7 @@ export class Market implements InputMarket {
         this.lastUpdate,
       );
 
-    if (elapsed === 0n) return this;
+    if (elapsed === 0n) return new Market(this);
```
oumar-fall, line 203: Why do we wanna re-create if no interests were accrued?

packages/blue-sdk/src/market/Market.ts
```diff
@@ -273,11 +265,7 @@ export class Market implements InputMarket {
     return { market, assets, shares };
   }
 
-  public borrow(
-    assets: bigint,
-    shares: bigint,
-    timestamp: bigint = this.lastUpdate,
-  ) {
+  public borrow(assets: bigint, shares: bigint, timestamp: BigIntish) {
```
Rubilmax, line 268: Remove default value which is misleading: it could be the last market's update, the current timestamp, based on the use case. Prefer to not give defaults and force the dev to provide the value, to think on the use case

PR #69, opened 2024-09-10
Add simulations & wagmi hooks

Rubilmax:
Requests are faster because each of them is now independent (no need to fetch the config before fetching the rest of each market/vault). But deployless reads are not multicall-able, meaning they can't be batched in a single onchain RPC call via multicall.

This state is arguable and does not seems better out-of-the-box. However, here is why it is better:
- requests can be batched with most RPC backends nowadays
- requests are already actually batched using HTTP/2 natively via the browser (so there's no actual "delay" due to HTTP overhead)

So the only matter is the request count for RPC backends, which will be the cost paid by the RPC user in the end. It should stay relatively low considering that now, each entity costs at most 1 RPC request (either through the multicall = 0, or through a single deployless read = 1)

Fixes INTER-1016


packages/blue-sdk-viem/contracts/interfaces/IMorpho.sol
julien-devatom, line 1: can't we just use the npm morpho blue package that is importing the contracts?

julien-devatom (commented):
can't we let the user choose the method to use? either with multicall or with deployless

packages/blue-sdk-viem/contracts/interfaces/IMorpho.sol
Rubilmax: For some reason I can't make it to work

julien-devatom (dismissed):
Since we already know that deployless fails for mainnet weth, wdyt about always skipping deployless with a set of address? It can prevent known rpc fails 

Else everything looks good

packages/blue-sdk-viem/src/fetch/Market.ts
```diff
@@ -1,49 +1,93 @@
...
   id: MarketId,
   client: Client,
   {
     chainId,
     overrides = {},
-  }: { chainId?: ChainId; overrides?: ViewOverrides } = {},
+    deployless = true,
+  }: FetchOptions & {
+    deployless?: boolean;
+  } = {},
```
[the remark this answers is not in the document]
Rubilmax: I will in the PR on top!

packages/blue-sdk-ethers/src/fetch/Holding.ts
```diff
@@ -126,7 +122,7 @@ export async function fetchHolding(
         runner,
       )
         .isWhitelisted(user, overrides)
-        .catch(() => undefined);
+        .catch(() => false);
```
oumar-fall, line 125: We don't want to allow everyone by default in case of rpc call fail?

packages/blue-sdk-viem/src/fetch/Market.ts
```diff
@@ -1,49 +1,93 @@
...
+        totalSupplyShares,
+        totalBorrowShares,
+        lastUpdate,
+        fee,
+        price,
+        rateAtTarget:
+          marketParams.irm === adaptiveCurveIrm ? rateAtTarget : undefined,
+      });
+    } catch {
+      // Fallback to multicall if deployless call fails.
```
oumar-fall, line 69:
```suggestion
      // TODO Fallback to multicall if deployless call fails.
```

oumar-fall (dismissed):

packages/blue-sdk-ethers/src/fetch/Holding.ts
Rubilmax: Yeah that's what we wanted. And in this case we have no way of knowing the call failed... But at the same time it seems unsafe to me to behave differently than when the call returns `false` if the call fails because of the call fails offchain, it will fail onchain so it won't work. Except if the spec changes without upfront warning - which happened before and was the reason we introduced this change

So I'll introduce undefined back again

packages/blue-sdk-viem/src/fetch/Market.ts
Rubilmax: No! The fallback is just below

julien-devatom (commented):
I don't have reviewed logic yet but few remarks from your Thursday presentation:

blue-sdk-viem-simulation should not depend on react imo. Ex if we want to use it in liquidation SDK. A solution would be to have a dedicated package for the hook useSimulationState (which is ok I think)

- I think it's not necessary to have "viem" in the SDK name for a non primitive SDK like blue-sdk. underthehood, as soon as you want to use simulation, you use viem.

PR #71, opened 2024-09-11
feat(blue-sdk): add wagmi-like hooks

Rubilmax:


packages/blue-sdk-wagmi/src/hooks/useChainId.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type UseChainIdParameters<config extends Config = Config> =
+  ChainIdParameter<config> & ConfigParameter<config>;
+
+export function useChainId<config extends Config = ResolvedRegister["config"]>(
+  parameters: UseChainIdParameters<config>,
+) {
+  const wagmiChainId = wagmi_useChainId(parameters);
+
+  return ChainUtils.parseSupportedChainId(parameters.chainId ?? wagmiChainId);
```
oumar-fall, line 18: Is it convenient to have react hooks throw errors?

oumar-fall (approved):
looks promising! I think some comments might arise with usage

packages/blue-sdk-wagmi/src/hooks/useChainId.ts
Rubilmax: Nice catch, I'll try to find a workaround

Rubilmax: I actually think it's fine to break everything if wagmi is connected to a chain id not accepted by the sdk

PR #72, opened 2024-09-16
feat(blue-sdk): add marketCapacities options

oumar-fall:
for [this issue](https://linear.app/morpho-labs/issue/INTER-1406/new-rules-for-max-for-both-borrow-and-withdraw-collateral-other), we need to be able to get maximum borrowable/withdrawable amount up to a specific ltv.

### DIFF
--- a/packages/blue-sdk/src/market/Market.ts
+++ b/packages/blue-sdk/src/market/Market.ts
@@ -18,9 +18,11 @@ export interface CapacityLimit {
   limiter: CapacityLimitReason;
 }
 
-export interface MaxCapacitiesOptions {
-  borrow?: { maxLtv?: bigint };
-  withdrawCollateral?: { maxLtv?: bigint };
+export interface MaxBorrowOptions {
+  maxLtv?: bigint;
+}
+export interface MaxWithdrawCollateralOptions {
+  maxLtv?: bigint;
 }
 
 export interface MaxPositionCapacities {
@@ -385,10 +387,10 @@ export class Market implements InputMarket {
    */
   public getMaxBorrowAssets(
     collateral: bigint,
-    options: MaxCapacitiesOptions["borrow"] = {},
+    { maxLtv = this.config.lltv }: MaxBorrowOptions = {},
   ) {
     return MarketUtils.getMaxBorrowAssets(collateral, this, {
-      lltv: options.maxLtv ?? this.config.lltv,
+      lltv: MathLib.min(maxLtv, this.config.lltv),
     });
   }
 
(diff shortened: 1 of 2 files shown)


packages/blue-sdk/src/market/Market.ts
```diff
@@ -378,8 +383,13 @@ export class Market implements InputMarket {
    * Returns the maximum amount of loan assets that can be borrowed given a certain amount of collateral.
    * @param collateral The amount of collateral to consider.
    */
-  public getMaxBorrowAssets(collateral: bigint) {
-    return MarketUtils.getMaxBorrowAssets(collateral, this, this.config);
+  public getMaxBorrowAssets(
+    collateral: bigint,
+    options: MaxCapacitiesOptions["borrow"] = {},
```
Rubilmax, line 388:
```suggestion
    { lltv = this.config.lltv }: { lltv?: bigint } = {},
```

Rubilmax: I prefer keeping the naming `lltv` because `maxLtv` may suggest that the resulting ltv may be lower than the given `maxLtv`

packages/blue-sdk/src/market/Market.ts
```diff
@@ -18,6 +18,11 @@ export interface CapacityLimit {
   limiter: CapacityLimitReason;
 }
 
+export interface MaxCapacitiesOptions {
+  borrow?: { maxLtv?: bigint };
+  withdrawCollateral?: { maxLtv?: bigint };
+}
```
Rubilmax, line 24: If defined, this should be defined at the `MarketUtils` level imo (because that's where the function is actually implemented) and in this case it is more semantically equal to `Partial<Pick<MarketConfig, "lltv">>`

packages/blue-sdk/src/market/Market.ts
oumar-fall: This isn't technically a lltv 🤔  so this naming would be wrong

The "max" here refers to the "max" in the name of the fonction: if the max ltv is X, you can borrow up to Y

Rubilmax: Ok then `ltv`

Rubilmax: It is technically a lltv (liquidation ltv): max borrow assets is only defined given an amount of collateral and a liquidation ltv

If given an amount of collateral and a ltv, then a better naming is "borrow assets" and not "max borrow assets" (or "borrowable" or whatever, but it's not a maximum anymore)

oumar-fall: Well it is the maximum borrowable assets given a maximum ltv 🤔 I don't get your point

Rubilmax: max ltv == liquidation ltv, but max ltv is not a naming used in the ecosystem

Rubilmax: In the end I'm ok with `maxLtv`

Rubilmax (approved):

PR #99, opened 2024-09-26
Feat/add viem tx utils

oumar-fall:


packages/blue-sdk/src/types.ts
```diff
@@ -1,14 +1,15 @@
-export type Hex64 = string & { __LENGTH__: 64 };
+export type Hex<L extends number = -1> = `0x${string}` &
+  (L extends -1 ? unknown : { __LENGTH__: L });
 
 /**
  * The address of a Contract, or an EOA
  */
-export type Address = string;
+export type Address = Hex;
```
Rubilmax, line 7:
```suggestion
export type Address = `0x{string}`;
```

A reviewer could be tempted to move to `Hex<20>` otherwise (semantically Hex<20> = Address < Hex)

Rubilmax: IMO we should get rid of the `Hex`

packages/blue-sdk/src/types.ts
```diff
@@ -1,14 +1,15 @@
...
 
 /**
  * The id of a market used on the Blue contract
  */
-export type MarketId = `0x${Hex64}` & { __TYPE__: "marketId" };
+export type MarketId = `0x${Hex<64>}` & { __TYPE__: "marketId" };
```
Rubilmax, line 12:
```suggestion
export type MarketId = `0x${string}` & { __TYPE__: "marketId" };
```

packages/blue-sdk-viem/src/signatures/index.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export * from "./types";
+
+export namespace SignatureUtils {
+  export const Permit = _Permit;
+  export const Permit2 = _Permit2;
+  export const Manager = _Manager;
+
+  export const verifySignature = _verifySignature;
+}
```
Rubilmax, line 14: Is this namespace really interesting? I feel like this is similar to ethers v5 design where you had to import blocks of code like `utils` and then you would use `utils.formatUnits` etc

I understand that here the dev has a choice to either use `SignatureUtils` or direct exports though. But it is really interesting to use a namespace

Rubilmax (commented):
Agreed on the type behavior

packages/blue-sdk-viem/src/signatures/index.ts
oumar-fall: I didn't think about this, I copied what was done in blue-sdk-ethers

packages/blue-api-sdk/src/types.ts
```diff
@@ -26,7 +26,7 @@ export type Scalars = {
   Boolean: { input: boolean; output: boolean };
   Int: { input: number; output: number };
   Float: { input: number; output: number };
-  Address: { input: string; output: string };
+  Address: { input: string; output: Address };
   BigInt: { input: string | number; output: bigint };
   HexString: { input: string; output: string };
```
Rubilmax, line 31: `0x${string}`

packages/blue-sdk-viem/src/signatures/index.ts
```diff
@@ -0,0 +1,11 @@
...
+export * from "./permit";
+export * from "./permit2";
+export * from "./utils";
+export * from "./types";
+
+export * as manager from "./manager";
+export * as permit from "./permit";
+export * as permit2 from "./permit2";
+export * as utils from "./utils";
+export * as types from "./types";
```
Rubilmax, line 11: To me it's just like namespaces
But they have too common names (like `types` which may get overridden by main export from root src)

I may even prefer not to export like this at the root src but it's not that of a big deal if it's only done at the highest level

And this made me think about namespaces: they are good to restrict function and constant definitions to a specific domain, so we don't have to specify the domain in the function's name for example

but here we have `getManagerApprovalMessage` it is clear enough that the domain is manager approval and the function is getMessage
Namespaces would be useful as: `ManagerApproval.getMessage`

PR #136, opened 2024-10-21
refactor(token): merge price into Token

Rubilmax:
BREAKING CHANGE: TokenWithPrice has been merged into Token.

The price field has been moved to Token.
The name and symbol properties are now possibly undefined. The name no longer defaults to the symbol. The decimals property defaults to 0.


packages/blue-sdk/src/vault/Vault.ts
```diff
@@ -49,6 +49,16 @@ export interface InputVault extends InputVaultConfig {
 }
 
 export class Vault extends VaultToken implements InputVault {
+  /**
+   * The vault's share token's name.
+   */
+  public declare readonly name: string;
```
oumar-fall, line 55: Why is the declare useful?

oumar-fall (approved):

PR #152, opened 2024-11-04
Improve formatters customisation

oumar-fall:
2 things done here:
- Allow dev to create custom formatters via `createFormat` to avoid misusages
- Fix bug related to sibling formatters sharing configuration

eg: before this PR:

const formatter1 = format.number.digits(10);
const formatter2 = formatter1.digits(0); // this would also set `digits` to `0` in `formatter1`


Rubilmax (approved):
Looks good!

PR #191, opened 2024-12-03
fix(blue-api-sdk): update type policies

Rubilmax:


packages/blue-api-sdk/src/cache.ts
```diff
@@ -16,9 +26,86 @@ export const readMaybeBigInt = (value: string | number | undefined | null) => {
...
+      path: string[],
+    ): PropertyKey | null | undefined => {
+      if (path.length === 0) return data;
+
+      const [key, ...rest] = path;
+
+      return _get(readField(key!, data), rest);
+    };
+
+    const getId = (
```
oumar-fall, line 60: This function deserves a comment since it took me a bit to really understand what it does

oumar-fall: esp. since it doesn't necessarily returns you an id but the value stored at the first one of `paths` at which a value is stored.

oumar-fall: _This function iterates over a list of possible `paths` and retrieves the first value found in `data` at any of these paths, or `undefined` if none is found._

oumar-fall: a better naming would be `getFirstValueAtPath` or sth like that imo

oumar-fall (approved):

oumar-fall (approved):

PR #198, opened 2024-12-10
refactor(ethers-liquidation): export market position type

crisog:
Without this change, if you wanted to do something like this:
```ts
const {
  assetByAddress: { priceUsd: wethPriceUsd },
  marketPositions: { items: positions }
} = await apiSdk.getLiquidatablePositions({
  chainId: this.config.chainId,
  wNative,
  marketIds
});

if (wethPriceUsd == null) {
  return;
}

await Promise.all(
  positions.map((position) => {
    return processPosition(position, wethPriceUsd);
  })
);
```
And preserve the types, you had to create your own market position type or do tricks. 

```ts
async processPosition(
    position: MarketPosition, // Self-created Market position type
    wethPriceUsd: number
  ): Promise<PositionResult>
```
I believe it would be feasible to export the type right away to avoid this.

This is my first PR so I'm not familiar with the process of getting this merged. I built the package and used it locally which worked fine. This change should not be breaking because its the same types, just used/exported differently.

I tried running both `test-jest` and `test-hardhat` but for the latter I need an 1INCH API KEY.

### DIFF
--- a/packages/blue-sdk-ethers-liquidation/src/api/types.ts
+++ b/packages/blue-sdk-ethers-liquidation/src/api/types.ts
@@ -9,40 +9,37 @@ export type GetLiquidatablePositionsQueryVariables = Types.Exact<{
   first?: Types.InputMaybe<Types.Scalars["Int"]["input"]>;
 }>;
 
+type Asset = {
+  __typename?: "Asset";
+  address: Types.Scalars["Address"]["output"];
+  decimals: number;
+  symbol: string;
+  priceUsd: number | null;
+  spotPriceEth: number | null;
+};
+
+export type MarketPosition = {
+  __typename?: "MarketPosition";
+  user: {
+    __typename?: "User";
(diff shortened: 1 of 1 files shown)


Rubilmax:
Hi, thanks a lot for your contribution!

The thing is this file is codegenerated (automatically), so this would work temporarily until the next time it is codegen-ed again. I'll investigate if we can export them right away using some specific codegen config

crisog:
> Hi, thanks a lot for your contribution!
> 
> The thing is this file is codegenerated (automatically), so this would work temporarily until the next time it is codegen-ed again. I'll investigate if we can export them right away using some specific codegen config

Didn't realize! Thanks for the clarification 💯

Rubilmax:
Also we have a huge prerelease incoming on `next` if you want, it's still unstable but it has a lot more features than `main` and should be release in the coming month!

PR #200, opened 2024-12-13
Fix/simulation/permissionned transfers

oumar-fall:


Rubilmax (changes requested):
It's not the spec of `canTransfer` so far: it is true if a transfer of 1 unit of token does not revert and false otherwise, and false if known permissioned tokens do not allow transfer

For known permissioned tokens, this implementation is correct
But the broader spec is incompatible: it's not because `to` cannot transfer that `to` cannot receive

PR #210, opened 2024-12-23
feat(shared-liquidity): add default supply target utilization

Rubilmax:


packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -302,7 +310,8 @@ export const populateSubBundle = (
 
     const supplyTargetUtilization =
       publicAllocatorOptions.supplyTargetUtilization?.[market.params.id] ??
-      DEFAULT_SUPPLY_TARGET_UTILIZATION;
+      publicAllocatorOptions.defaultSupplyTargetUtilization ??
+      90_5000000000000000n;
```
oumar-fall, line 314: Why?

oumar-fall: I mean why not using it in a constant?

oumar-fall (dismissed):
Why have you removed the constants?

packages/blue-sdk/src/constants.ts
```diff
@@ -10,8 +10,6 @@ export const LIQUIDATION_CURSOR = 30_0000000000000000n;
  */
 export const MAX_LIQUIDATION_INCENTIVE_FACTOR = 1_150000000000000000n;
 
-export const ORACLE_PRICE_OFFSET = 36;
```
Rubilmax, line 13: This because I mis-used it in the API: it is badly named and is useless

oumar-fall (approved):

PR #224, opened 2025-01-03
fix(eip5267): add domain to vault and fix token cache

Rubilmax:


packages/blue-sdk/src/token/Eip5267Domain.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export class Eip5267Domain implements IEip5267Domain {
+  /**
+   * A bit map where bit i is set to 1 if and only if domain field i is present (0 ≤ i ≤ 4).
+   * Bits are read from least significant to most significant, and fields are indexed in the order that is specified by EIP-712, identical to the order in which they are listed in the function type.
+   */
+  public readonly fields;
+
+  /**
+   * The user readable name of signing domain, i.e. the name of the DApp or the protocol.
```
oumar-fall, line 31: this is more likely to be the name of the contract no?

oumar-fall (approved):

packages/blue-sdk/src/token/Eip5267Domain.ts
Rubilmax: Correct, by there's no "name" on a contract per-se and I simply copied EIP712 docs here ^^

PR #246, opened 2025-01-20
chore(migration-sdk-viem): add pkg to deployment

oumar-fall:


Rubilmax (approved):
and it's tested right?

PR #251, opened 2025-01-23
feat: add `PreLiquidationPosition` class

Jean-Grimal:
Given that for the `accrualPosition` class, a significant portion of the logic is actually in the `Market` class (or in `MarketUtils`), I had several options for the functions of the new class I need to create/override:

1. Implement them (if they are new functions, otherwise override them) directly in the `preLiquidatablePosition` class without changing either `accrualPosition` or `Market`.

2. Implement them (if they are new functions, otherwise override them) directly in the `preLiquidatablePosition` class and move the logic from `MarketUtils` to `accrualPosition` (for better consistency).

3. Implement the new logic in `MarketUtils` and/or in a new utility file.

I chose 1 for these first version, but could change.


packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,322 @@
...
+  /**
+   * Whether this position is healthy.
+   * `undefined` iff the market's oracle is undefined or reverts.
+   */
+  get isLiquidatable() {
+    return this.market.isHealthy(this);
+  }
+
+  /**
+   * Whether this position is preLiquidatable.
```
Rubilmax, line 54:
```suggestion
   * Whether this position is liquidatable via the Pre liquidation contract.
```

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,322 @@
...
+    this.preLiquidation = preLiquidation;
+  }
+
+  /**
+   * Whether this position is healthy.
+   * `undefined` iff the market's oracle is undefined or reverts.
+   */
+  get isLiquidatable() {
+    return this.market.isHealthy(this);
+  }
```
Rubilmax, line 51: it should be defined in the position class not this one
And btw this implementation is incorrect it should be !isHealthy

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,322 @@
...
+  /**
+   * Whether this position is preLiquidatable.
+   * `undefined` iff the market's oracle is undefined or reverts.
+   */
+  get isPreLiquidatable() {
+    if (this.ltv === undefined || this.ltv === null) return undefined;
+
+    return (
+      this.ltv > this.preLiquidationParams.preLltv &&
+      this.ltv < this.market.params.lltv
```
Rubilmax, line 62:
```suggestion
      this.ltv <= this.market.params.lltv
```
 Shouldn't it be lte?

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,322 @@
...
+    );
+  }
+
+  /**
+   * Whether this position is healthy.
+   * `undefined` iff the market's oracle is undefined or reverts.
```
Rubilmax, line 68: Its never undefined?

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,322 @@
...
+  /**
+   * The address of the preLiquidation contract.
+   */
+  public readonly preLiquidation: Address;
+
+  constructor(
+    position: IAccrualPosition,
+    market: Market,
+    preLiquidationParams: PreLiquidationParams,
+    preLiquidation: Address,
```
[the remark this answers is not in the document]
Rubilmax: Indeed! I feel like it's easier as a dev instead of remembering parameter order

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
Jean-Grimal: True

packages/blue-sdk-viem/src/fetch/Position.ts
```diff
@@ -55,3 +107,33 @@ export async function fetchAccrualPosition(
 
   return new AccrualPosition(position, market);
 }
+
+export async function fetchPreLiquidatablePosition(
+  user: Address,
+  marketId: MarketId,
+  preLiquidation: Address,
```
Rubilmax, line 114: I thought we said always return a `PreLiquidationPosition`? Like if the dev calls this function, they are only interested in the `PreLiquidationPosition`

packages/blue-sdk-viem/src/fetch/Position.ts
```diff
@@ -55,3 +107,33 @@ export async function fetchAccrualPosition(
...
+) {
+  parameters.chainId = ChainUtils.parseSupportedChainId(
+    parameters.chainId ?? (await getChainId(client)),
+  );
+  const [position, market, preLiquidationParams, isPreLiquidationAuthorized] =
+    await Promise.all([
+      await fetchPosition(user, marketId, client, parameters),
+      await fetchMarket(marketId, client, parameters),
+      await fetchPreLiquidationParameters(preLiquidation, client, parameters),
+      await fetchPreLiquidationAuthorization(user, preLiquidation, client),
```
Rubilmax, line 126: I would make `isPreLiquidationAuthorized` a field of `PreLiquidationPosition` and have it looked for `isPreLiquidatable`

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,330 @@
...
+export interface PreLiquidationParams {
+  preLltv: bigint;
+  preLCF1: bigint;
+  preLCF2: bigint;
+  preLIF1: bigint;
+  preLIF2: bigint;
+  preLiquidationOracle: Address;
+}
+
+export interface IPreLiquidatablePosition extends IPosition {
```
Rubilmax, line 26:
```suggestion
export interface IPreLiquidatablePosition extends IAccrualPosition {
```

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,330 @@
...
+   */
+  public readonly preLiquidationParams: PreLiquidationParams;
+
+  /**
+   * The address of the PreLiquidation contract this position is associated to.
+   */
+  public readonly preLiquidation: Address;
+
+  constructor(
+    position: IAccrualPosition,
```
Rubilmax, line 46:
```suggestion
    position: IPreLiquidatablePosition,
```

Rubilmax: The goal is to have only 2 parameters here, to avoid having 4 parameters and more if we ever extend this class again

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,330 @@
...
+    preLiquidation: Address,
+  ) {
+    super(position, market);
+
+    this.preLiquidationParams = preLiquidationParams;
+    this.preLiquidation = preLiquidation;
+  }
+
+  /**
+   * Whether this position is liquidatable via the Pre liquidation contract.
```
Rubilmax, line 58:
```suggestion
   * Whether this position is liquidatable via the PreLiquidation contract.
```

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,330 @@
...
+   * `undefined` iff the market's oracle is undefined or reverts.
+   */
+  get isPreLiquidatable() {
+    const { ltv } = this;
+    if (ltv == null) return undefined;
```
Rubilmax, line 63:
```suggestion
    if (ltv == null) return ltv;
```

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  /**
+   * Whether this position is healthy.
+   * `undefined` iff the market's oracle is undefined or reverts.
+   */
+  get isHealthy() {
+    const { isPreLiquidatable, isLiquidatable } = this;
+    if (isPreLiquidatable === undefined || isLiquidatable === undefined)
+      return undefined;
```
Rubilmax, line 77:
```suggestion
    if (isPreLiquidatable == null) return isPreLiquidatable;
    if (isLiquidatable == null) return isLiquidatable;
```

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    return !isPreLiquidatable && !isLiquidatable;
+  }
+
+  /**
+   * The price of the collateral quoted in loan assets that would allow this position to be pre-liquidated.
```
Rubilmax, line 83: Sometimes it's `null`, it could be specified in the jsdoc in here

Rubilmax: Also the edge case of maxUint256

packages/blue-sdk-viem/src/fetch/Position.ts
Jean-Grimal: Actually if the PreLiquidation contract is not authorized, the position cannot be pre liquidated
Then I would rather return `undefined` or an error

packages/blue-sdk-viem/src/fetch/Position.ts
Jean-Grimal: So in case `preLiquidationParams` is missing we should return undefined ?
Or an error ?

packages/blue-sdk-viem/src/fetch/Position.ts
```diff
@@ -39,6 +41,56 @@ export async function fetchPosition(
...
+    ...parameters,
+    address: addresses[parameters.chainId].preLiquidationFactory,
+    abi: preLiquidationFactoryAbi,
+    functionName: "isPreLiquidation",
+    args: [preLiquidation],
+  });
+
+  if (!isPreLiquidation) {
+    return undefined;
+  }
```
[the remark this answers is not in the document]
Jean-Grimal: Ok so should I remove this check ?

packages/blue-sdk-viem/src/fetch/Position.ts
```diff
@@ -39,6 +41,56 @@ export async function fetchPosition(
...
+
+  const preLiquidationParameters = await readContract(client, {
+    ...parameters,
+    address: preLiquidation,
+    abi: preLiquidationAbi,
+    functionName: "preLiquidationParams",
+    args: [],
+  });
+
+  return preLiquidationParameters as PreLiquidationParams;
```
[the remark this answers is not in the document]
Jean-Grimal: It defined this function to keep the `await Promise.all` clean (because a lot of calls are parallelized there)

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
Jean-Grimal: Should I also add it to the jsdoc of `liquidationPrice()` ?

packages/blue-sdk-viem/src/fetch/Position.ts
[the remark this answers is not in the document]
Rubilmax: IMO yes

packages/blue-sdk-viem/src/fetch/Position.ts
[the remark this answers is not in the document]
Rubilmax: Was a good move! Just the force cast to change here

packages/blue-sdk-viem/src/fetch/Position.ts
Rubilmax: It either returns something or throws so it's always defined isn't it? You need to remove the factory check and then it'll always be defined

packages/blue-sdk-viem/src/fetch/Position.ts
Rubilmax: It cannot be pre-liquidated but the dev asked for a PreLiquidationPosition
If we want to protect the dev from trying to pre-liquidate something that we'll revert, we simply have to use this field's value in `isPreLiquidatable`

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
Rubilmax: Right, maybe it's already missing!

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
Rubilmax, line 1: Just realized that this naming is maybe not good: it makes me think the position is indeed pre-liquidatable
`PreLiquidationPosition` would be more accurate to designate a position correspondign to a PreLiquidation contract

Rubilmax (approved):

Rubilmax (changes requested):
missclicked

packages/blue-sdk/src/position/PreLiquidatablePosition.ts
Jean-Grimal: Agreed

packages/blue-sdk/src/market/MarketUtils.ts
```diff
@@ -520,7 +520,6 @@ export namespace MarketUtils {
   ) {
     borrowShares = BigInt(borrowShares);
     market.totalBorrowShares = BigInt(market.totalBorrowShares);
```
Rubilmax, line 522: These are not needed anymore

packages/blue-sdk/src/addresses.ts
```diff
@@ -69,6 +70,7 @@ export const addresses = {
     adaptiveCurveIrm: "0x46415998764C29aB2a25CbeA6254146D50D22687",
     publicAllocator: "0xA090dD1a701408Df1d4d0B85b716c87565f90467",
     metaMorphoFactory: "0xA9c3D3a366466Fa809d1Ae982Fb2c46E5fC41101",
+    preLiquidationFactory: "0x8cd16b62E170Ee0bA83D80e1F80E6085367e2aef",
```
Rubilmax, line 73: Can you use it in the liquidation sdk?

Rubilmax (changes requested):

packages/liquidation-sdk-viem/src/abis.ts
```diff
@@ -4222,7 +4222,7 @@ export const preLiquidationAbi = [
   { type: "error", name: "LiquidatablePosition", inputs: [] },
   { type: "error", name: "NonexistentMarket", inputs: [] },
   { type: "error", name: "NotMorpho", inputs: [] },
-  { type: "error", name: "NotPreLiquidatablePosition", inputs: [] },
+  { type: "error", name: "NotPreLiquidationPosition", inputs: [] },
```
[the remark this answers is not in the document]
Jean-Grimal: it is
Nice catch

packages/blue-sdk/src/addresses.ts
Jean-Grimal: Yes

packages/blue-sdk/src/market/MarketUtils.ts
```diff
@@ -518,11 +518,7 @@ export namespace MarketUtils {
     },
     marketParams: { lltv: BigIntish },
   ) {
-    borrowShares = BigInt(borrowShares);
-    market.totalBorrowShares = BigInt(market.totalBorrowShares);
-    if (borrowShares === 0n || market.totalBorrowShares === 0n) return null;
-
-    const borrowAssets = toBorrowAssets(borrowShares, market);
+    const borrowAssets = toBorrowAssets(BigInt(borrowShares), market);
```
Rubilmax, line 521:
```suggestion
    const borrowAssets = toBorrowAssets(borrowShares, market);
```

not needed either (but not blocking)

packages/blue-sdk/src/position/PreLiquidationPosition.ts
```diff
@@ -0,0 +1,326 @@
...
+   * The price of the collateral quoted in loan assets that would allow this position to be pre-liquidated.
+   * `null` if the position has no borrow.
+   */
+  get preLiquidationPrice() {
+    if (this.borrowShares === 0n || this.market.totalBorrowShares === 0n)
+      return null;
+
+    const collateralPower = MarketUtils.getCollateralPower(this.collateral, {
+      lltv: this.market.params.lltv,
+    });
```
Rubilmax, line 94:
```suggestion
    const collateralPower = MarketUtils.getCollateralPower(this.collateral, {
      lltv: this.preLiquidationParams.preLltv,
    });
```

packages/blue-sdk/src/position/PreLiquidationPosition.ts
```diff
@@ -0,0 +1,326 @@
...
+    const withdrawableCollateral = this.market.getWithdrawableCollateral(
+      this,
+      options?.maxLtv
+        ? {
+            maxLtv: MathLib.min(
+              options.maxLtv,
+              this.preLiquidationParams.preLltv,
+            ),
+          }
+        : {},
```
Rubilmax, line 311: same thing as in borrow

packages/blue-sdk/test/unit/PreLiquidationPosition.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    expect(preLiquidationPosition.borrowCapacityUsage).toBe(undefined);
+  });
+
+  test("should not be preLiquidatable because pre liquidation is not authorized", () => {
+    const preLiquidationPosition = new PreLiquidationPosition(
+      { ...position, isPreLiquidationAuthorized: false },
+      new Market(market),
+    );
+
+    expect(preLiquidationPosition.isPreLiquidatable).toBe(false);
```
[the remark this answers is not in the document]
Jean-Grimal: True, tests are missing

packages/blue-sdk/test/unit/PreLiquidationPosition.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      new Market({ ...market, price: undefined }),
+    );
+
+    expect(preLiquidationPosition.isPreLiquidatable).toBe(undefined);
+    expect(preLiquidationPosition.isLiquidatable).toBe(undefined);
+    expect(preLiquidationPosition.ltv).toBe(undefined);
+    expect(preLiquidationPosition.isHealthy).toBe(undefined);
+    expect(preLiquidationPosition.priceVariationToLiquidationPrice).toBe(
+      undefined,
+    );
```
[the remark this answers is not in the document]
Jean-Grimal: No it differentiates `null` and `undefined`, I've noticed it while testing

packages/blue-sdk/src/market/MarketUtils.ts
Rubilmax, line 1: In the end, I am convinced this is a breaking change and as such, you should rebase this PR on `next`

You will have many conflicts, notably with `addresses` (you'll have to add pre liquidation factory addresses for all chains) but it is much, much safer for integrators

Also, you can take this opportunity to squash all bad commits (those that do not comply with commitlint)

Rubilmax (changes requested):

PR #252, opened 2025-01-23
feat(liquidity-loader): add parameters to loader

oumar-fall:
User should be able to use his own parameters to compute the shared liquidity. For exemple, interface is enforcing 100% max withdraw utilisation which is not possible with the current implem

TODO: It is currently not possible to filter reallocating vaults by fee for instance

Fixes INFRA-206

### DIFF
--- a/packages/liquidity-sdk-viem/src/loader.ts
+++ b/packages/liquidity-sdk-viem/src/loader.ts
@@ -20,6 +20,15 @@ import { apiSdk } from "./api";
 export interface LiquidityParameters {
   /* The delay to consider between the moment reallocations are calculated and the moment they are committed onchain. Defaults to 1h. */
   delay?: bigint;
+
+  /** The default maximum utilization allowed to reach to find shared liquidity (scaled by WAD). */
+  defaultMaxWithdrawalUtilization?: bigint;
+
+  /**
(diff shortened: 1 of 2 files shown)


packages/liquidity-sdk-viem/src/loader.ts
```diff
@@ -18,8 +18,17 @@ import { getBlock } from "viem/actions";
...
   delay?: bigint;
+
+  /** The default maximum utilization allowed to reach to find shared liquidity (scaled by WAD). */
+  defaultMaxWithdrawalUtilization?: bigint;
+
+  /**
+   * If set, used to retrieve the max withdraw utilization for each market, defaulting to `defaultMaxWithdrawalUtilization`.
+   * If not, these values are fetched from morpho API
+   */
+  maxWithdrawUtilizations?: Record<MarketId, bigint>;
```
Rubilmax, line 31:
```suggestion
   * If provided, defines the maximum utilization allowed to reach for each market, defaulting to `defaultMaxWithdrawalUtilization`.
   * If not, these values are fetched from Morpho API.
   */
  maxWithdrawalUtilization?: Record<MarketId, bigint>;
```

Rubilmax (dismissed):

Rubilmax (approved):

PR #258, opened 2025-02-02
fix(liquidation-sdk-viem): fix fetching from Paraswap API

haydenshively:
Fixes two things:

- Since the executooor contract doesn't have tokens until inside a callback, we have to tell Paraswap to skip balance checks by passing `{ ignoreChecks: true }`
- Paraswap API expects slippage in basis points; we were passing percentages

### DIFF
--- a/packages/liquidation-sdk-viem/src/swap/1inch.ts
+++ b/packages/liquidation-sdk-viem/src/swap/1inch.ts
@@ -14,10 +14,15 @@ export namespace OneInch {
     apiKey = process.env.ONE_INCH_SWAP_API_KEY,
   ) {
     const url = new URL(getSwapApiPath(swapParams.chainId), API_BASE_URL);
-    url.searchParams.set("slippage", BigInt(swapParams.slippage).toString(16));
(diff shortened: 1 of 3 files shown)


packages/liquidation-sdk-viem/src/swap/1inch.ts
```diff
@@ -14,10 +14,15 @@ export namespace OneInch {
     apiKey = process.env.ONE_INCH_SWAP_API_KEY,
   ) {
     const url = new URL(getSwapApiPath(swapParams.chainId), API_BASE_URL);
-    url.searchParams.set("slippage", BigInt(swapParams.slippage).toString(16));
```
haydenshively, line 17: This line did nothing, as `"slippage"` would be overwritten in the following code block. This is fortunate, as hexadecimal slippage doesn't seem compatible with the [1inch API spec](https://portal.1inch.dev/documentation/apis/swap/classic-swap/Parameter%20Descriptions/swap_params).

Rubilmax (approved):

PR #259, opened 2025-02-03
Replicate liquidity sdk public allocator options to ethers

Rubilmax:


packages/liquidity-sdk-viem/src/loader.ts
```diff
@@ -205,26 +210,21 @@ export class LiquidityLoader<chain extends Chain = Chain> {
...
-            ),
-          );
+        parameters.maxWithdrawalUtilization ??= fromEntries(
+          allVaultsMarkets.flatMap(([, markets]) =>
+            markets.map((market) => [
+              market.uniqueKey,
+              market.targetWithdrawUtilization,
+            ]),
+          ),
+        );
```
oumar-fall, line 220: This makes the constructor not pure so i'm not a huge fan :/ It might be okay in this specific case

oumar-fall (dismissed):

PR #262, opened 2025-02-07
feat: new reallocation algorithm

Jean-Grimal:


packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -326,18 +326,41 @@ export const populateSubBundle = (
...
           : MathLib.wDivDown(newTotalBorrowAssets, supplyTargetUtilization) -
             newTotalSupplyAssets;
 
-      const { withdrawals } = data.getMarketPublicReallocations(
+      let { withdrawals } = data.getMarketPublicReallocations(
         market.id,
         publicAllocatorOptions,
       );
 
+      const asssetsWithdrawn = withdrawals.reduce(
```
Rubilmax, line 342:
```suggestion
      const reallocatedAssets = withdrawals.reduce(
```

sss 🐍

packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -326,18 +326,41 @@ export const populateSubBundle = (
...
+        (total, { assets }) => total + assets,
+        0n,
+      );
+
+      if (
+        MarketUtils.getUtilization({
+          totalSupplyAssets: newTotalSupplyAssets + asssetsWithdrawn,
+          totalBorrowAssets: newTotalBorrowAssets,
+        }) > MathLib.WAD
```
Rubilmax, line 351:
```suggestion
        newTotalSupplyAssets + asssetsWithdrawn < newTotalBorrowAssets
```

IMO it's much easier to reason this way here

packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -326,18 +326,41 @@ export const populateSubBundle = (
...
+      ) {
+        // If the target friendly reallocation is not enough, we fully withdraw from every market.
```
Rubilmax, line 353: This is redundant with L330. I would suggest to remove L330

Rubilmax:
```suggestion
        // If the "friendly" reallocations are not enough, we fully withdraw from every market.
```

packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -326,18 +326,41 @@ export const populateSubBundle = (
...
+
+        requiredAssets =
+          MathLib.wDivDown(newTotalBorrowAssets, MathLib.WAD) -
+          newTotalSupplyAssets;
+
+        withdrawals = data.getMarketPublicReallocations(market.id, {
+          defaultMaxWithdrawalUtilization: MathLib.WAD,
+        }).withdrawals;
```
Rubilmax, line 361:
```suggestion
        ({ withdrawals } = data.getMarketPublicReallocations(market.id, {
          ...publicAllocatorOptions,
          defaultMaxWithdrawalUtilization: MathLib.WAD,
          maxWithdrawalUtilization: {},
        }));
```

packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -326,18 +326,41 @@ export const populateSubBundle = (
...
+        MarketUtils.getUtilization({
+          totalSupplyAssets: newTotalSupplyAssets + asssetsWithdrawn,
+          totalBorrowAssets: newTotalBorrowAssets,
+        }) > MathLib.WAD
+      ) {
+        // If the target friendly reallocation is not enough, we fully withdraw from every market.
+
+        requiredAssets =
+          MathLib.wDivDown(newTotalBorrowAssets, MathLib.WAD) -
+          newTotalSupplyAssets;
```
Rubilmax, line 357: This is exactly `newTotalBorrowAssets - newTotalSupplyAssets`

packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -326,18 +326,41 @@ export const populateSubBundle = (
...
+      // We first try to find public reallocations that respect every markets targets.
+      // If this is not enough, the first market to be pushed above target is the supply market. Then we fully withdraw from every market.
       let requiredAssets =
         supplyTargetUtilization === 0n
           ? MathLib.MAX_UINT_160
           : MathLib.wDivDown(newTotalBorrowAssets, supplyTargetUtilization) -
             newTotalSupplyAssets;
 
-      const { withdrawals } = data.getMarketPublicReallocations(
+      let { withdrawals } = data.getMarketPublicReallocations(
```
Rubilmax, line 337: There's a field in the return value named `data` which already contains the simulated state of the borrowed market; which you can use to get the resulting utilization after the "friendly" reallocations

Rubilmax (changes requested):

packages/bundler-sdk-viem/src/operations.ts
Jean-Grimal: True

Jean-Grimal: But working with utilizations and `WAD` helps to understand the reasoning (about the 100% utilization)

packages/bundler-sdk-viem/src/operations.ts
Jean-Grimal: Yes but I feel more comfortable with the current way

Rubilmax: This would avoid having to calculate `reallocatedAssets` because it would automatically be included in the simulated data `market.totalSupplyAssets` but we would have to reassign `newTotalSupplyAssets` and `newTotalBorrowAssets`

It's not a huge simplification, but I still think it's preferable as we don't re-simulate the effect of reallocating (`newTotalSupplyAssets + reallocatedAssets`) and we instead rely on the liveness if the simulation to correctly simulate the effect of the reallocation but we re-simulate the effect of the hypothetical withdraw (`- withdrawnAssets`)

haydenshively (approved):
Difficult to verify correctness since a lot is done on the frontend, but I think this looks reasonable.

packages/bundler-sdk-viem/test/populateBundle.test.ts
```diff
@@ -5047,6 +5047,765 @@ describe("populateBundle", () => {
...
+                receiver: client.account.address,
+                slippage: DEFAULT_SLIPPAGE_TOLERANCE,
+              },
+            },
+          ]);
+        },
+      );
+
+      test[ChainId.EthMainnet](
+        "should borrow USDC with shared liquidity and full reallocation",
```
Rubilmax, line 5545: Actually wdyt about separating these to `shared-liquidity.spec.ts` as you did initially? I think this is a great opportunity to start segmenting tests

Jean-Grimal: No pb

packages/bundler-sdk-viem/test/sharedLiquidity.test.ts
```diff
@@ -0,0 +1,803 @@
...
+import { parseEther, parseUnits } from "viem";
+import { describe, expect } from "vitest";
+import { donator, setupBundle } from "./helpers.js";
+import { test } from "./setup.js";
+
+dotenv.config();
+
+configure({ asyncUtilTimeout: 10_000 });
+
+describe("populateBundle", () => {
```
Rubilmax, line 23:
```suggestion
describe("sharedLiquidity", () => {
```

packages/bundler-sdk-viem/test/sharedLiquidity.test.ts
```diff
@@ -0,0 +1,803 @@
...
+  describe("with signatures", () => {
+    describe("ethereum", () => {
```
Rubilmax, line 25: These 2 nested describe are not useful here (all tests in this file are on eth and use signatures) I would rather get rid of them

packages/bundler-sdk-viem/test/sharedLiquidity.test.ts
```diff
@@ -0,0 +1,803 @@
...
+import { useSimulationState } from "@morpho-org/simulation-sdk-wagmi";
+import { renderHook, waitFor } from "@morpho-org/test-wagmi";
+import { configure } from "@testing-library/dom";
+import dotenv from "dotenv";
+import { parseEther, parseUnits } from "viem";
+import { describe, expect } from "vitest";
+import { donator, setupBundle } from "./helpers.js";
+import { test } from "./setup.js";
+
+dotenv.config();
```
Rubilmax, line 19: This should not be necessary with `vitest` (populateBundle.test.ts does not have this)

Rubilmax (approved):

Rubilmax (approved):

PR #263, opened 2025-02-07
feat: chain support for OP mainnet and ARB

cashd:
- Add chain and deployment info for ARB 
https://github.com/morpho-org/deployments/blob/main/address-book.json#L11

- Add chain and deployment info for OP mainnet
https://github.com/morpho-org/deployments/blob/main/address-book.json#L47


oumar-fall (dismissed):

packages/blue-sdk/src/chain.ts
```diff
@@ -101,6 +101,28 @@ export namespace ChainUtils {
...
     },
+    [ChainId.ArbitrumMainnet]: {
+      name: "Arbitrum Mainnet",
+      id: ChainId.ArbitrumMainnet,
+      nativeCurrency: { name: "Ether", symbol: "ETH", decimals: 18 },
+      defaultRpcUrl: "https://arb1.arbitrum.io/rpc",
+      explorerUrl: "https://arbiscan.io",
+      isTestnet: false,
+      shortName: "Arbitrum",
+      logoSrc: "https://cdn.morpho.org/assets/chains/arbitrum.png",
```
Rubilmax, line 112: This logo does not exist, we should either upload it, make `logoSrc` optional OR ref another valid logo url (we probably should not as it would  not be a trusted url)

packages/blue-sdk/src/addresses.ts
```diff
@@ -75,6 +75,28 @@ export const addresses = {
     verUsdc: "0x59aaF835D34b1E3dF2170e4872B785f11E2a964b",
     testUsdc: "0xBC77067f829979812d795d516E523C4033b66409",
   },
+  [ChainId.ArbitrumMainnet]: {
+    morpho: "0x6c247b1F6182318877311737BaC0844bAa518F5e",
+    bundler: "0x044aE7bA624a047535F4A6ae5FE6E9EA8d1D93B0",
```
Rubilmax, line 80: This corresponds to `bundler3`, but it should correspond to bundler v2 which is not deployed on Arbitrum... This is blocking as it would introduce unexpected behaviors (most probably 100% failing bundles). Let's wait for bundler3 support coming this week

Rubilmax (changes requested):

oumar-fall (commented):
didn't check the addresses

oumar-fall (approved):
I didn't check the addresses but LGTM

PR #264, opened 2025-02-12
fix(morpho-ts): updated format with unit & sign

Oscarmacieira:
# Motivation.

Currently, when using `.sign().unit("$")` the output is `$-X` and `$+X`. We would need `-$X` and `+$X`.

Fixes:INFRA-262

# Solution

- Within the `_withUnit` function in `morpho-ts`, I handled the "$" case to include the sign and unit
- Added more tests for this in the suite.

Let me know if I am missing any implications that this changed might introduce!
Thanks


Rubilmax (commented):
I don't have enough context on this code to make sure these changes are optimal, but the changes look good!

oumar-fall (changes requested):
I request these changes

packages/morpho-ts/src/format/format/format.ts
```diff
@@ -261,11 +263,7 @@ function formatBI(
...
       : formatOptions.sign
         ? "+"
-        : "") + value;
-
-  if (formatOptions.unit) {
-    return _withUnit(formattedValue, formatOptions.unit);
-  }
+        : "") + _withUnit(value, formatOptions.unit);
 
   return formattedValue;
```
oumar-fall, line 267: no need to store it in `formattedValue`, you can return directly the value

oumar-fall (changes requested):
one last little thing ;-)

Oscarmacieira:
Thanks @oumar-fall for your comments. Indeed much clearer like this.

oumar-fall (approved):
works like a charm! thank you sir

PR #266, opened 2025-02-17
Simplify Holding permit2 allowances

Rubilmax:
A holding would always fetch the permit2 allowance of Morpho, adding an unnecessary rpc call. So this PR gets rid of Morpho's permit2 allowance in Holdings.

BREAKING CHANGE: replaced `permit2Allowances` with `permit2BundlerAllowance`


julien-devatom (approved):
Nothing to say around the content. Small remark around the context:
- should be flagged as breaking change PR (to not miss to mention it in the changelog) 
- can you please mention the rationale somewhere? I get why it's useless but I don't get why it's blocking

Rubilmax:
> * should be flagged as breaking change PR (to not miss to mention it in the changelog)

It was already via commits (the first one is breaking and is isolated to blue-sdk for this purpose)

> * can you please mention the rationale somewhere? I get why it's useless but I don't get why it's blocking

Yeah it seems to only be an optimization, but it's not blocking for bundler3 - i simply isolated this change for the other PR to be simpler to review

packages/blue-sdk-viem/src/fetch/Holding.ts
```diff
@@ -64,8 +58,11 @@ export async function fetchHolding(
     try {
       const {
         balance,
-        erc20Allowances,
-        permit2Allowances,
+        erc20Allowances: {
+          bundler: generalAdapter1Erc20Allowance,
```
oumar-fall, line 62: This is technically not true yet haha

oumar-fall (approved):

packages/blue-sdk-viem/src/fetch/Holding.ts
Rubilmax: Shit tant pis it's only naming

PR #270, opened 2025-02-24
feat(chains): add all low support chains

Rubilmax:
Fixes INFRA-302

### DIFF
--- a/packages/blue-sdk/src/addresses.ts
+++ b/packages/blue-sdk/src/addresses.ts
@@ -119,7 +119,6 @@ export const addresses = {
     metaMorphoFactory: "0xa9c87daB340631C34BB738625C70499e29ddDC98",
 
     wNative: "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270",
-    usdc: "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174",
   },
   [ChainId.ArbitrumMainnet]: {
     morpho: "0x6c247b1F6182318877311737BaC0844bAa518F5e",
--- a/packages/blue-sdk/src/addresses.ts
+++ b/packages/blue-sdk/src/addresses.ts
@@ -106,6 +106,21 @@ export const addresses = {
     verUsdc: "0x59aaF835D34b1E3dF2170e4872B785f11E2a964b",
     testUsdc: "0xBC77067f829979812d795d516E523C4033b66409",
   },
+  [ChainId.PolygonMainnet]: {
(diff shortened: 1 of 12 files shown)


packages/migration-sdk-viem/src/config.ts
```diff
@@ -140,6 +140,13 @@ export const MIGRATION_ADDRESSES = {
     [MigratableProtocol.aaveV2]: null,
     [MigratableProtocol.aaveV3Optimizer]: null,
   },
+  [ChainId.PolygonMainnet]: {
```
oumar-fall, line 143: Why not making it optional by chain instead of specifying null addresses

oumar-fall (dismissed):

packages/blue-sdk/src/addresses.ts
```diff
@@ -106,6 +106,21 @@ export const addresses = {
...
+      generalAdapter1: "0xB261B51938A9767406ef83bbFbaAFE16691b7047",
+      paraswapAdapter: "0x5F2617F12D1fDd1e43e72Cb80C92dFcE8124Db8d",
+    },
+    permit2: "0x000000000022D473030F116dDEE9F6B43aC78BA3",
+    adaptiveCurveIrm: "0xe675A2161D4a6E2de2eeD70ac98EEBf257FBF0B0",
+    publicAllocator: "0xfac15aff53ADd2ff80C2962127C434E8615Df0d3",
+    metaMorphoFactory: "0xa9c87daB340631C34BB738625C70499e29ddDC98",
+
+    wNative: "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270",
+    usdc: "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174",
```
cashd, line 122: This is the bridged USDC version. The native USDC is https://polygonscan.com/token/0x3c499c542cef5e3811e1192ce70d8cc03d5c3359. Although the bridged has higher total supply.

cashd (approved):
just reviewed all the addresses, lgtm!

packages/blue-sdk/src/addresses.ts
Rubilmax: I will simply remove it then, to make sure we don't use it in a bad way

packages/migration-sdk-viem/src/config.ts
Rubilmax: Because we declare it as `const` and use it as is ; we'd need a typed accessor

PR #272, opened 2025-02-26
feat(chain-id): replace with number

Rubilmax:
Fixes INFRA-317


packages/blue-sdk/src/addresses.ts
```diff
@@ -102,6 +146,7 @@ export const addresses = {
     metaMorphoFactory: "0xFf62A7c278C62eD665133147129245053Bbf5918",
 
     wNative: "0x4200000000000000000000000000000000000006",
+    // Must implement USDC permit version 2 (otherwise breaks permit signatures).
```
oumar-fall, line 149: Might be an idea to expose a script that checks these requirements

oumar-fall: so that someone that extends the config could ensure it's compatible

packages/migration-sdk-viem/src/positions/supply/aaveV3Optimizer.supply.ts
```diff
@@ -71,10 +71,10 @@ export class MigratableSupplyPosition_AaveV3Optimizer
     if (aaveV3OptimizerMigrationAdapter == null)
       throw new Error("missing aaveV3OptimizerMigrationAdapter address");
 
-    const migrationAddresses =
-      MIGRATION_ADDRESSES[chainId][MigratableProtocol.aaveV3Optimizer];
+    const aaveV3OptimizerAddresses =
+      migrationAddresses[chainId]?.[MigratableProtocol.aaveV3Optimizer];
 
-    if (!migrationAddresses) throw new UnsupportedChainIdError(chainId);
+    if (!aaveV3OptimizerAddresses) throw new UnsupportedChainIdError(chainId);
```
oumar-fall, line 77: shouldn't we expose a `getMigrationAddresses` that works the same as `getChainAddresses`?

oumar-fall (approved):

packages/migration-sdk-viem/src/positions/supply/aaveV3Optimizer.supply.ts
Rubilmax: Indeed but less prio

packages/blue-sdk/src/addresses.ts
Rubilmax: Better would be to have the version parameter in args of permit signature

Rubilmax: But we need the frontend to give it to us and if we push it too early they won't be compatible and it'll break signatures in production

PR #273, opened 2025-02-27
feat: add option to attach anvil logs to playwright report

0xbulma:
- add option to attach anvil logs in `createViemTest` to playwright report. 

Can be seen here:
Screenshot 2025-02-27 at 15 15 15

### DIFF
--- a/packages/test/src/anvil.ts
+++ b/packages/test/src/anvil.ts
@@ -314,12 +314,6 @@ const basePort =
 
 let workerInstances = 0;
 
-export interface Props {
-  args: AnvilArgs;
-  index?: number;
-  callbacks?: AnvilProcessCallbacks;
-}
-
 export const spawnAnvil = async (
   args: AnvilArgs,
   index = workerInstances++,
--- a/packages/test/src/anvil.ts
+++ b/packages/test/src/anvil.ts
@@ -1,8 +1,6 @@
 import { spawn } from "node:child_process";
 import _kebabCase from "lodash.kebabcase";
 
-import type { TestInfo } from "@playwright/test";
-
 export interface AnvilArgs {
   /**
    * Number of dev accounts to generate and configure.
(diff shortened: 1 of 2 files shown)


packages/test/src/anvil.ts
```diff
@@ -302,20 +304,25 @@ let workerInstances = 0;
 export const spawnAnvil = async (
   args: AnvilArgs,
   index = workerInstances++,
+  testInfo?: TestInfo,
```
Rubilmax, line 307: IMO most versatile would be to expect 2 callbacks here: `onMessage` and `onError` and simply call each of them in place of your changes

Rubilmax: So the implementation is not only bound to playwright (or playwright-like `testInfo`) and you can do many more things than attaching logs

0xbulma: yes indeed I broke the layer

packages/test/src/playwright.ts
```diff
@@ -7,9 +7,40 @@ export interface PlaywrightTestContext<chain extends Chain = Chain> {
...
+      contentType: "text/plain",
+    });
+  };
+
+  return {
+    onMessage: attachLog,
+    onError: attachLog,
+    onClose: onClose,
+  };
+};
```
Rubilmax, line 31: To me this should be end user code because we're preventing devs from passing their own callbacks to `createViemTest`

I would simply expect the same optional argument `callbacks?: AnvilProcessCallbacks`

And for its vitest counterpart too! (also named `createViemTest`)

0xbulma: Don't you think that making it native to your playwright wrapper is a good use case? 
But happy to move it wherever you want or even outside of the repo.

Rubilmax: We can export the resulting object (the 3 callbacks) as a constant `attachLogs` from this file if you want! So you just have to pass it as the third optional parameter, in place of callbacks

PR #274, opened 2025-02-27
Feat/infra 301/borrow migration

oumar-fall:
Fixes: INFRA-301


packages/migration-sdk-viem/src/positions/supply/aaveV2.supply.ts
```diff
@@ -55,6 +55,8 @@ export class MigratableSupplyPosition_AaveV2
     chainId: ChainId,
     supportsSignature = true,
   ) {
+    this.validateMigration({ amount });
```
Rubilmax, line 58: If this should be enforced IMO we should call this in the abstract super class and then call `_getMigrationTx` which child classes should implement: so we are 100% certain migration is validated when `_getMigrationTx` is called (beware of making it protected)

PR #289, opened 2025-03-25
feat(pre-liquidation-params): add registry

Rubilmax:
Fixes INFRA-445

### DIFF
--- a/packages/blue-sdk/src/position/PreLiquidationPosition.ts
+++ b/packages/blue-sdk/src/position/PreLiquidationPosition.ts
@@ -7,20 +7,63 @@ import {
   type MaxWithdrawCollateralOptions,
 } from "../market";
 import { MathLib, SharesMath } from "../math";
+import type { BigIntish } from "../types";
 import { AccrualPosition, type IAccrualPosition } from "./Position";
 
-export interface PreLiquidationParams {
-  preLltv: bigint;
-  preLCF1: bigint;
-  preLCF2: bigint;
-  preLIF1: bigint;
-  preLIF2: bigint;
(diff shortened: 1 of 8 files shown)


packages/blue-sdk/src/preLiquidation.ts
Jean-Grimal, line 1: Maybe should be renamed `defaultPreLiquidationParams`

oumar-fall (approved):
Don't have enough context to be confident about all the logic, but code seems fine to me

packages/blue-sdk/src/preLiquidation.ts
Rubilmax: For now let's keep the broad naming because we may have other utilities/constants related to pre-liquidation to store in it

packages/blue-sdk/src/position/PreLiquidationPosition.ts
```diff
@@ -1,228 +1,188 @@
...
+      preLiquidationOraclePrice,
+      ...position
+    }: IPreLiquidationPosition,
+    market: IMarket,
+  ) {
+    super(position, {
+      ...market,
+      params: {
+        ...market.params,
+        lltv: BigInt(preLiquidationParams.preLltv),
```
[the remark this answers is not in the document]
Rubilmax: This changes are made so calculations of the position's metrics are using the pre-liquidation values (prices, lltv)

But changing the others don't seem relevant and may lead to silent bugs... So I'm against

packages/blue-sdk/src/preLiquidation.ts
Rubilmax: Need to export the file from index though

PR #298, opened 2025-04-02
Feat/set authorization operation

oumar-fall:


Rubilmax (approved):
SetAuthorization is still not an input bundler operation right? But maybe they can manually add it before finalizeBundle

oumar-fall:
> SetAuthorization is still not an input bundler operation right? But maybe they can manually add it before finalizeBundle

Yes it is due to the changes in bundler-sdk-viem

PR #299, opened 2025-04-02
fix(bundler-sdk-viem): remove duplicates native transfer

oumar-fall:
This line was adding a duplicate transfer since native transfer is already handled from the action

### DIFF
--- a/packages/bundler-sdk-viem/test/populateBundle.test.ts
+++ b/packages/bundler-sdk-viem/test/populateBundle.test.ts
@@ -4773,6 +4773,93 @@ describe("populateBundle", () => {
           ); // we normally didn't experienced any slippage
         },
       );
+
+      test[ChainId.EthMainnet](
+        "should wrap ETH",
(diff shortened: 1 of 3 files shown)


Rubilmax (approved):

packages/bundler-sdk-viem/test/populateBundle.test.ts
```diff
@@ -4773,6 +4773,93 @@ describe("populateBundle", () => {
...
+                sender: client.account.address,
+                address: wNative,
+                args: {
+                  amount: assets,
+                  owner: client.account.address,
+                  slippage: DEFAULT_SLIPPAGE_TOLERANCE,
+                },
+              },
+            ],
+            { gasPrice: 0n },
```
Rubilmax, line 4819: gasPrice should be 1 or at least deterministic so the gas fees should be too
But let's keep this

Rubilmax (approved):

PR #308, opened 2025-04-11
fix(bundler3): force permit/authorize generalAdapter1

Rubilmax:


packages/simulation-sdk/src/handlers/blue/setAuthorization.ts
oumar-fall, line 1: I implemented this recently, it's needed for batching pre-liquidation (you need to authorize pre-liquidation contract)

packages/bundler-sdk-viem/src/BundlerAction.ts
```diff
@@ -1020,7 +1027,15 @@ export namespace BundlerAction {
         data: encodeFunctionData({
           abi: blueAbi,
           functionName: "setAuthorizationWithSig",
-          args: [authorization, { v: yParity + 27, r, s }],
+          args: [
+            {
+              ...authorization,
+              // Never permit any other address than the GeneralAdapter1 otherwise
+              // the signature can be extracted and used independently.
+              authorized: generalAdapter1,
```
oumar-fall, line 1035: That becomes wrong

That's why I would rather throw here is unexpected value rather than silently overiding

oumar-fall (changes requested):

packages/bundler-sdk-viem/src/BundlerAction.ts
adhusson, line 1: For simplicity, continuing [discussion](https://github.com/morpho-org/sdks-bundler3-review/pull/1#discussion_r2048643292) about keeping the default `??= generalAdapter1` on migrations:

> Enforced recipient to be defined except for migration adapters where it is handy to:

> * not have to fetch the adapter's address from the chain's registry of addresses BEFORE encoding it
> * not have to check it is actually defined in the chain's registry of addresses UNTIL it is strictly required (ie at encoding)

I don't completely get the above 2 points, do you mean that
- the receiver can be a contract different from the migration adapter, but
- when the receiver is the migration adapter itself, it's hard to get its address at the call site of the function?

If yes I would either
- add an argument `useMigrationAdapter: bool`, with the checked invariant `useMigrationAdapter == (recipient != null)`, or
- make specialized functions `compoundV2RedeemToMigrationAdapter` + `compoundV2Redeem` for each.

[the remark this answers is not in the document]
Rubilmax: Skip revert was a feature request from paperclip on all actions in order to sometimes do something with the dust instead of skimming it

Rubilmax: I forced recipient to be defined in the action as first suggested, it's ok now

packages/bundler-sdk-viem/src/BundlerAction.ts
```diff
@@ -451,28 +436,21 @@ export namespace BundlerAction {
   }
 
   /**
-   * Encodes a call to the Adapter to transfer ERC20 tokens.
+   * Encodes a call to the GeneralAdapter1 to transfer ERC20 tokens.
```
oumar-fall, line 439:
```suggestion
   * Encodes a call to the requested adapter to transfer ERC20 tokens.
```

packages/bundler-sdk-viem/src/BundlerAction.ts
```diff
@@ -489,26 +467,24 @@ export namespace BundlerAction {
   }
 
   /**
-   * Encodes a call to the Adapter to transfer ERC20 tokens from the sender to the Bundler.
+   * Encodes a call to the GeneralAdapter1 to transfer ERC20 tokens with `transferFrom`.
```
oumar-fall, line 470: Here this is indeed `GeneralAdapter1`

packages/bundler-sdk-viem/src/BundlerAction.ts
```diff
@@ -1011,9 +998,15 @@ export namespace BundlerAction {
...
-    const { morpho } = getChainAddresses(chainId);
+    const {
+      morpho,
+      bundler3: { bundler3 },
+    } = getChainAddresses(chainId);
     const { r, s, yParity } = parseSignature(signature);
 
+    if (authorization.authorized === bundler3)
+      throw new BundlerErrors.UnexpectedSignature(authorization.authorized);
+
```
oumar-fall, line 1009: Shuoldn't we do the same with permits and approvals?

Rubilmax: In these cases the space of spenders is very limited

adhusson (approved):

oumar-fall (approved):

PR #317, opened 2025-04-28
feat(bundler-sdk-viem): add morpho wrapping action encoding

oumar-fall:


packages/bundler-sdk-viem/src/BundlerAction.ts
```diff
@@ -707,6 +712,40 @@ export namespace BundlerAction {
     ];
   }
 
+  /**
+   * Encodes a call to the GeneralAdapter1 to wrap legacy MORPHO tokens.
+   * @param chainId The chain id for which to encode the call.
+   * @param recipient The recipient of MORPHO tokens.
+   * @param amount The amount of tokens to wrap.
+   * @param skipRevert Whether to allow the wrapp to revert without making the whole bundler revert. Defaults to false.
```
Rubilmax, line 720:
```suggestion
   * @param skipRevert Whether to allow the wrap to revert without making the whole bundler revert. Defaults to false.
```

Rubilmax (approved):

packages/bundler-sdk-viem/src/BundlerAction.ts
```diff
@@ -707,6 +712,40 @@ export namespace BundlerAction {
...
+    chainId: ChainId,
+    recipient: Address,
+    amount: bigint,
+    skipRevert = false,
+  ): BundlerCall[] {
+    if (chainId !== ChainId.EthMainnet)
+      throw new Error("MORPHO wrapping is only available on ethereum mainnet");
+    const {
+      bundler3: { generalAdapter1 },
+    } = getChainAddresses(chainId);
```
adhusson, line 732: `morphoWrapperWithdrawTo` is missing, but maybe never used?

adhusson (approved):

packages/bundler-sdk-viem/src/BundlerAction.ts
oumar-fall: I thought that since it's not likely to be used (is there a use-case where we wanna retrieve legacy tokens?) we better not expose it to avoid confusion and useless maintainance

oumar-fall: wdyt?

PR #326, opened 2025-05-02
fix(addresses): add camp mainnet

tomrpl:
Sibling PR to https://github.com/morpho-org/deployments/pull/84

Addresses are matching the one from the doc:
https://docs.morpho.org/overview/resources/addresses/

And from the deployment batch3:
https://github.com/morpho-org/deployments/blob/999de1d4b9ad3ccd0abba86f9a1e89cb4b42df19/address-book.json


Rubilmax (approved):

oumar-fall (approved):
checked all infos

PR #331, opened 2025-05-06
fix(migration-sdk-viem): fix aave pricing

oumar-fall:
changes: 
- rename variables as price is in USD and not in ETH on aaveV3
- priceSource should be set to market address if not defined (equal to `0x0000...`)


Rubilmax (approved):
Ok to merge this breaking change because feature was not used until 2d ago where we started to export MigratableBorrowPosition

PR #334, opened 2025-05-07
fix(simulation): replace infinite balance with dynamic tracking

Rubilmax:
Inside callbacks, we were not able to tell how much the GeneralAdapter1 had (all of its holdings were faked to be at least maxUint160) which is inconvenient when you want to use `maxUint256` edge cases of the GeneralAdapter1

This PR changes the way we simulate these balances by simply keeping track of whenever the balances of the adapter falls short and by how much, in another variable that is reported at the end

### DIFF
--- a/packages/bundler-sdk-viem/src/operations.ts
+++ b/packages/bundler-sdk-viem/src/operations.ts
@@ -922,7 +922,7 @@ class VirtualHolding extends Holding {
   }
   set balance(value: bigint) {
     if (value < 0n) {
-      this.required = (this.required ?? 0n) + value;
+      this.required += -value;
 
       this._balance = 0n;
     } else this._balance = value;
@@ -955,7 +955,7 @@ export const simulateRequiredTokenAmounts = (
     .map(([token, holding]) => ({
       token,
       // Safe cast because the holding was transformed to a VirtualHolding.
-      required: -(holding as VirtualHolding).required,
+      required: (holding as VirtualHolding).required,
     }))
     .filter(({ required }) => required > 0n);
 };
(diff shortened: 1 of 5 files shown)


packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -919,6 +914,21 @@ export const populateBundle = (
...
 
+class VirtualHolding extends Holding {
+  public required = 0n;
+
+  get balance() {
+    return this._balance;
+  }
+  set balance(value: bigint) {
+    if (value < 0n) {
+      this.required = (this.required ?? 0n) + value;
```
oumar-fall, line 925: Why is `this.required` optional?

oumar-fall: required is negative, isn't it weird?

packages/simulation-sdk-wagmi/test/hooks/useSimulationState.test.ts
```diff
@@ -289,6 +289,13 @@ describe("useSimulationState", () => {
     client,
     expect,
   }) => {
+    const amount = 1_000000n;
+
+    await client.approve({
+      address: usdc,
+      args: [morpho, amount],
+    });
```
oumar-fall, line 297: what does it change?

oumar-fall (commented):
shouldn't we add a use case for the maxUint256 use-case?

packages/simulation-sdk-wagmi/test/hooks/useSimulationState.test.ts
Rubilmax: Was failing with "insufficient allowance" insteaf of "insufficient balance"

oumar-fall (approved):

PR #340, opened 2025-05-15
feat(blue-sdk-viem): provide utility for restructuring viem arrays->objs

haydenshively:
Provide a utility that converts viem's array-style outputs back to objects by referencing the ABI. Works whenever Solidity tuples are named, which is the case for most (all?) Morpho contracts.

One thing that may be unexpected a priori is that you have to pass placeholder args to the restructuring function, i.e. "0x" for an `address` param -- this is so that the type system can disambiguate between overloaded methods.

I've adapted some of the SDK code to use the new utility, so we have TypeScript-level correctness guarantees going forward (no chance of accidentally ordering array destructuring elements incorrectly).

### DIFF
--- a/packages/blue-sdk-viem/src/fetch/Market.ts
+++ b/packages/blue-sdk-viem/src/fetch/Market.ts
@@ -12,6 +12,7 @@ import type { DeploylessFetchParameters } from "../types";
 
 import { adaptiveCurveIrmAbi, blueAbi, blueOracleAbi } from "../abis";
 import { abi, code } from "../queries/GetMarket";
+import { restructure } from "../utils";
 
 export async function fetchMarket(
   id: MarketId,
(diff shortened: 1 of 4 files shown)


packages/blue-sdk-viem/src/fetch/Position.ts
```diff
@@ -22,20 +23,21 @@ export async function fetchPosition(
...
-  });
+  const position = restructure(
+    await readContract(client, {
+      ...parameters,
+      address: morpho,
+      abi: blueAbi,
+      functionName: "position",
+      args: [marketId, user],
+    }),
+    { abi: blueAbi, name: "position", args: ["0x", "0x"] },
```
Rubilmax, line 34: The thing that bothers me is that the dev has to pass dummy args to identify the abi item

Looking at viem's code, it makes sense: you have to make sure that the arg type matches the input types of the abi item, because 2 abi items can have the same name but not the same inputs

But it was non trivial for me to understand why this param is required and what to put in here when using it

Maybe defining dummy args exported from utils like this could help:
```ts
export const addressArg: Address = "0x";
export const hexArg: Hex = "0x";
export const bigintArg: bigint = 0n;
export const intArg: number = 0;
export const stringArg: string = "";
```

I built this reverse registry based on abitype's registry used by viem in their `AbiParameter` type

packages/blue-sdk-viem/src/fetch/Market.ts
```diff
@@ -62,38 +63,18 @@ export async function fetchMarket(
     }
   }
 
-  const [loanToken, collateralToken, oracle, irm, lltv] = await readContract(
-    client,
-    {
+  const params = new MarketParams(
```
Rubilmax, line 66: It makes me think we could improve speed of such queries for idle markets because `params` is only used afterwards to fetch the oracle price
Something for another PR I guess

packages/blue-sdk-viem/src/utils.ts
```diff
@@ -28,3 +44,156 @@ export const safeParseUnits = (strValue: string, decimals = 18) => {
...
+>(
+  client: Client<Transport, chain>,
+  parameters: ReadContractParameters<abi, name, args>,
+) {
+  const outputs = await readContract(client, parameters);
+
+  if (!isTuple(outputs)) {
+    throw new Error(
+      `Attempted to restructure non-tuple return values ${parameters.functionName} -> ${outputs}`,
+    );
```
Rubilmax, line 191: Should we throw actually? Or simply return `outputs`?
But then it complexifies the static types: we need to overload the function type to return outputs when abi item outputs is not an array... Seems do-able but so heavy it's not worth it for now

Rubilmax (dismissed):

oumar-fall (approved):

PR #346, opened 2025-05-21
feat(liquidation): fetch preliq data from indexer

Jean-Grimal:
Need tests


packages/liquidation-sdk-viem/src/preLiquidation/positionGetters.ts
```diff
@@ -4,69 +4,98 @@ import { fetchAccrualPosition } from "@morpho-org/blue-sdk-viem";
...
+      method: "POST",
+      body: JSON.stringify({ marketIds: whitelistedMarkets }),
+    });
 
-      const loanAsset = market !== null ? market[0]?.loanAsset : undefined;
-      const collateralAsset =
-        market !== null ? market[0]?.collateralAsset : undefined;
+    const data = (await response.json()) as {
+      preLiquidationData: PreLiquidationData[];
+    };
```
haydenshively, line 29: The response now includes both `warnings` and `results` keys -- you'll want to just access results here.

haydenshively (dismissed):
LGTM except for the one data type change. We can improve efficiency next week🙂

packages/liquidation-sdk-viem/src/preLiquidation/positionGetters.ts
```diff
@@ -4,69 +4,99 @@ import { fetchAccrualPosition } from "@morpho-org/blue-sdk-viem";
...
 export async function getPreLiquidablePositions(
   client: Client<Transport, Chain, Account>,
   whitelistedMarkets: MarketId[],
 ) {
   const chainId = client.chain.id;
 
-  const preLiquidations = (await preLiquidationLogs(client)).filter(
-    (preLiquidation) => whitelistedMarkets.includes(preLiquidation.marketId),
-  );
+  const url = `${process.env.INDEXER_API_URL}/chain/${chainId}/preliquidations`;
```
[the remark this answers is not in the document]
Rubilmax: Also, I would set it as parameter to this function instead of environment value because we are inside the liquidation SDK ; IMO sdk is only supposed to expose useful functions with all parameters, but is not expected to constrain the consumer to feed values via environment variables or via a database or a file etc
I would extract loading the value from the environment variable outside of this function and put it wherever we call this function, outside of the SDK (and still have a default in here)

packages/liquidation-sdk-viem/src/preLiquidation/positionGetters.ts
```diff
@@ -4,69 +4,99 @@ import { fetchAccrualPosition } from "@morpho-org/blue-sdk-viem";
...
+        }),
+    );
 
-  return preLiquidablePositions
-    .flat()
-    .filter((position) => position.preSeizableCollateral !== undefined);
+    return preLiquidablePositions
+      .flat()
+      .filter((position) => position.preSeizableCollateral !== undefined);
+  } catch (error) {
```
Rubilmax, line 96: There are many reasons why this execution would throw (RPC error, indexer error, API error, ...)
All of these paths are handled the same way: log the error to the console and return an empty array

This is an anti-pattern for the consumer: they cannot handle the different errors however they want

Instead, just let the function fail and let the consumer handle the errors (ideally document them in the jsdoc!)

packages/liquidation-sdk-viem/src/preLiquidation/positionGetters.ts
```diff
@@ -4,69 +4,99 @@ import { fetchAccrualPosition } from "@morpho-org/blue-sdk-viem";
...
+        }),
+    );
+
+    const preLiquidablePositions = await Promise.all(
+      preLiquidationInstances
+        .filter((position) => position !== undefined)
+        .map(async (preLiquidationPosition) => {
+          return await Promise.all(
+            preLiquidationPosition.enabledPositions.map(async (borrower) => {
+              return await getPreLiquidablePosition(
```
Rubilmax, line 81:
```suggestion
        .map((preLiquidationPosition) => Promise.all(
            preLiquidationPosition.enabledPositions.map((borrower) => getPreLiquidablePosition(
```

No need to async/await when you return a promise (it's overwhelming to review), but i'm not that opinionated on this)

packages/liquidation-sdk-viem/src/preLiquidation/positionGetters.ts
```diff
@@ -4,69 +4,99 @@ import { fetchAccrualPosition } from "@morpho-org/blue-sdk-viem";
...
+            },
+          };
+        })
+        .map(async (preLiquidation) => {
+          const {
+            markets: { items: market },
+          } = await apiSdk.getMarketAssets({
+            chainId,
+            marketId: preLiquidation.marketId,
+          });
```
Rubilmax, line 54: We're making N API calls here, where N is the nb of markets
Instead, group them all in a single API query to decrease the API and bandwidth consumption

Rubilmax (changes requested):

Rubilmax (approved):

PR #351, opened 2025-06-07
Fix/miscellaneous

Rubilmax:
Fixes INFRA-773, INFRA-775, INFRA-777, INFRA-778, INFRA-779, INFRA-805

### DIFF
--- a/packages/blue-sdk/src/addresses.ts
+++ b/packages/blue-sdk/src/addresses.ts
@@ -383,6 +383,20 @@ const _addressesRegistry = {
 
     wNative: "0x1aE9c40eCd2DD6ad5858E5430A556d7aff28A44b",
   },
+  [ChainId.KatanaMainnet]: {
+    morpho: "0xD50F2DffFd62f94Ee4AEd9ca05C61d0753268aBc",
+    bundler3: {
+      bundler3: "0xA8C5e23C9C0DF2b6fF716486c6bBEBB6661548C8",
+      generalAdapter1: "0x916Aa175C36E845db45fF6DDB886AE437d403B61",
+    },
+    adaptiveCurveIrm: "0x4F708C0ae7deD3d74736594C2109C2E3c065B428",
+    publicAllocator: "0x39EB6Da5e88194C82B13491Df2e8B3E213eD2412",
+    metaMorphoFactory: "0x1c8De6889acee12257899BFeAa2b7e534de32E16",
+    chainlinkOracleFactory: "0x7D047fB910Bc187C18C81a69E30Fa164f8c536eC",
+    preLiquidationFactory: "0x678EB53A3bB79111263f47B84989d16D81c36D85",
+
+    wNative: "0xEE7D8BCFb72bC1880D0Cf19822eB0A2e6577aB62",
+  },
 } as const;
 
 export type AddressLabel = DottedKeys<(typeof _addressesRegistry)[ChainId]>;
@@ -464,6 +478,9 @@ const _unwrappedTokensMapping: Record<number, Record<Address, Address>> = {
   },
   [ChainId.CampMainnet]: {
     [_addressesRegistry[ChainId.CampMainnet].wNative]: NATIVE_ADDRESS,
+  },
+  [ChainId.KatanaMainnet]: {
+    [_addressesRegistry[ChainId.KatanaMainnet].wNative]: NATIVE_ADDRESS,
   },
 };
 
--- a/packages/blue-sdk/src/chain.ts
+++ b/packages/blue-sdk/src/chain.ts
@@ -15,6 +15,7 @@ export enum ChainId {
   CornMainnet = 21000000,
   PlumeMainnet = 98866,
   CampMainnet = 123420001114,
+  KatanaMainnet = 747474,
 }
 
 export interface ChainMetadata {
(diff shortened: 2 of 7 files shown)


haydenshively (dismissed):

packages/simulation-sdk/src/SimulationState.ts
```diff
@@ -582,16 +582,22 @@ export class SimulationState implements InputSimulationState {
...
-
-            const suppliable =
-              cap -
+            const { cap, pendingCap, publicAllocatorConfig } =
+              data.getVaultMarketConfig(vault, marketId);
+
+            // If a pending cap is known to be valid, consider worst-case scenario
+            // where it is accepted before the reallocation is committed.
+            const validCap =
+              pendingCap.validAt >= data.block.timestamp
```
oumar-fall, line 591: 1. How could we have a `pendingCap.validAt` smaller than `data.block.timestamp` ?

2. Shouldn't we take the smaller of both? in case where the pending cap is higher but tx is included before cap gets accepted  we might have a revert

3. Is it that relevant to consider the pending cap if it gets valid after a long time? (we have a deadline of 2H for tx signatures while timelocks are more than 1 day)

packages/simulation-sdk/src/handlers/blue/borrow.ts
```diff
@@ -28,6 +28,9 @@ export const handleBlueBorrowOperation: OperationHandler<
       throw new BlueSimulationErrors.UnauthorizedBundler(onBehalf);
   }
 
+  if ((assets === 0n && shares === 0n) || (assets !== 0n && shares !== 0n))
```
oumar-fall, line 31:
```suggestion
  if ((assets === 0n) === (shares === 0n))
```

packages/simulation-sdk/src/handlers/blue/repay.ts
```diff
@@ -32,10 +34,25 @@ export const handleBlueRepayOperation: OperationHandler<
...
+    }
+
+    if (shares === maxUint256) {
+      shares = data.getPosition(onBehalf, id).borrowShares;
+
+      if (shares === 0n) throw new BlueSimulationErrors.ZeroShares();
+    }
+  }
+
+  if ((assets === 0n && shares === 0n) || (assets !== 0n && shares !== 0n))
```
oumar-fall, line 54: idem

packages/simulation-sdk/src/handlers/blue/supply.ts
```diff
@@ -33,10 +34,11 @@ export const handleBlueSupplyOperation: OperationHandler<
   const market = data.getMarket(id);
 
   // Simulate the bundler's behavior on supply.
-  if (sender === generalAdapter1 && assets === MathLib.MAX_UINT_256)
-    assets = data.getHolding(sender, market.params.loanToken).balance;
+  if (sender === generalAdapter1 && assets === maxUint256)
+    assets = data.getHolding(generalAdapter1, market.params.loanToken).balance;
 
-  if (assets === 0n && shares === 0n) throw new BlueErrors.InconsistentInput();
+  if ((assets === 0n && shares === 0n) || (assets !== 0n && shares !== 0n))
```
oumar-fall, line 40: idem

packages/simulation-sdk/src/operations.ts
```diff
@@ -301,11 +306,21 @@ export interface Erc20OperationArgs {
     spender: Address;
     amount: bigint;
     nonce: bigint;
+    /**
+     * The maximum block timestamp (included) for which the permit will be signed and valid.
+     * Defaults to the simulation state's timestamp + 2h.
+     */
+    deadline?: bigint;
```
oumar-fall, line 313: wouldn't it be more convenient to only ask for the validity of the signature (default to 2H) so that we don't need to check

oumar-fall: this could lead to an overflow though if we don't take care

packages/simulation-sdk/src/SimulationState.ts
mario-eth: agree here, should be `min(pendingCap, cap)`
furthermore if `validAt  =< block.timestamp + 2h` then consider this

Rubilmax: I would still keep `validAt <= block.timestamp` to be conservative and because it's not an issue if shared liquidity is under-estimated in this specific case

packages/simulation-sdk/src/operations.ts
Rubilmax: But as a consumer i do want to set an absolute deadline so my simulation reverts for example if my bundle is too old (example: i leave my interface open for 2h and state did not update for some reason)

oumar-fall: My suggestion was just to let the user define a duration instead of the deadline:
so that the deadline would be computed in the sdk as : `deadline = Time.timestamp() + duration` instead of directly specifying the `deadline`. This would save us from having to check that `deadline >= Time.timestamp()` everywhere.

oumar-fall: This is not that important

haydenshively (approved):

oumar-fall (approved):

PR #352, opened 2025-06-10
fix(addresses): add katana

Rubilmax:


0xbulma (commented):
If I am not mistaken. The following object needs to be updated as well. 
`const _unwrappedTokensMapping: Record<number, Record<Address, Address>> = {`

PR #378, opened 2025-07-23
fix(erc4626Deposit): update vault.toShares rounding to "Down"

Foulks-Plb:
The share calculation on deposit should be rounded down.


Rubilmax (commented):
Why?

It seems to me the corresponding amount of `shares` that yields `assets` should be rounded up, because the amount of `assets` that corresponds to `shares` is rounded down

PR #379, opened 2025-07-24
fix(paraswap): skim possible paraswap surplus

Rubilmax:


packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -915,40 +914,89 @@ export const finalizeBundle = (
...
+            address,
+            sender: generalAdapter1,
+            args: {
+              amount: maxUint256,
+              from: generalAdapter1,
+              to: receiver,
+            },
+          }) as const,
+      ),
+  );
```
oumar-fall, line 948: Equivalent but you could push every addresses in a set here and at the end push all the skims based on the Set's content

packages/bundler-sdk-viem/src/operations.ts
```diff
@@ -915,40 +914,89 @@ export const finalizeBundle = (
...
+
+  const uniqueSkimTokens = new Set(skims.map(({ address }) => address));
+  const pushCustomSkim = (operation: UnionOmit<BundlerOperation, "sender">) => {
+    // Paraswap does not guarantee that the amount effectively bought (resp. sold) corresponds to
+    // the requested amount to buy (resp. sell), so we force skim the possible surplus of bought (resp. sold) token.
+    switch (operation.type) {
+      case "Paraswap_Buy":
+      case "Paraswap_Sell":
+        if (!uniqueSkimTokens.has(operation.address)) {
```
oumar-fall, line 957: do we have the guarenty that operation.address is the sold/bought token?

oumar-fall (dismissed):

Foulks-Plb (dismissed):

PR #387, opened 2025-08-12
Feat/vault v2 class

oumar-fall:


packages/blue-sdk-viem/src/fetch/v2/VaultV2.ts
```diff
@@ -0,0 +1,162 @@
...
+  parameters: DeploylessFetchParameters = {},
+) {
+  parameters.chainId ??= await getChainId(client);
+
+  const vaultV2 = await fetchVaultV2(address, client, parameters);
+
+  const [assetBalance, ...adapters] = await Promise.all([
+    readContract(client, {
+      ...parameters,
+      address: vaultV2.asset,
```
Rubilmax, line 151: It would require 1 less RPC call to first fetch the vault's asset and adapters count and then parallelize the rest (2 calls vs 3 currently)

But I guess we don't really care as of now while implementing this is not trivial

packages/blue-sdk-viem/src/fetch/v2/VaultV2Adapter.ts
```diff
@@ -0,0 +1,78 @@
...
+    address: vaultV2MorphoVaultV1AdapterFactory,
+    abi: vaultV2MorphoVaultV1AdapterFactoryAbi,
+    functionName: "isMorphoVaultV1Adapter",
+    args: [address],
+  });
+
+  if (isMorphoVaultV1Adapter)
+    return fetchVaultV2MorphoVaultV1Adapter(address, client, parameters);
+
+  throw "Unknown adapter type";
```
Rubilmax, line 44: Throw a dedicated error class to catch it more easily

packages/blue-sdk-viem/src/fetch/v2/VaultV2MorphoVaultV1Adapter.ts
```diff
@@ -0,0 +1,81 @@
...
+  client: Client,
+  parameters: DeploylessFetchParameters = {},
+) {
+  const adapter = await fetchVaultV2MorphoVaultV1Adapter(
+    address,
+    client,
+    parameters,
+  );
+  const [vaultV1, shares] = await Promise.all([
+    fetchAccrualVault(adapter.morphoVaultV1, client, parameters),
```
Rubilmax, line 70: Fetching an accrual vault v2 becomes ultra expensive in terms of RPC. Deploy less reads will be very interesting in this case I believe

packages/blue-sdk/src/addresses.ts
```diff
@@ -200,6 +204,10 @@ const _addressesRegistry = {
     usdc: "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
     verUsdc: "0x59aaF835D34b1E3dF2170e4872B785f11E2a964b",
     testUsdc: "0xBC77067f829979812d795d516E523C4033b66409",
+
+    vaultV2Factory: "0xc2f730a67ae81a7d2c17268a9652338449bf264b",
```
Rubilmax, line 208: Addresses seem not checksummed

packages/blue-sdk/src/errors.ts
```diff
@@ -41,6 +41,12 @@ export class UnsupportedPreLiquidationParamsError extends Error {
   }
 }
 
+export class UnsupportedVaultV2AdapterError extends Error {
+  constructor(public readonly address: Address) {
+    super(`type of adapter ${address} is not supported`);
```
Rubilmax, line 46:
```suggestion
    super(`vault v2 adapter "${address}" is not supported`);
```

packages/bundler-sdk-viem/src/types/operations.ts
```diff
@@ -99,33 +101,54 @@ export type Erc20BundlerOperations = {
 export type Erc20BundlerOperation =
   Erc20BundlerOperations[Erc20BundlerOperationType];
 
+export const VAULTV2_BUNDLER_OPERATIONS = [
```
Rubilmax, line 104:
```suggestion
export const VAULT_V2_BUNDLER_OPERATIONS = [
```

Because we write `VaultV2` 😬😬

packages/simulation-sdk-wagmi/src/hooks/useSimulationState.ts
```diff
@@ -249,9 +262,45 @@ export function useSimulationState<
...
+      ...parameters.query,
+      enabled: block != null && parameters.query?.enabled,
+    },
+  });
+
   const data = useMemo(() => {
     if (block == null) return;
 
+    for (const token of values(tokens.data)) {
+      if (!token) continue;
```
Rubilmax, line 287: Can we check `token == null`? I know most of the time it's ok but the fact that js also considers empty string as false is scary to me so if we only need to check it's not null I feel safer relying on pure checks

packages/simulation-sdk/package.json
```diff
@@ -18,6 +18,8 @@
     "build": "tsc --noEmit && tsc --build tsconfig.build.json"
   },
   "dependencies": {
+    "@types/lodash.merge": "^4.6.9",
+    "lodash.merge": "^4.6.2",
```
Rubilmax, line 22: Why?

Rubilmax: Also types should only be dev

packages/simulation-sdk-wagmi/src/hooks/useSimulationState.ts
```diff
@@ -249,9 +262,45 @@ export function useSimulationState<
...
+      const vaultV2 = vaultV2s.data[token.address];
+      if (!vaultV2) continue;
+
+      const eip5267Domain = new Eip5267Domain({
```
Rubilmax, line 291:
```suggestion
      // Vaults V2 are not EIP-5267 compliant so we hardcode their EIP-712 domain specification to support EIP-2612 permits.
      const eip5267Domain = new Eip5267Domain({
```

packages/simulation-sdk/package.json
oumar-fall: I use it in `simulation-sdk` test fixtures so both can be in dev

PR #393, opened 2025-08-26
fix: correct customizer typing in registerCustomAddresses

bigint:
- Update `customizer` to use generic `<T>` with proper `string | number | symbol` key type  
- Return `undefined` explicitly for compatibility with lodash `mergeWith`  
- Ensures stricter type-safety when merging custom addresses, deployments, and unwrapped tokens


Rubilmax (approved):
The value of this type is not much but lgtm

PR #395, opened 2025-08-26
chore: make _get generic and type-safe when indexing by dotted paths

bigint:
- Convert internal `_get` helper to a generic function with safe indexing  
- Preserve null/undefined semantics and existing public API (`getValue`, `hasValue`)  
- Improves type safety without changing runtime behaviour


Rubilmax:
This is not substantially more type-safe so I'm closing

PR #424, opened 2025-10-01
Feat/vault v2 maxdeposit withdraw

Rubilmax:


packages/blue-sdk/src/market/Market.ts
oumar-fall, line 13: Why is this in Market.ts if it is generic?

packages/blue-sdk-viem/contracts/vault-v2/GetVaultV2.sol
```diff
@@ -21,14 +22,20 @@ struct VaultV2Response {
     uint64 lastUpdate;
     address[] adapters;
     address liquidityAdapter;
+    bool isLiquidityAdapterKnown;
+    Caps liquidityCaps;
```
oumar-fall, line 26: This is not future-proof as the liquidity adapter could have several ids (eg. marketV1Adapter) and thus `liquidityCaps` should be a mapping and you should return the ids list (as you would need to compute them anyways)

packages/blue-sdk/src/vault/v2/VaultV2MorphoVaultV1Adapter.ts
```diff
@@ -20,7 +20,16 @@ export class VaultV2MorphoVaultV1Adapter
...
   }: IVaultV2MorphoVaultV1Adapter) {
-    super(vaultV2Adapter);
+    super({
+      ...vaultV2Adapter,
+      adapterId: keccak256(
+        encodeAbiParameters(
+          [{ type: "string" }, { type: "address" }],
+          ["this", vaultV2Adapter.address],
+        ),
+      ),
```
oumar-fall, line 30: Maybe export it to a function

packages/blue-sdk/src/vault/v2/VaultV2.ts
```diff
@@ -114,19 +129,86 @@ export interface IAccrualVaultV2 extends Omit<IVaultV2, "adapters"> {}
...
+    if (this.liquidityAdapter === zeroAddress)
+      return {
+        value: this.assetBalance,
+        limiter: CapacityLimitReason.liquidity,
+      };
+
+    if (
+      this.accrualLiquidityAdapter instanceof AccrualVaultV2MorphoVaultV1Adapter
+    ) {
+      return this.accrualLiquidityAdapter.accrualVaultV1.maxWithdraw(shares);
```
oumar-fall, line 198: maxWithdraw should be caped by vault's `totalAssets`

oumar-fall (changes requested):

Foulks-Plb (approved):

PR #428, opened 2025-10-08
feat(addresses): add USDC address to the registry for hyperliquid EVM

Foulks-Plb:


packages/blue-sdk/src/addresses.ts
```diff
@@ -444,8 +444,9 @@ const _addressesRegistry = {
     registryList: "0x857B55cEb57dA0C2A83EE08a8dB529B931089aee",
     chainlinkOracleFactory: "0xeb476f124FaD625178759d13557A72394A6f9aF5",
     preLiquidationFactory: "0x1b6782Ac7A859503cE953FBf4736311CC335B8f0",
-
```
oumar-fall, line 447: Don't remove the space, it's there to split "contracts" and tokens

oumar-fall (approved):

PR #451, opened 2025-10-22
chore: fix vault v2 cache type policies

alexbensimon:


Rubilmax (approved):
There will be the adapter to add some day

PR #455, opened 2025-10-31
fix(deepFreeze): handle null & undefined value

Foulks-Plb:
The deepFreeze function was throwing a TypeError: Cannot convert undefined or null to object when called with null or undefined values. This occurred because `Object.getOwnPropertyNames()` cannot be called on primitive values.

Added a guard clause at the beginning of deepFreeze to handle null and undefined values. Since these primitives are already immutable in JavaScript, the function now returns them directly without attempting to freeze them.

### DIFF
--- a/packages/blue-api-sdk/src/types.ts
+++ b/packages/blue-api-sdk/src/types.ts
@@ -1855,12 +1855,15 @@ export type OracleVault = {
   address: Scalars["Address"]["output"];
   assetId: Maybe<Scalars["String"]["output"]>;
   chain: Chain;
+  /** @deprecated No longer maintained. */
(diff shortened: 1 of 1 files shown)


Rubilmax (approved):

packages/morpho-ts/src/utils.ts
```diff
@@ -154,6 +154,11 @@ export function getLastDefined<T>(array: T[]) {
 }
 
 export function deepFreeze<T>(obj: T): T {
+  if (obj === null || obj === undefined) {
```
oumar-fall:
```suggestion
  if (obj == null) {
```

equivalent

PR #527, opened 2026-03-13
fix(bundler-sdk): rename slippageAmount to correct parameter names

Kubudak90:
This PR fixes #470 by renaming the `slippageAmount` parameter in the Bundler SDK to the correct parameter names that match the Solidity contract.

## Changes

Updated the following functions in `BundlerAction.ts`:

- `morphoSupply`: `slippageAmount` → `maxSharePriceE27`
- `morphoBorrow`: `slippageAmount` → `minSharePriceE27`  
- `morphoRepay`: `slippageAmount` → `maxSharePriceE27`
- `morphoWithdraw`: `slippageAmount` → `minSharePriceE27`

Also updated the JSDoc comments to clarify that these parameters represent share prices scaled by 1e27, not slippage amounts.

This makes the SDK API consistent with the underlying Solidity contract and helps developers understand the actual meaning of these parameters.

### DIFF
--- a/packages/bundler-sdk-viem/src/BundlerAction.ts
+++ b/packages/bundler-sdk-viem/src/BundlerAction.ts
@@ -1091,7 +1091,7 @@ export namespace BundlerAction {
    * @param market The market params to supply to.
    * @param assets The amount of assets to supply.
    * @param shares The amount of supply shares to mint.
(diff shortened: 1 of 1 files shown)


Rubilmax:
Hi thanks for your submission, this package is being deprecated

PR #548, opened 2026-04-21
fix(blue-sdk): throw UnknownMarketAllocation on stale withdrawQueue

Kubudak90:
## Problem

The `AccrualVault.accrueInterest()` method uses a non-null assertion (`!`) when looking up allocations by market ID from the withdrawQueue:

```typescript
const { config, position } = this.allocations.get(marketId)!;
```

This is unsafe because the `withdrawQueue` may contain market IDs that are not present in `allocations`. For example, after a market is removed from a vault's allocations but is still referenced in the withdraw queue, this would cause a runtime crash.

## Solution

Replace the non-null assertion with a safe lookup and filter out any missing allocations:

```typescript
const allocation = this.allocations.get(marketId);
if (!allocation) return null;
// ...
```

This preserves the withdraw queue order while gracefully handling missing markets.

## Changes

- `packages/blue-sdk/src/vault/Vault.ts`: Safe allocation lookup with null filter
- `packages/blue-sdk/test/unit/Vault.test.ts`: Regression test for missing allocation in withdrawQueue

## Testing

```bash
npx vitest --run --project blue-sdk test/unit/Vault.test.ts
```

All tests pass (2/2).


Rubilmax:
@prd-carapulse[bot] review this PR

packages/blue-sdk/src/vault/Vault.ts
```diff
@@ -426,14 +425,21 @@ export class AccrualVault extends Vault implements IAccrualVault {
...
-
-        return {
-          config,
-          position: position.accrueInterest(timestamp),
-        };
-      }),
+      this.withdrawQueue
+        .map((marketId) => {
+          const allocation = this.allocations.get(marketId);
+          if (!allocation) return null;
```
Rubilmax: IMO we should throw a custom error `UnknownMarketAllocation` here instead of silently not accruing so we ensure the returned vault is 100% accurate

Rubilmax (commented):
This PR is worth implementing! We just have a different POV regarding missing allocations: we'd prefer loudly failing instead of silently ignoring missing markets

Kubudak90:
Switched to the fail-loud approach per @Rubilmax's feedback. The non-null assertion now becomes an explicit `throw new UnknownMarketAllocation(marketId)` instead of a silent skip + filter, so the returned vault is always 100% accurate (or the caller gets a typed error).

Summary of changes (single commit, rebased onto current `main`):

- `packages/blue-sdk/src/errors.ts` — new `UnknownMarketAllocation extends UnknownDataError`, mirroring the existing `Unknown*` error pattern. Callers that want graceful degradation can opt in via `_try(() => vault.accrueInterest(), UnknownDataError)`.
- `packages/blue-sdk/src/vault/Vault.ts` — `accrueInterest()` now throws `UnknownMarketAllocation` if `withdrawQueue` references a market that isn't in `allocations`, instead of crashing with the opaque `Cannot destructure property 'config' of 'undefined'`.
- `packages/blue-sdk/src/vault/Vault.test.ts` — added a regression test using the project's existing fixture helpers (`accrualVault`, `vaultInput`). The test mutates `withdrawQueue` after construction to introduce the stale entry (the only way to reach this branch, since the `AccrualVault` constructor rebuilds `withdrawQueue` from `allocations`) and asserts `.toThrow(UnknownMarketAllocation)`.

Addresses both review comments:
- @Rubilmax: silent skip → loud fail with a typed error.
- @prd-carapulse: build is green locally (`pnpm --filter @morpho-org/blue-sdk run build` ✅), and the test exercises the actual bug path by mutating `withdrawQueue` after construction; the new `expect(...).toThrow(UnknownMarketAllocation)` assertion would fail on the unfixed code (the `!` assertion throws `TypeError: Cannot destructure property 'config' of 'undefined'`, not `UnknownMarketAllocation`).

Local verification:
- `pnpm --filter @morpho-org/blue-sdk run build` → ✅
- `pnpm exec vitest --project blue-sdk --run --exclude '**/e2e/**'` → 333/333 passing
- `pnpm exec biome check packages/blue-sdk/src/{errors.ts,vault/Vault.ts,vault/Vault.test.ts}` → clean

Ready for re-review.

packages/blue-sdk/src/vault/Vault.test.ts
```diff
@@ -272,4 +273,16 @@ describe("AccrualVault", () => {
...
     expect(accrued.totalAssets).toBeGreaterThanOrEqual(accrued.lastTotalAssets);
   });
+
+  test("accrueInterest throws UnknownMarketAllocation when withdrawQueue references a market not in allocations", () => {
+    const vault = accrualVault();
+    const ghostMarketId = "0x1234" as MarketId;
+    // Mutate the public `withdrawQueue` after construction so it references a
+    // market that has no corresponding entry in `allocations`. Without this
+    // mutation the bug cannot be reached, because the AccrualVault constructor
+    // rebuilds `withdrawQueue` from the allocations argument.
```
Rubilmax, line 283: This alone makes the bug unreachable with standard behavior so in the end is it worth it? Sorry for the back-and-forth

WDYT @Foulks-Plb ?

Foulks-Plb: IMO, readonly + freeze ?

Kubudak90: Implemented this in `ffff390`: `withdrawQueue` is now typed as `readonly`, stored as a frozen copy, and defined as non-writable on `Vault`. I also removed the `UnknownMarketAllocation` branch/test since the stale queue state is no longer reachable through the public API.

Verified:
- `pnpm exec vitest --project blue-sdk --run packages/blue-sdk/src/vault/Vault.test.ts`
- `pnpm --filter @morpho-org/blue-sdk run build`
- `pnpm exec vitest --project blue-sdk --run --exclude '**/e2e/**'`
- `pnpm exec biome check packages/blue-sdk/src/errors.ts packages/blue-sdk/src/vault/Vault.ts packages/blue-sdk/src/vault/Vault.test.ts`

Rubilmax: Wait this went too fast. It is the first time we introduce frozen fields in class instances, i am against because it limits the possibilities for integrators while there is not transaction information stored. If the consumer mutates this field, they are responsible for edge cases that are implied. I would keep this field not readonly and not frozen and instead implement your PR as it was (ie revert your last commit)

Kubudak90: Reverted the freeze/readonly change in `985d4fd`, so the PR is back to the previous custom-error approach: `withdrawQueue` remains mutable, and `accrueInterest()` throws `UnknownMarketAllocation` if a mutated queue references a market missing from `allocations`.

Verified after the revert:
- `corepack pnpm exec vitest --project blue-sdk --run packages/blue-sdk/src/vault/Vault.test.ts`
- `corepack pnpm --filter @morpho-org/blue-sdk run build`
- `corepack pnpm exec vitest --project blue-sdk --run --exclude '**/e2e/**'`
- `corepack pnpm exec biome check packages/blue-sdk/src/errors.ts packages/blue-sdk/src/vault/Vault.ts packages/blue-sdk/src/vault/Vault.test.ts`

Rubilmax (dismissed):

Rubilmax:
@codex

Rubilmax:
@prd-carapulse[bot] review this PR

packages/blue-sdk/src/vault/Vault.ts
```diff
@@ -427,8 +428,15 @@ export class AccrualVault extends Vault implements IAccrualVault {
...
       this.withdrawQueue.map((marketId) => {
-        const { config, position } = this.allocations.get(marketId)!;
-
+        const allocation = this.allocations.get(marketId);
+        // Fail loudly rather than silently dropping the market: a stale
+        // `withdrawQueue` entry (e.g., one mutated after construction to
+        // reference a market that is no longer allocated) would otherwise
+        // crash with an opaque "Cannot destructure property 'config' of
+        // 'undefined'" via the non-null assertion below.
+        if (allocation == null) throw new UnknownMarketAllocation(marketId);
```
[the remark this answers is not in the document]
Rubilmax: indeed, missing patch changeset

packages/blue-sdk/src/errors.ts
```diff
@@ -33,6 +33,12 @@ export class UnknownVaultConfigError extends UnknownDataError {
   }
 }
 
+export class UnknownMarketAllocation extends UnknownDataError {
```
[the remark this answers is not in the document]
Kubudak90: Added JSDoc for the exported error in `bdbc58ee`, aligned with the JSDoc-covered `errors.ts` from current `main`.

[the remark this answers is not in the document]
Kubudak90: Renamed the public error to `UnknownMarketAllocationError` in `bdbc58ee` and updated the `Vault.ts` throw path plus the regression test import/assertion.

packages/blue-sdk/src/vault/Vault.ts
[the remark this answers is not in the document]
Kubudak90: Added `.changeset/vault-accrual-error.md` in `bdbc58ee` as a patch changeset for `@morpho-org/blue-sdk`. `corepack pnpm changeset status --since origin/main` now reports the expected patch bump.

Rubilmax (approved):

Rubilmax:
@codex

Rubilmax:
@prd-carapulse[bot] re-review

Foulks-Plb (approved):

Rubilmax:
@Kubudak90 Thanks for your iterations, we required commits to have verified signatures to go further.

Kubudak90:
Rewrote the branch to a single GitHub-verified signed commit: `25ea528187ad04546f5e64e57f0afb40a1df4186` (`verified=true`, `reason=valid`).

The resulting tree is unchanged from the previously reviewed final diff, just with the unsigned/reverted intermediate commits removed. Aikido is passing on the new head.

GitHub still shows the PR as blocked because the required Actions contexts `lint / biome`, `test / vitest`, and `build / Build` are not present on the fork PR head; the only check currently reported on the new commit is `Aikido Security: check code`.

Rubilmax (approved):

Rubilmax:
Superseded by #783.

PR #552, opened 2026-04-23
fix(migration-sdk): residual sweep + cleanup repay with skipRevert=false

Foulks-Plb:
## Summary

Reapplies the v1 of [PR #551][pr551] (commit `84707d2`) on top of `main`, with **two deliberate changes**: split `skipRevert` flags + add a final skim from `generalAdapter1` to the user.

For every Aave V2 / Aave V3 / Compound V3 borrow migration, after the source repay leg, we now unconditionally append three actions:

1. `erc20Transfer(loanToken, generalAdapter1, maxUint256, <migrationAdapter>, skipRevert=false)` — sweep residual destination loan tokens off the public migration adapter back to `generalAdapter1`.
2. `morphoRepay(marketTo, maxUint256, 0n, maxUint256, user, [], skipRevert=true)` — apply any recovered residual to the destination market on behalf of the user.
3. `erc20Transfer(loanToken, user, maxUint256, generalAdapter1, skipRevert=false)` — drain any leftover loan tokens from `generalAdapter1` back to the user.

| Action | `skipRevert` | Why |
| --- | --- | --- |
| Sweep migration adapter -> `generalAdapter1` | `false` | Gracieux on happy path. `CoreAdapter.erc20Transfer` resolves `maxUint256` to `balanceOf(self)`, the `ZeroAmount` guard sits in the `else` branch, and `if (amount > 0)` skips the transfer when the balance is zero. No revert. |
| `morphoRepay(maxUint256, maxUint256)` cleanup | `true` | Not gracieux. `GeneralAdapter1.morphoRepay` reverts with `ZeroAmount` when its loan-token balance is zero, and Morpho enforces `(assets == 0) != (shares == 0)`. Without `skipRevert`, the bundle reverts on every happy-path migration. |
| Final skim `generalAdapter1` -> user | `false` | Same gracieux behaviour as the first sweep (`maxUint256` -> balance -> `if (amount > 0)`); no revert when balance is already zero. |

## Why the final skim

If the cleanup `morphoRepay` no-ops via `skipRevert: true` (e.g. there is no destination borrow position to repay against, or Morpho reverts for any reason), the residual stays on `generalAdapter1`. Since `generalAdapter1` inherits the same `erc20Transfer(maxUint256)` entrypoint from `CoreAdapter` as the protocol-specific migration adapters, that residual is sweepable by any later Bundler3 caller — i.e. the SDK-155 / MORP2-4 stranding vulnerability would just be moved from one public adapter to another.

The final skim closes that gap by always returning anything left on `generalAdapter1` to the user. It is itself gracieux (`maxUint256`), so it never reverts on the happy path where the cleanup repay consumed everything.

## Why split the `skipRevert` flags

`skipRevert: true` swallows any revert on an action, not just the well-defined no-op cases. We want it as narrow as possible:

- The two `erc20Transfer(maxUint256)` calls are gracieux by design at the CoreAdapter level, so they stay strict (`false`) — a paused / blocked / blacklisted loan token reverting on a transfer will surface a real failure.
- Only the cleanup `morphoRepay` needs `skipRevert: true`, because there is no `maxUint256` resolution path on the GeneralAdapter side that turns "nothing to repay" into a clean no-op.

The proper fix that avoids `skipRevert` entirely is the v2 of #551 (overborrow `slippageFrom` on `morphoBorrow` + cap source repay at `migratedBorrow`, gated on `slippageFrom > 0n`); this PR is the minimal-diff alternative.

## Tests

Test assertions in `test/e2e/{aaveV2,aaveV3,compoundV3}/borrow.test.ts` are updated to reflect the new three-action suffix on every borrow path.

Forked-mainnet tests are not run locally; CI will confirm.

## Tracking

- Linear: [SDK-155][sdk-155]
- Reference: [PR #551][pr551] (v1 commit `84707d2`, v2 commit `1113842`)

[pr551]: https://github.com/morpho-org/sdks/pull/551
[sdk-155]: https://linear.app/morpho-labs/issue/SDK-155/morp2-4low-aave-and-compound-borrow-migrations-can-strand-recoverable


packages/migration-sdk-viem/src/positions/borrow/aaveV2.borrow.ts
```diff
@@ -255,26 +255,23 @@ export class MigratableBorrowPosition_AaveV2
...
+                marketTo.loanToken,
+                generalAdapter1,
+                maxUint256,
+                aaveV2MigrationAdapter,
+                false,
+              ],
+            },
+            {
+              type: "morphoRepay",
+              args: [marketTo, maxUint256, 0n, maxUint256, user, [], true],
```
Foulks-Plb, line 270: Skipping the revert here is dangerous.

Foulks-Plb: It’s necessary because if the adapter migration balance is 0, it reverts.

Foulks-Plb: I’ll add another skim from the General Adapter to the user in case the repay fails for another reason and there are residual funds left on the migration adapter.

Foulks-Plb:
@codex review

Rubilmax (approved):

alexbensimon (approved):

Rubilmax (approved):

PR #553, opened 2026-04-27
Set up Changesets releases

Rubilmax:
## Summary
- add Changesets config and root release scripts for the pnpm workspace
- replace the custom per-package publish matrix with a workspace-level Changesets release workflow
- baseline package versions to the current npm `latest` dist-tags and remove the old custom release scripts
- simplify build CI to use the root workspace build command

Fixes: SDK-172

## Notes
- version commits are created directly on `main`/`next` through the GitHub App token using GitHub GraphQL `createCommitOnBranch`
- `next` enters Changesets prerelease mode when pending changesets exist and no `.changeset/pre.json` is present
- added a `read-yaml-file` override so Changesets works with the repository's existing `js-yaml` override

## Verification
- `HUSKY=0 pnpm install --frozen-lockfile`
- `pnpm lint`
- `pnpm build`
- `pnpm exec vitest --run packages/morpho-ts packages/test-wagmi`
- disposable Changesets smoke test for consuming a tracked changeset and bumping only `@morpho-org/morpho-ts`

Full `pnpm test --run` was attempted locally but hit public RPC rate limits / Cloudflare 429 from `eth.merkle.io`.

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/553" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


Rubilmax:
@codex

Rubilmax:
@codex

Rubilmax:
@codex

Rubilmax:
@codex

Foulks-Plb:
**CODEOWNERS coverage is too narrow on workflows**

Right now only `publish.yml` is owned by `@morpho-org/security`. The other workflows (`push.yml`, `build.yml`, `lint.yml`, `test.yml`) orchestrate what runs before `publish.yml` — for example, a PR could edit `push.yml` to remove `needs: [lint, build, test]` or change the `if` gate, and it would only need 2 standard reviews to land, with no security approval.

Could we extend CODEOWNERS to the whole workflows directory?

```
/.github/workflows/  @morpho-org/security
```

.github/workflows/publish.yml
```diff
@@ -4,44 +4,30 @@ on:
   workflow_call:
```
Foulks-Plb, line 4: IMO, 
jobs:
  publish:
    if: github.ref_name == 'main' || github.ref_name == 'next'

Rubilmax:
@codex

.github/workflows/publish.yml
```diff
@@ -4,44 +4,31 @@ on:
...
+      CI: true
+      HUSKY: 0
 
     steps:
+      - name: Generate GitHub App Token
+        id: app-token
+        uses: actions/create-github-app-token@d72941d797fd3113feb6b93fd0dec494b13a2547 # v1
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
```
Foulks-Plb, line 29: still use ?

Foulks-Plb: validated together

.github/workflows/publish.yml
Foulks-Plb, line 1: Maybe add concurrency with cancel-in-progress: false ?

Foulks-Plb (dismissed):

.github/CODEOWNERS
```diff
@@ -1,2 +1,3 @@
 /.github/CODEOWNERS @morpho-org/security
 /.github/workflows/publish.yml @morpho-org/security
```
0xbulma: **[LOW]** Line 2 (`/.github/workflows/publish.yml @morpho-org/security`) is now fully subsumed by the new line 3 (`/.github/workflows/ @morpho-org/security`).

Suggestion: drop the redundant line so future edits can't drift between the two.

.github/workflows/publish.yml
```diff
@@ -59,23 +50,165 @@ jobs:
...
+            printf '%s\n' "${pending[@]}"
+          else
+            echo "has_changesets=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Enter next prerelease mode
+        if: github.ref_name == 'next' && steps.pending-changesets.outputs.has_changesets == 'true' && hashFiles('.changeset/pre.json') == ''
+        run: pnpm changeset pre enter next
+
+      - name: Version packages
```
0xbulma, line 84: **[MEDIUM]** If `.changeset/pre.json` ever lands on `main` (accidental cherry-pick from `next`, failed `pre exit`, manual edit), `pnpm run version` here will silently produce prerelease versions on `main` — the workflow only ever *enters* prerelease mode (line 74-76, gated on `next`) and never exits it.

Suggestion: add a guard before `Version packages` that fails the job when `github.ref_name == 'main' && hashFiles('.changeset/pre.json') != ''`, or auto-run `pnpm changeset pre exit` on `main` when `pre.json` is present.

0xbulma (commented):

**Reviewed commit:** `ef52e51`

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 1 |
| Medium | 4 |
| Low | 8 |

Reviewed by 5 parallel agents (code quality & security, silent failures, docs, test coverage, simplification). Web3 Security agent skipped — no contract code touched.

_Automated review. Will re-run when new commits are pushed._

.github/workflows/publish.yml
```diff
@@ -59,23 +50,165 @@ jobs:
...
+        if: steps.pending-changesets.outputs.has_changesets == 'true'
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - run: echo "- [RELEASE] ${{ steps.release.outputs.version }}" >> $GITHUB_STEP_SUMMARY
-        if: steps.release.outputs.version
+          BRANCH: ${{ github.ref_name }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          node <<'NODE'
```
0xbulma, line 94: **[MEDIUM]** ~107 lines of inline Node.js embedded in YAML (lines 88–194) cannot be linted, type-checked, or unit-tested. The previous release plumbing in `scripts/release/*.js` was extracted for exactly this reason; putting Node back inside YAML is a maintainability regression.

Suggestion: extract to `scripts/release/commit-version.js` and invoke via `node scripts/release/commit-version.js`. This restores typechecking/linting and makes the rename-handling parser branch (lines 117-127) and additions/deletions construction testable.

.github/workflows/publish.yml
```diff
@@ -59,23 +50,165 @@ jobs:
...
+          const { execFileSync } = require("node:child_process");
+          const { readFileSync } = require("node:fs");
+
+          const main = async () => {
+            const exec = (command, args) =>
+              execFileSync(command, args, { encoding: "utf8" }).trim();
+
+            execFileSync("git", ["add", "."]);
```
0xbulma, line 102: **[MEDIUM]** `git add .` stages every non-gitignored file in the workspace, not just the changeset version output. If a prior step (husky hook, future tooling, generated artifact) writes a non-gitignored file before this point, it gets folded into the GraphQL "version packages" commit and pushed to `main`/`next`.

Suggestion: stage only the paths Changesets is known to write — e.g. `git add packages/*/package.json packages/*/CHANGELOG.md .changeset/` — or `git add -u` plus an explicit allowlist for newly created CHANGELOG.md / `.changeset/pre.json` state.

.github/workflows/publish.yml
```diff
@@ -59,23 +50,165 @@ jobs:
...
+            }
+
+            const fields = diff.toString("utf8").split("\0").filter(Boolean);
+            const additions = [];
+            const deletions = [];
+
+            for (let i = 0; i < fields.length; ) {
+              const status = fields[i++];
+
+              if (status.startsWith("R")) {
```
0xbulma: **[LOW]** The `git diff --name-status -z` parser handles `R` (rename) but not `C` (copy). Copies share the same 3-token layout (`C<score>\0<oldpath>\0<newpath>`) and would be misread as a 2-token entry by the `else` branch, corrupting the additions/deletions arrays. `changeset version` is unlikely to produce copies, but it's a one-line fix.

Suggestion: `if (status.startsWith("R") || status.startsWith("C"))`.

.github/workflows/publish.yml
```diff
@@ -59,23 +50,165 @@ jobs:
...
+              }),
+            });
+
+            const result = await response.json();
+            if (!response.ok || result.errors) {
+              console.error(JSON.stringify(result, null, 2));
+              process.exit(1);
+            }
+
+            const commit = result.data.createCommitOnBranch.commit;
```
0xbulma, line 199: **[LOW]** `result.data.createCommitOnBranch.commit` will throw `Cannot read properties of null` if GitHub returns `data: { createCommitOnBranch: null }` without populating `errors` (rare branch-protection edge case — e.g., the API silently drops the commit). The thrown error is caught by the outer `.catch` and exits non-zero, so it's noisy not silent, but the surfaced message is a stack trace instead of the API response.

Suggestion: `if (!result.data?.createCommitOnBranch?.commit) { console.error(JSON.stringify(result, null, 2)); process.exit(1); }`.

.github/workflows/publish.yml
```diff
@@ -59,23 +50,165 @@ jobs:
...
+      - name: Publish packages
+        if: steps.pending-changesets.outputs.has_changesets == 'false'
+        shell: bash
+        run: |
+          publish_tag="latest"
+          if [ "${GITHUB_REF_NAME}" = "next" ]; then
+            publish_tag="next"
+          fi
+
+          git tag --list | sort > /tmp/tags-before
```
0xbulma: **[LOW]** The before/after tag-diff using hardcoded `/tmp/tags-before`, `/tmp/tags-after`, `/tmp/new-tags` is fragile (no `mktemp`, no cleanup, can collide on workspace reuse) and more elaborate than needed. `git push origin --tags` is idempotent — already-pushed refs are skipped without erroring — and replaces the entire dance.

Suggestion: replace the diff/loop with `git push origin --tags`, unless excluding pre-existing local-only tags is intentional (in which case `mktemp` for the temp paths).

.github/workflows/publish.yml
```diff
@@ -59,23 +50,165 @@ jobs:
...
+          pnpm release --tag "$publish_tag"
+          git tag --list | sort > /tmp/tags-after
+
+          comm -13 /tmp/tags-before /tmp/tags-after > /tmp/new-tags
+          if [ -s /tmp/new-tags ]; then
+            while IFS= read -r tag; do
+              git push origin "refs/tags/${tag}"
```
0xbulma: **[HIGH]** GitHub Actions `shell: bash` runs with `set -eo pipefail` by default. If any `git push origin refs/tags/${tag}` in this loop fails — network blip, or a tag that already exists on the remote with a different SHA from a prior partial run — the loop aborts and the remaining tags are never pushed. Result: silent partial state where some packages were published to npm by `pnpm release` but their git tags never reach GitHub, and a retry of the same workflow will see those packages as already-published (so `pnpm release` skips them) and never recreates the missing tags.

Suggestion: collect failures and exit non-zero only after attempting every tag, e.g.
```bash
failed=0
while IFS= read -r tag; do
  git push origin "refs/tags/${tag}" || { echo "::error::failed to push $tag"; failed=1; }
done < /tmp/new-tags
exit "$failed"
```
Or use `git push origin --atomic $(printf 'refs/tags/%s ' $(cat /tmp/new-tags))` for an all-or-nothing push.

README.md
```diff
@@ -44,6 +44,34 @@
...
+
+Commit the generated `.changeset/*.md` file with the source change. Do not update package versions or changelogs manually; the release workflow does that after the change is merged.
+
+If a package changes but no release is needed, add an empty changeset instead:
+
+```sh
+pnpm changeset --empty
+```
+
+After changes land on `main` or `next`, CI runs `pnpm version` and commits the generated version/changelog changes directly to that branch. The next CI run publishes only package versions that have not already been published. Releases from `main` use the `latest` npm tag; releases from `next` use Changesets prerelease mode and publish with the `next` npm tag.
```
0xbulma: **[LOW]** The Changesets section never mentions how to *exit* prerelease mode on `next`. Once `pnpm changeset pre enter next` runs (publish.yml:74-76), `.changeset/pre.json` persists until explicitly exited; merging `next` → `main` without first running `pnpm changeset pre exit` will keep producing prerelease-tagged versions on `main`.

Suggestion: add a sentence noting maintainers must run `pnpm changeset pre exit` and commit the result before merging `next` into `main`. (Bonus: clarify `pnpm version` here is the npm script `pnpm run version` → `changeset version`, not `npm version`, since the two are easy to confuse.)

packages/test/src/vitest.ts
```diff
@@ -19,6 +20,24 @@ export interface ViemTestContext<chain extends Chain = Chain> {
   client: AnvilTestClient<chain>;
 }
 
+const spawnAnvilWithRetry = async (parameters: AnvilArgs) => {
+  const maxAttempts = process.env.CI ? 3 : 1;
```
0xbulma, line 24: **[LOW]** `process.env.CI ? 3 : 1` is truthy on the literal string `"false"` (and on any non-empty value). A developer who has `CI=false` set in their shell silently gets 3 retries with linear backoff locally, masking failures and slowing iteration.

Suggestion: `process.env.CI === "true" ? 3 : 1`, mirroring the convention used elsewhere (the surrounding code reads the same env var on lines 51-52).

packages/test/src/vitest.ts
```diff
@@ -19,6 +20,24 @@ export interface ViemTestContext<chain extends Chain = Chain> {
...
+  let lastError: unknown;
+
+  for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+    try {
+      return await spawnAnvil(parameters);
```
0xbulma, line 29: **[MEDIUM]** When `spawnAnvil` rejects after the child process has already started (e.g., stderr fired before "Listening on"), the underlying anvil subprocess may not be killed before the promise rejects. `spawnAnvilWithRetry` then catches and retries up to 3 times in CI — leaking up to 2 abandoned anvil processes per failing test. Across a long CI run with flaky public RPCs this exhausts ports/memory.

Suggestion: verify `spawnAnvil`'s reject path calls `subprocess.kill('SIGKILL')` before rejecting; if it doesn't, either fix it there, or have this retry wrapper attach a kill in the catch before looping.

packages/test/src/vitest.ts
```diff
@@ -19,6 +20,24 @@ export interface ViemTestContext<chain extends Chain = Chain> {
...
+    } catch (error) {
```
0xbulma, line 30: **[LOW]** The retry catches *all* errors, including programmer errors (TypeError, missing required args). A genuine config bug therefore costs 3 attempts × ~3s of linear backoff per test in CI before surfacing the real error, slowing CI and burying the diagnostic.

Suggestion: narrow the retry to known transient anvil failures, e.g. match on `EADDRINUSE`, `fork`, `timeout`, or a specific Error subclass thrown by `spawnAnvil`.

packages/test/src/vitest.ts
```diff
@@ -29,59 +48,65 @@ export const createViemTest = <chain extends Chain>(
...
-      const client = createAnvilTestClient(
-        http(rpcUrl, {
-          fetchOptions: {
-            cache: "force-cache",
-          },
-          timeout: 30_000,
-        }),
-        chain,
-      );
+      let stop: (() => boolean) | undefined;
```
0xbulma: **[LOW]** `let stop: (() => boolean) | undefined;` plus the rename `const { rpcUrl, stop: stopAnvil } = ...; stop = stopAnvil;` is unnecessary bookkeeping. The `await spawnAnvilWithRetry` on line 61 either resolves with `stop` defined (we enter the try) or throws (we exit the function before the try/finally) — `stop` is never observably undefined.

Suggestion: simplify to
```ts
const { rpcUrl, stop } = await spawnAnvilWithRetry(parameters);
try { ... } finally { stop(); }
```
Drops the `let`, the rename, and the `?.` on line 108.

.github/workflows/publish.yml
Rubilmax: We dont care about maintainability for these since its maintained by agents

.github/workflows/publish.yml
Rubilmax: It's actually a benefit: it syncs up the repository with some configuration-induced diff that can happen on remote installs

packages/test/src/vitest.ts
Rubilmax: Its value is "1" in github actions by default (same on gitlab and other runners)

packages/test/src/vitest.ts
Rubilmax: Expected

packages/test/src/vitest.ts
Rubilmax: Too complex for what it's worth. Let's see if we end up with too many leaking instances

Rubilmax:
@codex

0xbulma (approved):

Foulks-Plb (approved):

PR #564, opened 2026-04-29
chore: Add SDK consolidation TIB

Rubilmax:
## Summary

Adds `TIB-0002`, a proposed technical/business decision record for consolidating the public SDK package surface into `@morpho-org/morpho-sdk` through a staged migration.

## Impact

This is documentation only. It adds `docs/tibs/TIB-0002-consolidate-sdk-packages.md`, outlining the current package split, target consolidation strategy, direct dependency plan for `morpho-sdk`, minimized package export map with root, `/fetch`, `/abis`, and isolated `/augment`, tree-shakeability requirements including `sideEffects` metadata and bundle-impact verification, canonical `blue-sdk-viem` ownership of `FetchParameters` with `chainId` omitted where `morpho-sdk` needs the narrower shape, root re-exports for merged utility/type/entity/action/client surfaces, full `bundler-sdk-viem` ABI extraction into `@morpho-org/morpho-sdk/abis`, broad migrated-code coverage requirements, changeset cascade rules for downstream package patch bumps, review sign-off gates for extracted behavior, deprecation messages for replaced workspace packages and currently published `@morpho-labs/*` packages, no retained migration SDK APIs, and no deprecated-package compatibility window.

## Validation

- `git diff --check`
- Tests not run; staged files are documentation/non-code only.

### DIFF
--- a/docs/tibs/TIB-0002-consolidate-sdk-packages.md
+++ b/docs/tibs/TIB-0002-consolidate-sdk-packages.md
@@ -1,0 +1,527 @@
+# TIB-0002: Consolidate SDK Package Consumption around @morpho-org/morpho-sdk
+
+| Field             | Value       |
+| ----------------- | ----------- |
+| **Status**        | Proposed    |
+| **Date**          | 2026-04-29  |
+| **Author**        | @Rubilmax   |
+| **Scope**         | Repo-wide   |
+| **Supersedes**    | N/A         |
+| **Superseded by** | N/A         |
+
+---
+
+## Context
+
+The SDK monorepo exposes many public packages that are technically distinct but hard for end
+consumers to reason about. Application developers must choose between protocol entity packages,
+viem packages, Wagmi packages, simulation packages, bundler packages, and shared utility packages
+before they can build a basic integration. This increases installation friction, documentation
+surface, dependency management, and support cost.
+
+The business goal is to simplify the end-consumer developer experience by lowering the number of
+packages that can be consumed directly, streamlining installation, and making
+`@morpho-org/morpho-sdk` the canonical package for Morpho application developers.
+
+Several packages still have independent runtime consumers in this monorepo. This TIB therefore
+uses a staged consolidation strategy: preserve current consumer-facing APIs where feasible, make
+`morpho-sdk` own the internal packages it exposes as direct dependencies, keep the rest of the
+package graph on peer dependency semantics, publish replacement APIs and migration notes first where
+a replacement exists, then deprecate only the packages whose public surfaces are replaced or
+explicitly unsupported.
+
+## Goals / Non-Goals
+
+**Goals**
+
+- Make `@morpho-org/morpho-sdk` the primary public SDK entrypoint for application developers.
+- Reduce the number of packages consumers need to install, import, and understand.
+- Preserve the pre-existing SDK developer experience where feasible, including import style, public
+  interfaces, runtime APIs, and consumption patterns.
+- Keep `@morpho-org/morpho-ts`, `@morpho-org/blue-sdk`, and `@morpho-org/blue-sdk-viem` as
+  maintained packages, while making them direct dependencies of `morpho-sdk` only.
+- Re-export every fetcher from `blue-sdk-viem` and the classes from `blue-sdk` through
+  `morpho-sdk`.
+- Move selected bundler helpers and narrow reallocation helpers behind `morpho-sdk`.
+- Deprecate legacy packages only after replacement APIs exist or the no-replacement decision is
+  documented.
+- Deprecate migration SDK packages, starting with `@morpho-org/migration-sdk-viem`, with no retained
+  migration APIs.
+- Deprecate all currently published `@morpho-labs/*` packages on npm with clear replacement
+  guidance.
+- Simplify reallocation computation by removing the broad simulation engine dependency from
+  `morpho-sdk`.
+
+**Non-Goals**
+
+- Replacing deprecated Wagmi packages with new first-party React hooks in this TIB.
+- Deprecating `@morpho-org/morpho-ts`, `@morpho-org/blue-sdk`, or
+  `@morpho-org/blue-sdk-viem`.
+- Converting every internal Morpho package relationship from peer dependencies to direct
+  dependencies.
+- Removing packages before maintained monorepo consumers have migrated.
+- Rewriting unrelated SDKs such as liquidation, liquidity, or evm simulation beyond import-path
+  migration needed by this consolidation.
+- Editing generated build output under `lib/`.
+
+## Current Solution
+
+The current public SDK surface is split across several packages:
+
+- `@morpho-org/morpho-sdk` exposes higher-level client, entity, action, helper, and type APIs.
+- `@morpho-org/blue-sdk` exposes core protocol entities, math, constants, errors, addresses, and
+  utility types.
+- `@morpho-org/blue-sdk-viem` exposes viem ABIs, fetchers, signatures, typed-data helpers, viem
+  utilities, and optional module augmentation for `blue-sdk` entities.
+- `@morpho-org/bundler-sdk-viem` exposes action and bundle encoding utilities, with runtime
+  dependencies on `simulation-sdk` and `blue-sdk-viem`.
+- `@morpho-org/migration-sdk-viem` exposes migration position models, protocol migration
+  addresses, third-party protocol fetchers, and migration transaction requirement types.
+- `@morpho-org/simulation-sdk` exposes a broad mutable simulation state, operation types,
+  dispatchers, handlers, simulation constants, and public allocator reallocation helpers.
+- `@morpho-org/blue-sdk-wagmi`, `@morpho-org/simulation-sdk-wagmi`, and
+  `@morpho-org/test-wagmi` expose React/Wagmi hooks and testing helpers.
+- `@morpho-org/morpho-ts` exposes shared TypeScript utilities such as `Time`, `format`,
+  `deepFreeze`, collection helpers, retry helpers, and shared URLs.
+- The legacy `@morpho-labs/*` npm scope still contains published packages that should no longer be
+  presented as current SDK entrypoints.
+
+The dependency graph shows that several target packages are still consumed independently:
+
+| Package | Used independently in monorepo? | Consequence |
+| --- | --- | --- |
+| `@morpho-org/morpho-ts` | Yes. Runtime dependency of `blue-sdk`, `evm-simulation`, `liquidity-sdk-viem`, `liquidation-sdk-viem`, `migration-sdk-viem`, `morpho-test`, `morpho-sdk`, and deprecated packages. | Keep as a maintained package. Only `morpho-sdk` moves it to a direct dependency for this consolidation; other packages keep their existing peer dependency model unless a separate decision changes it. |
+| `@morpho-org/blue-sdk` | Yes. Heavily used by maintained packages. | Keep as a maintained package. `morpho-sdk` depends on it directly and re-exports its classes; `blue-sdk-viem` and other packages keep `blue-sdk` as a peer dependency. |
+| `@morpho-org/blue-sdk-viem` | Yes. Runtime use in `liquidity-sdk-viem`, `liquidation-sdk-viem`, `migration-sdk-viem`, and `morpho-sdk`. | Keep as a maintained viem integration package. `morpho-sdk` depends on it directly and re-exports every fetcher; other packages keep their existing peer dependency model. |
+| `@morpho-org/bundler-sdk-viem` | Yes. Runtime use in `migration-sdk-viem` and `morpho-sdk`. | Move selected action/bundle APIs into `morpho-sdk`; migrate or deprecate `migration-sdk-viem` before removal. |
+| `@morpho-org/migration-sdk-viem` | No maintained monorepo runtime consumers outside its own package, tests, and documentation. | Deprecate as a public package. Retain no migration APIs under `morpho-sdk`; document that migration SDK workflows are no longer a supported public SDK surface. |
+| `@morpho-org/simulation-sdk` | Yes. Runtime use in `bundler-sdk-viem`, `liquidity-sdk-viem`, `migration-sdk-viem`, and `morpho-sdk`. | Deprecate broad simulation engine, but first move required constants/types/helpers into owning packages. |
+| `@morpho-org/blue-sdk-wagmi` | Only consumed by deprecated `simulation-sdk-wagmi`. | Deprecate directly; no first-party replacement hooks in this TIB. |
+| `@morpho-org/simulation-sdk-wagmi` | Runtime only for itself; otherwise dev/test use. | Deprecate directly after internal consumers migrate. |
+| `@morpho-org/test-wagmi` | Test-only helper for Wagmi package tests and some package test suites. | Deprecate public package; move any needed test helper internally or into `@morpho-org/test`. |
+
+## Proposed Solution
+
+Consolidate the public application-developer SDK experience around `@morpho-org/morpho-sdk`, while
+keeping `morpho-ts`, `blue-sdk`, and `blue-sdk-viem` as maintained packages. `morpho-sdk` is the
+single Morpho package entrypoint for application consumers, so it owns those internal packages as
+direct dependencies. Other packages keep their peer dependency model.
+
+`morpho-sdk` becomes the canonical application package by:
+
+- depending directly on `blue-sdk` for protocol entities, math, constants, errors, addresses, and
+  protocol value types when `morpho-sdk` imports them directly;
+- re-exporting the public classes from `blue-sdk` through the `morpho-sdk` root export;
+- consuming viem ABIs, fetchers, signatures, typed-data helpers, viem utilities, and deployless
+  fetch types through a direct dependency on `blue-sdk-viem`;
+- re-exporting every fetcher from `blue-sdk-viem` through `@morpho-org/morpho-sdk/fetch`;
+- depending directly on `morpho-ts` where shared TypeScript utilities are part of its runtime or
+  public API implementation;
+- exposing selected bundler action and bundle encoding APIs currently required by `morpho-sdk` and
+  migration flows;
+- owning narrow public allocator reallocation types and helpers currently depending on
+  `simulation-sdk`.
+
+### Canonical `morpho-sdk` Export Hierarchy
+
+Keep all consolidated consumer imports under the single `@morpho-org/morpho-sdk` package name.
+Do not create package-name variants such as `@morpho-org/morpho-sdk-*` for these surfaces. Minimize
+package subpaths: thin value, type, utility, action, client, and entity domains are re-exported from
+the package root, while only fetch utilities, ABI literals, and side-effecting augmentation keep
+dedicated public subpaths.
+
+- `@morpho-org/morpho-sdk`: the primary application import surface. Re-export current
+  `morpho-sdk` client, action, entity, helper, and type APIs from the root, and add root re-exports
+  for `blue-sdk` entity barrels, addresses, constants, errors, math helpers, utility types,
+  capacity helpers, viem parsing/restructuring helpers, and `blue-sdk-viem` signature helpers.
+- `@morpho-org/morpho-sdk/fetch`: re-export `packages/blue-sdk-viem/src/fetch/index.ts`. This
+  remains a dedicated subpath because fetchers are the network/read layer and are expected to be
+  imported together.
+- `@morpho-org/morpho-sdk/abis`: expose `blue-sdk-viem` ABI literals and the complete
+  `bundler-sdk-viem` ABI barrel. This remains a dedicated subpath because ABI arrays are large,
+  specialized constants that many root consumers do not need.
+- `@morpho-org/morpho-sdk/augment`: expose the optional side-effecting module augmentation
+  entrypoint. This is the only side-effecting public subpath.
+
+Example imports after consolidation:
+
+```ts
+import { ChainId, Market, Token, Vault, safeParseUnits } from "@morpho-org/morpho-sdk";
+import { fetchMarket, fetchToken, fetchVault } from "@morpho-org/morpho-sdk/fetch";
+import { bundler3Abi } from "@morpho-org/morpho-sdk/abis";
+import "@morpho-org/morpho-sdk/augment";
+```
+
+Implement the package export map exactly with these public subpaths:
+
+```json
+{
+  "sideEffects": ["./lib/*/augment/**/*.js"],
+  "publishConfig": {
+    "exports": {
+      ".": {
+        "types": "./lib/esm/index.d.ts",
+        "import": "./lib/esm/index.js",
+        "require": "./lib/cjs/index.js"
+      },
+      "./abis": {
+        "types": "./lib/esm/abis/index.d.ts",
+        "import": "./lib/esm/abis/index.js",
+        "require": "./lib/cjs/abis/index.js"
+      },
+      "./augment": {
+        "types": "./lib/esm/augment/index.d.ts",
+        "import": "./lib/esm/augment/index.js",
+        "require": "./lib/cjs/augment/index.js"
+      },
+      "./fetch": {
+        "types": "./lib/esm/fetch/index.d.ts",
+        "import": "./lib/esm/fetch/index.js",
+        "require": "./lib/cjs/fetch/index.js"
+      }
+    }
+  }
+}
+```
+
+Implement only `src/fetch/index.ts`, `src/abis/index.ts`, and `src/augment/index.ts` as package
+subpath barrels. All other new barrels are internal source organization and are re-exported from
+`packages/morpho-sdk/src/index.ts`.
+Do not add public package subpaths such as `./actions`, `./client`, `./entities`, `./helpers`,
+`./utils`, `./types`, `./math`, `./constants`, `./addresses`, `./errors`, or `./signatures`.
+
+- `src/index.ts` re-exports the current `morpho-sdk` action, client, entity, helper, and type
+  barrels, plus the new internal barrels for root-level `blue-sdk` and `blue-sdk-viem` surfaces.
+- `src/entities/index.ts` is an internal root barrel that re-exports public `blue-sdk` market,
+  token, holding, user, position, and vault entity entrypoints rather than a hand-maintained class
+  list, so abstract/base entities such as vault adapter classes are not silently omitted.
+- `src/utils/index.ts` is an internal root barrel. It re-exports current `morpho-sdk` helper
+  functions from `src/helpers/index.ts`, `CapacityLimitReason` and `CapacityLimit` from `blue-sdk`,
+  and `safeParseNumber`, `safeParseUnits`, `safeGetAddress`, `restructure`, and
+  `readContractRestructured` from `blue-sdk-viem`.
+- `src/types/index.ts` remains the merged root type barrel. It keeps the existing `morpho-sdk` type
+  exports, except that the public `FetchParameters` name is owned by `blue-sdk-viem` and includes
+  optional `chainId`. It also adds `Address`, `MarketId`, `BigIntish`, `Loadable`, `Failable`,
+  `Fetchable`, `TransactionType`, `ChainId`, `ChainMetadata`, and `DeploylessFetchParameters`.
+  `morpho-sdk` call sites that must not accept `chainId` should consume `FetchParameters` as
+  `Omit<FetchParameters, "chainId">` or a clearly named local alias instead of exporting a second
+  conflicting `FetchParameters` name or silently widening those signatures.
+- `src/addresses/index.ts`, `src/constants/index.ts`, `src/errors/index.ts`, `src/math/index.ts`,
+  and `src/signatures/index.ts` are internal root barrels, not package subpaths.
+
+### Tree-Shakeability and Side Effects
+
+The consolidation must not make the `morpho-sdk` root import pull in unused fetchers, ABI literals,
+or side-effecting augmentation code.
+
+- Set `package.json#sideEffects` to the narrow augmentation whitelist shown in the export map
+  snippet. This marks every non-augmentation artifact as side-effect free while preserving the
+  explicit `@morpho-org/morpho-sdk/augment` side effect.
+- Do not set package-wide `sideEffects` to `true`; the whitelist is the side-effect-free posture
+  with only augmentation opted back in.
+- If augmentation is removed in a later TIB, replace the whitelist with `"sideEffects": false`.
+- `src/index.ts`, `src/fetch/index.ts`, and `src/abis/index.ts` must never top-level import
+  `src/augment/index.ts` or any module that mutates `blue-sdk` prototypes.
+- Root imports from `blue-sdk-viem` must target pure source modules directly, such as signatures,
+  types, or utils. The `morpho-sdk` root must not import the `blue-sdk-viem` root barrel because
+  that barrel also re-exports fetchers and ABIs.
+- `src/augment/index.ts` is the only entrypoint allowed to top-level import side-effecting
+  augmentation modules.
+- Public entrypoint barrels must use explicit named re-exports rather than `export *` chains.
+  Internal entity/type export lists may be generated or covered by reviewer checklist items so new
+  upstream public entities are not silently omitted, but the published entrypoint files should stay
+  explicit for older bundlers' dead-code elimination.
+- Phase 2 must include a bundle-impact verification step that compares the consolidated branch
+  against `origin/main` using the same bundler configuration. The check must cover at least a root
+  import, a `/fetch` import, a `/abis` import, and an `/augment` import, and report whether root
+  imports accidentally retain fetch, ABI, or augmentation code.
+
+### Developer Experience Preservation
+
+Preserve current SDK consumption patterns where feasible. This includes module augmentation when
+consumers already rely on it, but module augmentation is an implementation detail of the broader
+developer-experience goal.
+
+- Preserve existing public interfaces and runtime behavior unless a package boundary change makes a
+  migration unavoidable.
+- Keep existing `blue-sdk-viem` ergonomics available through `blue-sdk-viem` exports and
+  `morpho-sdk` root or subpath exports during the transition.
+- Avoid forcing application consumers to install or import `morpho-ts`, `blue-sdk`, or
+  `blue-sdk-viem` when they only use `morpho-sdk`.
+- Document any import-path changes with one-to-one migration guidance. Deprecated packages have no
+  wrapper compatibility window.
+
+The implementation may still expose side-effect augmentation entrypoints and explicit helper
+functions where those preserve existing usage, but the success criterion is stable SDK consumption,
+not preserving a specific augmentation mechanism for its own sake.
+
+### Reallocation Simplification
+
+Keep the current friendly-then-aggressive public allocator algorithm, but remove the broad
+simulation engine from `morpho-sdk`:
+
+- replace `SimulationState` inputs with a narrow local reallocation data shape containing only the
+  market, vault, vault market config, position, holding, and block data needed by the algorithm;
+- extract small helper functions for candidate withdrawal computation, reallocation state updates,
+  target-market capacity checks, and vault grouping;
+- preserve current output invariants: positive withdrawal amounts, no withdrawals from the target
+  market, strictly sorted withdrawals per vault, vault fee inclusion, and deterministic ordering;
+- move approval and reallocation constants currently imported from `simulation-sdk` into the owning
+  `morpho-sdk` API surface before deprecating `simulation-sdk`.
+
+### Extraction Targets Before Deprecation
+
+Before `bundler-sdk-viem` or `simulation-sdk` can be deprecated, `morpho-sdk` must own the narrow
+runtime surfaces it currently imports or exposes through its higher-level actions, plus the full
+bundler ABI surface that consumers may need after `bundler-sdk-viem` is deprecated.
+
+From `packages/bundler-sdk-viem` extract into `morpho-sdk`:
+
+- `src/BundlerAction.ts`: `BundlerAction.encodeBundle`, `BundlerAction.encode`, `BundlerCall`, and
+  the action encoders needed by current `morpho-sdk` action builders: `nativeTransfer`,
+  `wrapNative`, `erc20TransferFrom`, `erc20Transfer`, `permit`, `approve2`, `transferFrom2`,
+  `erc4626Deposit`, `erc4626Redeem`, `morphoSupplyCollateral`, `morphoBorrow`, `morphoRepay`,
+  `morphoWithdrawCollateral`, and `publicAllocatorReallocateTo`.
+- `src/types/actions.ts`: the `Action` discriminated union and supporting types used to build those
+  bundles, including `ActionArgs`, `ActionType`, `Actions`, `Authorization`, `InputReallocation`,
+  `Permit2PermitSingleDetails`, and `Permit2PermitSingle`.
+- `src/abis.ts`: the complete ABI barrel from `bundler-sdk-viem`, not only the ABI literals used by
+  current `morpho-sdk` actions. Extract and re-export `universalRewardsDistributorAbi`,
+  `bundler3Abi`, `coreAdapterAbi`, `generalAdapter1Abi`, `ethereumGeneralAdapter1Abi`,
+  `paraswapAdapterAbi`, `erc20WrapperAdapterAbi`, `aaveV2MigrationAdapterAbi`,
+  `aaveV3MigrationAdapterAbi`, `aaveV3OptimizerMigrationAdapterAbi`,
+  `compoundV2MigrationAdapterAbi`, and `compoundV3MigrationAdapterAbi` from
+  `@morpho-org/morpho-sdk/abis`.
+- `src/errors.ts`: the `BundlerErrors` namespace members thrown by the extracted encoders:
+  `MissingSignature`, `UnexpectedAction`, and `UnexpectedSignature`.
+- `src/operations.ts`: `DEFAULT_SUPPLY_TARGET_UTILIZATION`, currently used by
+  `computeReallocations`.
+
+Do not extract `ActionBundle`, `ActionBundleRequirements`, operation-level simulation helpers
+(`populateBundle`, `populateSubBundle`, `finalizeBundle`, `simulateBundlerOperation`,
+`simulateBundlerOperations`, or `handleBundlerOperation`), migration adapter actions, Paraswap
+actions, or reward-claim actions unless `morpho-sdk` intentionally exposes those workflows later.
+The ABI literals above are the only parts of those workflows that move unconditionally. No
+`bundler-sdk-viem` classes are required by the current `morpho-sdk` action API.
+
+From `packages/simulation-sdk` extract or replace inside `morpho-sdk`:
+
+- `src/constants.ts`: `APPROVE_ONLY_ONCE_TOKENS` and `MAX_TOKEN_APPROVALS`, used by approval
+  requirement encoding.
+- `src/SimulationState.ts`: replace the broad `SimulationState` dependency with a local
+  reallocation data shape and helpers covering only `MinimalBlock`, `PublicAllocatorOptions`,
+  `PublicReallocation`, market lookup, vault lookup, vault market config lookup, accrual position
+  lookup, and the `getMarketPublicReallocations` algorithm used by `computeReallocations`.
+- `src/SimulationState.ts`: keep the `DEFAULT_WITHDRAWAL_TARGET_UTILIZATION` default with the
+  local public allocator options.
+- `src/errors.ts`: recreate only the local typed failures needed by the extracted reallocation
+  helpers, such as unknown market, vault, vault-market-config, and position cases. Do not copy the
+  full simulation error namespace.
+- `src/operations.ts`: no broad operation unions are required for `morpho-sdk` reallocations after
+  the local reallocation data shape exists; keep `BlueOperation`, `MetaMorphoOperation`,
+  `Erc20Operation`, `ParaswapOperation`, and dispatcher types out of `morpho-sdk`.
+- `src/helpers/mutative.ts` and `src/helpers/paraswap.ts`: no utilities are required for the
+  current extraction. Do not extract `getCurrent`, `augustusV6_2Address`,
+  `paraswapContractMethodOffsets`, or `getParaswapContractMethodOffsets` unless Paraswap workflows
+  move into `morpho-sdk`.
+
+Before removing either dependency, reviewers must explicitly sign off that extracted bundle
+calldata encoding, approval constants, and friendly/aggressive public allocator reallocation
+behavior match the intended existing behavior.
+
+### Workspace Deprecation Scope
+
+Deprecate replaced `@morpho-org/*` workspace packages only after the implementation phases provide
+their replacement paths or document that no replacement exists. There is no compatibility window and
+no wrapper release for deprecated packages.
+
+Use these npm deprecation messages:
+
+| Package | Replacement | npm deprecation message |
+| --- | --- | --- |
+| `@morpho-org/bundler-sdk-viem` | `@morpho-org/morpho-sdk` | `Deprecated: use @morpho-org/morpho-sdk. Bundler action helpers are consolidated under @morpho-org/morpho-sdk and ABIs are under @morpho-org/morpho-sdk/abis.` |
+| `@morpho-org/migration-sdk-viem` | None | `Deprecated: no replacement package. Migration SDK workflows are no longer a supported public SDK surface.` |
+| `@morpho-org/simulation-sdk` | `@morpho-org/morpho-sdk` | `Deprecated: use @morpho-org/morpho-sdk for supported reallocation helpers. The broad simulation engine has no replacement package.` |
+| `@morpho-org/blue-sdk-wagmi` | None | `Deprecated: no replacement package. First-party Wagmi hooks are no longer a supported public SDK surface.` |
+| `@morpho-org/simulation-sdk-wagmi` | None | `Deprecated: no replacement package. First-party Wagmi hooks are no longer a supported public SDK surface.` |
+| `@morpho-org/test-wagmi` | None | `Deprecated: no replacement package. This public Wagmi test helper package is no longer supported.` |
+
+If additional migration SDK packages are introduced or discovered before implementation, include
+them in the same migration SDK deprecation campaign with replacement `None` unless a later TIB
+explicitly adds the workflow to `morpho-sdk`.
+
+### npm Deprecation Scope
+
+Deprecate every currently published `@morpho-labs/*` package on npm. Do not unpublish these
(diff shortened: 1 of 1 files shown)


Foulks-Plb (commented):
why not a folder for TIB, but the convention is to keep them under docs/tibs.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,445 @@
...
+- **Phase 3 -- Preserve peer dependencies elsewhere:** Keep `blue-sdk`, `blue-sdk-viem`, and
+  `morpho-ts` as peer dependencies for packages other than `morpho-sdk` unless a separate package
+  decision changes that relationship.
+- **Phase 4 -- Extract bundler and simulation prerequisites:** Move the explicit extraction targets
+  above into `morpho-sdk`, keep the extracted public API behind `morpho-sdk` exports, and add parity
+  tests before any deprecation notice is published.
+- **Phase 5 -- Simplify simulation and reallocations:** Replace `morpho-sdk`'s `SimulationState`
+  dependency with local reallocation helpers. Move approval constants, public allocator option
+  types, and other narrow constants out of `simulation-sdk` into their owning APIs.
+- **Phase 6 -- Migration SDK deprecation prep:** Decide which migration APIs remain public. Move
```
Foulks-Plb: Remove all migrations?

Why not archive them in a separate package too?

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+  the public `morpho-sdk` path for viem fetch utilities, including `fetchUser`, `fetchToken`,
+  `fetchHolding`, `fetchMarketParams`, `fetchMarket`, `fetchPosition`,
+  `fetchAccrualPosition`, `fetchPreLiquidationParams`, `fetchPreLiquidationPosition`,
+  `fetchVaultConfig`, `fetchVault`, `fetchAccrualVault`, `fetchVaultUser`,
+  `fetchVaultMarketConfig`, `fetchVaultMarketAllocation`,
+  `fetchVaultMarketPublicAllocatorConfig`, `fetchVaultV2`, `fetchAccrualVaultV2`,
+  `fetchVaultV2Adapter`, `fetchAccrualVaultV2Adapter`,
+  `fetchVaultV2MorphoMarketV1Adapter`, `fetchAccrualVaultV2MorphoMarketV1Adapter`,
+  `fetchVaultV2MorphoMarketV1AdapterV2`, `fetchAccrualVaultV2MorphoMarketV1AdapterV2`,
+  `fetchVaultV2MorphoVaultV1Adapter`, and `fetchAccrualVaultV2MorphoVaultV1Adapter`. Helper
```
0xbulma: **[HIGH] Hard-coded re-export lists are staleness traps.** L134–147 enumerate 26 fetcher names and L148–156 enumerate 28 entity class names. Any addition upstream becomes a silent omission discoverable only by a downstream consumer — no test or build failure. Mandate either (a) explicit `export *` barrels from the source package's `src/index.ts`, or (b) a CI check that diffs the source package's exported symbol set against the `morpho-sdk` re-export barrels and fails on additions not propagated. Pick one in this TIB.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+  surface.
+- `@morpho-org/morpho-sdk/entities`: re-export public `blue-sdk` entity classes that consumers
+  instantiate or receive from fetchers, including `Token`, `WrappedToken`, `ConstantWrappedToken`,
+  `ExchangeRateWrappedToken`, `VaultToken`, `Eip5267Domain`, `MarketParams`, `Market`, `Holding`,
+  `AssetBalances`, `User`, `Position`, `AccrualPosition`, `PreLiquidationParams`,
+  `PreLiquidationPosition`, `VaultConfig`, `VaultMarketAllocation`, `VaultMarketConfig`,
+  `VaultMarketPublicAllocatorConfig`, `VaultUser`, `Vault`, `AccrualVault`, `VaultV2`,
+  `AccrualVaultV2`, `VaultV2MorphoMarketV1Adapter`, `AccrualVaultV2MorphoMarketV1Adapter`,
+  `VaultV2MorphoMarketV1AdapterV2`, `AccrualVaultV2MorphoMarketV1AdapterV2`,
+  `VaultV2MorphoVaultV1Adapter`, and `AccrualVaultV2MorphoVaultV1Adapter`.
```
0xbulma: **[MEDIUM] Silent omission in `/entities` enumeration.** `VaultV2Adapter` — the abstract base class at `packages/blue-sdk/src/vault/v2/VaultV2Adapter.ts:13` — is missing from the list, yet its fetchers `fetchVaultV2Adapter` and `fetchAccrualVaultV2Adapter` are listed at L141. Consumers can fetch instances but cannot import the abstract base for type narrowing. Add it.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- `@morpho-org/morpho-sdk/utils`: re-export generic utilities from `blue-sdk` and
+  `blue-sdk-viem`, including capacity types and viem parsing/restructuring helpers. Where names
+  collide with existing `morpho-sdk` helpers, keep existing `morpho-sdk` names stable and add an
+  explicit alias rather than silently shadowing an export.
```
0xbulma: **[HIGH] Subpath collision audit only covers `/utils`.** Bundler extraction at L320–322 pulls `Action`, `Actions`, `ActionType`, `Authorization`, `Permit2PermitSingle`, `Permit2PermitSingleDetails`, `InputReallocation` while `packages/morpho-sdk/src/types/action.ts` already exports `BaseAction`, `TransactionAction`, `Transaction`, `MorphoAuthorizationAction`, `Permit2Args`, `Permit2Action`, `Requirement`, `RequirementSignature` (verified). Re-exporting both families through `/actions` and `/types` creates TS name clashes on `Action`/`Actions`/`ActionType`/`Authorization`. The collision rule established here for `/utils` must extend explicitly to every shared identifier across `/actions`, `/types`, `/errors`, `/constants`, `/abis`. Add an aliasing convention (e.g., `BundlerAction`, `BundlerActions`, `BundlerActionType`, `BundlerAuthorization`).

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- `@morpho-org/morpho-sdk/math`, `@morpho-org/morpho-sdk/constants`,
+  `@morpho-org/morpho-sdk/addresses`, `@morpho-org/morpho-sdk/errors`, and
```
0xbulma: **[MEDIUM] `/errors` collision risk between upstream and existing `morpho-sdk` errors.** `blue-sdk/src/errors.ts` exports `UnknownTokenError`, `UnknownMarketParamsError`, `UnknownVaultConfigError`, `UnsupportedVaultV2AdapterError`, `UnknownDataError`, plus `BlueErrors`/`VaultV2Errors` namespaces. `simulation-sdk/src/errors.ts` exports `UnknownMarketError`, `UnknownUserError`, `UnknownPositionError`, `UnknownVaultV2AdapterError`, `UnknownVaultV2Error`, etc. The TIB at L352 recreates the simulation-sdk failures locally inside `morpho-sdk` without stating whether those recreated names collide with `blue-sdk`'s classes (also re-exported through `/errors`) or with `morpho-sdk`'s own existing error classes in `src/types/error.ts`. List the recreated class names and a dedup policy before approval.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+  `@morpho-org/morpho-sdk/types`: expose the corresponding `blue-sdk` surfaces alongside existing
+  `morpho-sdk` types under the same subpath naming convention.
+- `@morpho-org/morpho-sdk/signatures`: expose the corresponding `blue-sdk-viem` typed-data
+  helpers.
```
0xbulma: **[MEDIUM] `/signatures` plan is shallow.** `blue-sdk-viem/src/signatures` barrels `manager.ts`, `permit.ts`, `permit2.ts` (e.g. `getPermitTypedData`, `getAuthorizationTypedData`) and `packages/morpho-sdk/src/actions/requirements/` already consumes typed-data helpers in `getRequirementsPermit.ts` / `getRequirementsPermit2.ts` / `encode/encodeErc20Permit.ts`. Enumerate which signature helpers re-export through `/signatures` and the collision policy with `morpho-sdk`'s existing requirements-flow exports (`Permit2Action`, `PermitAction`, `RequirementSignature`).

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+  `VaultV2MorphoMarketV1AdapterV2`, `AccrualVaultV2MorphoMarketV1AdapterV2`,
+  `VaultV2MorphoVaultV1Adapter`, and `AccrualVaultV2MorphoVaultV1Adapter`.
+- `@morpho-org/morpho-sdk/utils`: re-export generic utilities from `blue-sdk` and
+  `blue-sdk-viem`, including capacity types and viem parsing/restructuring helpers. Where names
+  collide with existing `morpho-sdk` helpers, keep existing `morpho-sdk` names stable and add an
+  explicit alias rather than silently shadowing an export.
+- `@morpho-org/morpho-sdk/math`, `@morpho-org/morpho-sdk/constants`,
+  `@morpho-org/morpho-sdk/addresses`, `@morpho-org/morpho-sdk/errors`, and
+  `@morpho-org/morpho-sdk/types`: expose the corresponding `blue-sdk` surfaces alongside existing
+  `morpho-sdk` types under the same subpath naming convention.
```
0xbulma: **[MEDIUM] Subpath proliferation unjustified.** 13 subpaths defined here (`/fetch`, `/entities`, `/utils`, `/math`, `/constants`, `/addresses`, `/errors`, `/types`, `/signatures`, `/abis`, `/actions`, `/client`, plus the root). `/math`, `/constants`, `/addresses`, `/errors`, `/types` are five subpaths for what is essentially primitive value/type surface and could reasonably collapse into `/types` + `/utils`, or even into the root. Each subpath multiplies `package.json` exports entries, dual ESM/CJS `lib/` artifacts, TS path-mapping, and review surface. Add a justification sentence per subpath or collapse the thin ones.

0xbulma (commented):
Pre-merge review of the consolidation TIB. Most findings flag under-specified implementation details that will become regression risks at code-landing time. The recent rewrite resolved the migration-SDK contradictions, package.json export map, and compatibility-window hedging — those drops are reflected here. Severity legend: **[HIGH]** must-fix before approval; **[MEDIUM]** should-fix; **[LOW]** stylistic.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- `@morpho-org/morpho-sdk/actions` and `@morpho-org/morpho-sdk/client`: keep the current
+  `morpho-sdk` action and client surfaces.
+
+Example imports after consolidation:
+
+```ts
+import { Market, Token, Vault } from "@morpho-org/morpho-sdk/entities";
+import { fetchMarket, fetchToken, fetchVault } from "@morpho-org/morpho-sdk/fetch";
+import { safeParseUnits } from "@morpho-org/morpho-sdk/utils";
+```
```
0xbulma, line 153: **[HIGH] Tree-shakeability is not addressed.** `morpho-sdk/package.json` does not currently set `sideEffects: false` (verified). With augmentation entrypoints preserved (L289–291) and 13 new subpaths re-exporting from `blue-sdk-viem`, the consolidation must specify: (1) `morpho-sdk` adds `sideEffects: false` to package.json and the `sideEffects` array narrowly whitelists only the augmentation subpath (e.g., `morpho-sdk/augment`); (2) the main entrypoint and pure subpaths never top-level-import any side-effecting augmentation module; (3) subpath barrels use explicit named re-exports rather than `export *` chains that can defeat dead-code elimination in older bundlers; (4) a bundle-impact verification step in Phase 2 with a baseline diff against `main`.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+        "require": "./lib/cjs/math/index.js"
+      },
+      "./signatures": {
+        "types": "./lib/esm/signatures/index.d.ts",
+        "import": "./lib/esm/signatures/index.js",
+        "require": "./lib/cjs/signatures/index.js"
+      },
+      "./types": {
+        "types": "./lib/esm/types/index.d.ts",
+        "import": "./lib/esm/types/index.js",
```
0xbulma: **[HIGH] `MAX_ABSOLUTE_SHARE_PRICE` is exported from two packages with identical values today.** `packages/bundler-sdk-viem/src/actions.ts:50` and `packages/morpho-sdk/src/helpers/constant.ts:10` both export `MAX_ABSOLUTE_SHARE_PRICE = 100n * MathLib.RAY` (verified). Root CLAUDE.md says: *"Prefer deleting unclear helpers, constants, exports, or duplicated logic before adding abstractions."* Once bundler symbols flow through `morpho-sdk/constants` or `morpho-sdk/actions` after extraction, this duplicate becomes an ambient TS collision. Pick a canonical owner (`morpho-sdk` is the natural home given `packages/morpho-sdk/CLAUDE.md` already references this constant), delete the other, and cite the resolution in this TIB before the consolidation lands.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+  and `safeParseNumber`, `safeParseUnits`, `safeGetAddress`, `restructure`, and
+  `readContractRestructured` from `blue-sdk-viem`.
+- `src/types/index.ts` remains the merged type barrel. It keeps the existing `morpho-sdk` type
+  exports and adds `Address`, `MarketId`, `BigIntish`, `Loadable`, `Failable`, `Fetchable`,
+  `TransactionType`, `ChainId`, `ChainMetadata`, `FetchParameters`, and
+  `DeploylessFetchParameters`.
+
+### Developer Experience Preservation
+
+Preserve current SDK consumption patterns where feasible. This includes module augmentation when
```
0xbulma, line 241: **[MEDIUM] "where feasible" hedge anchors the whole Developer Experience Preservation section.** The phrase repeats at L276, L280 ("unless a package boundary change makes a migration unavoidable"), L284 ("when they only use `morpho-sdk`"), and L289–291. Without a concrete contract, integrators cannot plan migrations. Replace with explicit enumeration: which existing public exports are preserved verbatim, which require an import-path migration, which break.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- Preserve existing public interfaces and runtime behavior unless a package boundary change makes a
+  migration unavoidable.
+- Keep existing `blue-sdk-viem` ergonomics available through `blue-sdk-viem` exports and
+  `morpho-sdk` subpath exports during the transition.
+- Avoid forcing application consumers to install or import `morpho-ts`, `blue-sdk`, or
+  `blue-sdk-viem` when they only use `morpho-sdk`.
+- Document any import-path changes with one-to-one migration guidance. Deprecated packages have no
+  wrapper compatibility window.
+
+The implementation may still expose side-effect augmentation entrypoints and explicit helper
```
0xbulma, line 254: **[HIGH] Module augmentation handoff requires a binding test.** `blue-sdk-viem` ships 12 augmented entity classes under `packages/blue-sdk-viem/src/augment/` (`Token`, `Market`, `Vault`, `Holding`, `Position`, `User`, etc.) that attach static fetch helpers via module-augmentation side effects. After consolidation users will obtain those classes through `@morpho-org/morpho-sdk/entities`. Failure mode is silent: if the augmentation side-effect entry is not transitively loaded, augmented methods (`Market.fetch`, etc.) are `undefined` at runtime with no static type error. Mandate (a) per-augmented-class runtime test that imports the entity only via the new `morpho-sdk` subpath, asserts the augmented method exists, and executes it against an Anvil fork; (b) a packaged downstream-style fixture that imports only `@morpho-org/morpho-sdk` (no direct `blue-sdk-viem` import) and exercises each augmented method end-to-end. Without both, tree-shaking or a `sideEffects` misconfiguration silently downgrades to an unaugmented class.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+functions where those preserve existing usage, but the success criterion is stable SDK consumption,
+not preserving a specific augmentation mechanism for its own sake.
+
+### Reallocation Simplification
+
+Keep the current friendly-then-aggressive public allocator algorithm, but remove the broad
```
0xbulma, line 260: **[MEDIUM] "friendly-then-aggressive public allocator algorithm" appears once and is undefined.** Define inline (one sentence on the two-phase strategy) or link to the source file/function that implements it. External maintainers and future agents will not know what this means.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+simulation engine from `morpho-sdk`:
+
+- replace `SimulationState` inputs with a narrow local reallocation data shape containing only the
+  market, vault, vault market config, position, holding, and block data needed by the algorithm;
+- extract small helper functions for candidate withdrawal computation, reallocation state updates,
+  target-market capacity checks, and vault grouping;
+- preserve current output invariants: positive withdrawal amounts, no withdrawals from the target
+  market, strictly sorted withdrawals per vault, vault fee inclusion, and deterministic ordering;
```
0xbulma, line 268: **[HIGH] Output invariants are prose, not test deliverables.** L302–303 list five invariants (positive withdrawal amounts, no withdrawals from target market, strictly sorted per vault, fee inclusion, deterministic ordering). Existing tests at `packages/morpho-sdk/src/helpers/computeReallocations.test.ts` only partially cover sort + amount; grep finds no assertion of target-market exclusion or deterministic vault ordering across runs. Rewrite this bullet as a hard deliverable: *"Each invariant must have a dedicated unit test that fails if the invariant guard is removed from the algorithm."* Root CLAUDE.md says: *"Codify security-sensitive behavior as tests, especially routing, authorization, chain, and accounting invariants."*

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- move approval and reallocation constants currently imported from `simulation-sdk` into the owning
+  `morpho-sdk` API surface before deprecating `simulation-sdk`.
```
0xbulma, line 270: **[LOW] Colocation convention not called out.** `packages/morpho-sdk/vitest.config.ts` includes both `src/**/*.test.ts` and `test/**/*.test.ts` (verified). Code extracted into `morpho-sdk` under Phases 4–5 must ship colocated (`foo.ts` ↔ `foo.test.ts`), and tests migrated from `bundler-sdk-viem/test`/`simulation-sdk/test` must land next to their new source file rather than under a top-level `test/` folder. Add an explicit bullet here.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+runtime surfaces it currently imports or exposes through its higher-level actions, plus the full
+bundler ABI surface that consumers may need after `bundler-sdk-viem` is deprecated.
+
+From `packages/bundler-sdk-viem` extract into `morpho-sdk`:
+
+- `src/BundlerAction.ts`: `BundlerAction.encodeBundle`, `BundlerAction.encode`, `BundlerCall`, and
+  the action encoders needed by current `morpho-sdk` action builders: `nativeTransfer`,
+  `wrapNative`, `erc20TransferFrom`, `erc20Transfer`, `permit`, `approve2`, `transferFrom2`,
+  `erc4626Deposit`, `erc4626Redeem`, `morphoSupplyCollateral`, `morphoBorrow`, `morphoRepay`,
+  `morphoWithdrawCollateral`, and `publicAllocatorReallocateTo`.
```
0xbulma, line 284: **[MEDIUM] Protocol-specific terms appear without gloss.** `Permit2`, `approve2`, `transferFrom2`, `ERC-2612` (implied by `permit`), `Paraswap` (L324, 337, 357), `GeneralAdapter1` (implied by `generalAdapter1Abi` L325), `bundler3` (L325) — none are defined in the TIB or linked. `packages/morpho-sdk/CLAUDE.md` references some (e.g., `GeneralAdapter1`, `bundler3`) but no formal glossary location is established. Either (a) add a one-line *Glossary* subsection to this TIB, or (b) inline parentheticals on first use (e.g., `Permit2 (Uniswap signature-based approval contract)`).

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- `src/operations.ts`: no broad operation unions are required for `morpho-sdk` reallocations after
+  the local reallocation data shape exists; keep `BlueOperation`, `MetaMorphoOperation`,
+  `Erc20Operation`, `ParaswapOperation`, and dispatcher types out of `morpho-sdk`.
+- `src/helpers/mutative.ts` and `src/helpers/paraswap.ts`: no utilities are required for the
+  current extraction. Do not extract `getCurrent`, `augustusV6_2Address`,
+  `paraswapContractMethodOffsets`, or `getParaswapContractMethodOffsets` unless Paraswap workflows
+  move into `morpho-sdk`.
+
+The implementation should add parity tests before removing either dependency: bundle calldata
+parity for every extracted `BundlerAction` action type, approval constant parity, and reallocation
```
0xbulma: **[HIGH] "Bundle calldata parity for every extracted `BundlerAction` action type" is too generic.** Today there are zero inline-snapshot calldata tests in `packages/bundler-sdk-viem/test` (verified: grep for `toMatchInlineSnapshot`/`toMatchSnapshot` returns no matches). Require explicitly: (a) one inline-snapshot test per extracted encoder (`nativeTransfer`, `wrapNative`, `erc20TransferFrom`, `erc20Transfer`, `permit`, `approve2`, `transferFrom2`, `erc4626Deposit`, `erc4626Redeem`, `morphoSupplyCollateral`, `morphoBorrow`, `morphoRepay`, `morphoWithdrawCollateral`, `publicAllocatorReallocateTo`) capturing the exact encoded hex; (b) a cross-package parity test asserting byte-identical `Hex` output from `bundler-sdk-viem` (pre-extraction) and `morpho-sdk` (post-extraction) for the same input — this is the contract that protects integrators during the deprecation window; (c) fixture coverage of edge cases per action (zero amount, `MaxUint256`, `address(0)` where allowed, infinite Permit2 allowance via `MathLib.MAX_UINT_160`).

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+output parity for friendly and aggressive public allocator paths.
```
0xbulma: **[MEDIUM] "Approval constant parity" has no specified test shape.** Require: per-`ChainId` deep-equality assertion for `APPROVE_ONLY_ONCE_TOKENS` and `MAX_TOKEN_APPROVALS` between the original `simulation-sdk` exports and the new `morpho-sdk` exports — set membership, keys, and values must match. `APPROVE_ONLY_ONCE_TOKENS` is consumed by `packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Approval.ts` today; a silent drift here changes approval routing and trips the security invariant called out in the Security section.

0xbulma: **[HIGH] "Reallocation output parity for friendly and aggressive public allocator paths" is too generic.** Require: (a) golden/byte-identical deep-equality on the returned `VaultReallocation[]` against the original `simulation-sdk` `computeReallocations` for the same fixture inputs, gating the deprecation of `simulation-sdk`; (b) fork-based parity at a pinned mainnet block exercising the friendly-only path and the friendly→aggressive escalation path via `@morpho-org/test`'s `createViemTest` harness; (c) property-based tests with `fast-check` over vault count, withdrawal amounts, fees, and target utilization. Today's `reallocations.test.ts` covers fork-borrow but not pre/post-extraction parity.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+| `@morpho-org/simulation-sdk-wagmi` | None | `Deprecated: no replacement package. First-party Wagmi hooks are no longer a supported public SDK surface.` |
+| `@morpho-org/test-wagmi` | None | `Deprecated: no replacement package. This public Wagmi test helper package is no longer supported.` |
+
+If additional migration SDK packages are introduced or discovered before implementation, include
+them in the same migration SDK deprecation campaign with replacement `None` unless a later TIB
+explicitly adds the workflow to `morpho-sdk`.
+
+### npm Deprecation Scope
+
+Deprecate every currently published `@morpho-labs/*` package on npm. Do not unpublish these
```
0xbulma, line 355: **[MEDIUM] `@morpho-labs/*` npm deprecation campaign is scope bloat for a consolidation TIB.** L390–407 lists 8 legacy ethers-era packages with bespoke deprecation messages. This is unrelated to consolidating `morpho-sdk` consumption — it needs `@morpho-labs` publish permissions (per L508) and inflates change risk + rollout coordination without advancing the stated goal of "lower the number of packages consumers need to install" (L22–24). Split into a separate maintenance-only TIB or PR so the consolidation can ship without waiting on legacy-scope deprecation.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+| `@morpho-labs/gitbook-cli` | None | `Deprecated: no replacement package.` |
+| `@morpho-labs/gnosis-tx-builder` | None | `Deprecated: no replacement package.` |
+| `@morpho-labs/morpho-ethers-contract` | `@morpho-org/morpho-sdk` | `Deprecated: use @morpho-org/morpho-sdk for maintained Morpho SDK APIs.` |
+| `@morpho-labs/morpho-rewards` | `@morpho-org/morpho-sdk` | `Deprecated: use @morpho-org/morpho-sdk for maintained Morpho SDK APIs and ABIs.` |
+| `@morpho-labs/v2-deployment` | None | `Deprecated: no replacement package.` |
+
+If new `@morpho-labs/*` packages are discovered before implementation, include them in the same
+deprecation campaign.
+
+### Implementation Phases
```
0xbulma, line 374: **[MEDIUM] Phase ordering and parallelism not documented.** Phases 1–8 are listed sequentially but several have no dependency on each other (e.g., Phase 3 — peer-dep maintenance — has no code-landing dependency on Phase 4). Phase 8 conflates two independent workstreams: workspace `@morpho-org/*` deprecation (which depends on Phases 4–6) and `@morpho-labs/*` npm deprecation (which has no dependency on any earlier phase and could run in parallel from Phase 1). Add a dependency graph or explicit "can run in parallel" notes per phase.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+
+- **Phase 1 -- Add canonical consumer path:** Update `morpho-sdk` to depend directly on
+  `blue-sdk-viem`, `blue-sdk`, and `morpho-ts` where needed, and update docs to present
+  `pnpm add @morpho-org/morpho-sdk` as the primary Morpho application install path.
+- **Phase 2 -- Re-export dependency surfaces:** Re-export every fetcher from `blue-sdk-viem` and
```
0xbulma: **[HIGH] Phase 2 contradiction with the kept-maintained packages.** Phase 2 says re-export **every** fetcher from `blue-sdk-viem` and **the classes** from `blue-sdk` through `morpho-sdk` subpaths. Yet `blue-sdk`/`blue-sdk-viem` are NOT in the Workspace Deprecation Scope (L367–386) and stay maintained per Goal at L41–42. If every fetcher and class is mirrored through `morpho-sdk`, what is the user-facing reason to import `blue-sdk`/`blue-sdk-viem` directly? Either explicitly state that they remain for **non-morpho-sdk monorepo consumers** (`liquidation-sdk-viem`, `liquidity-sdk-viem`) while application developers use `morpho-sdk` subpaths, or note that the two paths are deliberately equivalent. Without one of these, the contract is ambiguous.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+  the classes from `blue-sdk` through the canonical `morpho-sdk` subpath hierarchy, especially
+  `@morpho-org/morpho-sdk/fetch`, `@morpho-org/morpho-sdk/entities`, and
+  `@morpho-org/morpho-sdk/utils`, so application code can keep a single Morpho package entrypoint.
```
0xbulma: **[MEDIUM] Subpath import-reachability smoke tests are missing as a Phase 2 deliverable.** 13 new subpaths but no required smoke tests. Add a Phase 2 deliverable: for each new subpath, (a) a runtime import test that resolves a representative symbol from `@morpho-org/morpho-sdk/<subpath>`; (b) exercised under **both** ESM and CJS resolution (per the `lib/esm`/`lib/cjs` dual-publish in the export map at L180–248); (c) a downstream-style `tsc --noEmit` consumer fixture that validates type reachability. `packages/CLAUDE.md` says: *"Subpath exports need both package exports and TS path support."*

0xbulma: **[MEDIUM] JSDoc coverage on re-exports unaddressed.** `packages/morpho-sdk/CLAUDE.md` says: *"Add JSDoc for exported functions and interfaces, and follow neighboring patterns before introducing abstractions."* With ~50 new re-exports through the consolidated subpaths, this is real burden. Add a Phase 2 deliverable: re-exports inherit upstream JSDoc where available, alias re-exports (collision-resolved names) add a fresh JSDoc block in `morpho-sdk`, and `pnpm jsdoc:coverage` baseline is captured before Phase 2 and verified non-regressed after.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- **Phase 3 -- Preserve peer dependencies elsewhere:** Keep `blue-sdk`, `blue-sdk-viem`, and
+  `morpho-ts` as peer dependencies for packages other than `morpho-sdk` unless a separate package
+  decision changes that relationship.
+- **Phase 4 -- Extract bundler and simulation prerequisites:** Move the explicit extraction targets
+  above into `morpho-sdk`, keep the extracted public API behind `morpho-sdk` exports, and add parity
+  tests before any deprecation notice is published.
```
0xbulma: **[LOW] Coverage baseline-compare is unspecified.** Phase 4 says "add parity tests before any deprecation notice is published" but never mentions coverage. Root `vitest.config.ts` reports lcov with no threshold gate today. For every file moved into `morpho-sdk` under Phases 4–5, the coverage on the moved file must meet-or-exceed the baseline measured on the original file in the source package (captured before extraction). Without a baseline-and-compare clause, coverage can regress silently when tests are not migrated alongside the source.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- **Phase 5 -- Simplify simulation and reallocations:** Replace `morpho-sdk`'s `SimulationState`
+  dependency with local reallocation helpers. Move approval constants, public allocator option
+  types, and other narrow constants out of `simulation-sdk` into their owning APIs.
+- **Phase 6 -- Migration SDK deprecation prep:** Retain no migration APIs. Document that migration
+  SDK workflows are no longer a supported public SDK surface and deprecate
+  `@morpho-org/migration-sdk-viem` with replacement `None`.
+- **Phase 7 -- Release workflow updates:** Ensure changesets and release tooling cascade dependent
+  package versions when a direct dependency range changes or when a dependent package must guarantee
+  a newly published internal dependency version.
```
0xbulma: **[MEDIUM] Phase 7 (release tooling) lands too late.** Changeset cascade for direct-dependency range flips first matters in Phase 4 — extracted code lands in `morpho-sdk` and dependent packages (`migration-sdk-viem`, `liquidity-sdk-viem`, `liquidation-sdk-viem`) need version bumps to guarantee compatible internal-dep ranges. If Phase 7 lands after Phase 4, the first extraction releases ship with hand-managed version bumps and dependent packages can publish against stale ranges, breaking downstream lockfiles silently. Move release-tooling work before Phase 4, or gate Phase 4 entry on Phase 7's cascade tooling being live.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- npm publish permissions are required for replacement releases and deprecation notices across both
+  the `@morpho-org` and `@morpho-labs` npm scopes.
+
+## Security
+
+Consolidation changes import paths and package ownership, but it must not weaken transaction
+encoding, authorization, approval, permit, or reallocation invariants.
+
+- Reallocation simplification must preserve public allocator checks, max-in/max-out constraints,
+  supply cap handling, sorted withdrawals, fee accounting, and target-market exclusion.
```
0xbulma, line 495: **[MEDIUM] Security invariants listed prose-only — restate as binding test deliverables.** Root CLAUDE.md says: *"Codify security-sensitive behavior as tests, especially routing, authorization, chain, and accounting invariants."* `packages/morpho-sdk/CLAUDE.md` says: *"Security invariants are part of the SDK contract; test routing, authorization, chainId, builder/signer equality, LLTV buffers, and accounting behavior."* Restate each Security bullet as a binding test that fails if the invariant is removed: (a) public-allocator max-in/max-out, (b) supply-cap respect, (c) sorted withdrawals, (d) fee accounting (tx.value includes total fees), (e) target-market exclusion, (f) calldata byte-parity per extracted action, (g) approval/permit non-arbitrary-spender (per `packages/morpho-sdk/CLAUDE.md`: *"must not permit arbitrary spenders when GeneralAdapter1 is required"*). Mark each as gating Phase 4 / Phase 5 sign-off.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- Moving selected bundler helpers into `morpho-sdk` and changing internal dependency ownership must
+  preserve encoded calldata parity for existing tested actions.
+- `morpho-sdk` subpath exports must not introduce hidden side effects beyond existing augmentation
+  behavior.
```
0xbulma, line 499: **[MEDIUM] Security assertion has no enforcement mechanism.** L520–521 states `morpho-sdk` subpath exports "must not introduce hidden side effects beyond existing augmentation behavior" but specifies no detection. A reviewer cannot catch a regression here without one. Mandate (a) a `sideEffects` allow-list audit in CI that fails when a new module is added to the `sideEffects` array without changelog justification; (b) per-subpath isolated-import test that asserts the import graph reaches no module beyond the documented augmentation entry.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+
+## Observability
+
+Track migration success through:
+
+- npm download trends for deprecated packages, including the legacy `@morpho-labs/*` packages,
+  versus their replacement packages where applicable;
+- npm download trends for `@morpho-org/migration-sdk-viem` after its no-replacement deprecation;
+- support issues caused by import path migration or missing exports;
+- package install size and dependency count for the canonical consumer path;
```
0xbulma, line 513: **[MEDIUM] Bundle-size observability has no target or trigger.** "Package install size and dependency count for the canonical consumer path" is a metric without a budget. With `morpho-sdk` absorbing `blue-sdk`/`blue-sdk-viem`/`morpho-ts` as direct deps plus bundler/simulation extracts, size grows materially. Set: (a) a baseline measurement at Phase 1, (b) a maximum regression delta (e.g., +20% gzipped before a justification is required), (c) a CI gate wired in Phase 1/2 (`size-limit`, `bundlewatch`, or equivalent) that fails PRs exceeding the budget. Without a number, the metric is decorative.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: No planed addition on these in the short term

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Easier to keep the metric decorative because it is purely for human appreciation

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Should be part of the test coverage TIB

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Yes, during the migration it's fine

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: No need, the spec is clear enough

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: The file will just be moved along with the tests

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Yes, still we keep this part of the TIB

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Yes, just discussed w Foulks, do we require this level of precision in a TIB? IMO phase ordering is up to the implementation

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Ignore jsdoc because this is only re-wiring

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+        "types": "./lib/esm/errors/index.d.ts",
+        "import": "./lib/esm/errors/index.js",
+        "require": "./lib/cjs/errors/index.js"
+      },
+      "./fetch": {
+        "types": "./lib/esm/fetch/index.d.ts",
+        "import": "./lib/esm/fetch/index.js",
+        "require": "./lib/cjs/fetch/index.js"
+      },
+      "./helpers": {
```
0xbulma: **[HIGH] `./helpers` is shipped as a public subpath but missing from the canonical hierarchy.** L231–235 declares `./helpers` in `publishConfig.exports`, but the prose at L132–170 enumerates only 12 subpaths and never names `./helpers`. L265 then describes `src/utils/index.ts` as re-exporting `morpho-sdk` helper functions from `src/helpers/index.ts`, implying helpers fold *into* `/utils`. As written, the TIB ships two public entry points for the same set of symbols, doubling JSDoc + binding-test surface (see L417). Resolve: either drop `./helpers` from `publishConfig.exports` (keep `src/helpers/` as an internal source dir), or add it to the canonical hierarchy with an explicit relationship to `/utils`.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+separate package subpaths such as `./blue-utils`, `./viem-utils`, `./sdk-types`, or
+`./blue-types`.
+
+- `src/utils/index.ts` is the merged utility barrel. It re-exports current `morpho-sdk` helper
+  functions from `src/helpers/index.ts`, `CapacityLimitReason` and `CapacityLimit` from `blue-sdk`,
+  and `safeParseNumber`, `safeParseUnits`, `safeGetAddress`, `restructure`, and
+  `readContractRestructured` from `blue-sdk-viem`.
+- `src/types/index.ts` remains the merged type barrel. It keeps the existing `morpho-sdk` type
+  exports and adds `Address`, `MarketId`, `BigIntish`, `Loadable`, `Failable`, `Fetchable`,
+  `TransactionType`, `ChainId`, `ChainMetadata`, `FetchParameters`, and
```
0xbulma: **[HIGH] `FetchParameters` shape conflict in the merged `/types` barrel.** `packages/morpho-sdk/src/types/data.ts:3` defines `FetchParameters` without `chainId`, while `packages/blue-sdk-viem/src/types.ts:3` defines a different `FetchParameters` adding `chainId?: number` (verified). L269–272 says `/types` "remains the merged type barrel. It keeps the existing morpho-sdk type exports and adds … `FetchParameters`, `DeploylessFetchParameters`." Either the two non-identical types collide at the barrel (ambient TS2308) or the local one is silently widened with `chainId` — changing every morpho-sdk entity-method signature transitively. Pick a canonical owner; if the `blue-sdk-viem` shape wins, document the entity-method signature change as a Phase 2 breaking change.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
0xbulma, line 254: **[FOLLOWUP L289] Augmentation coverage gap — only 12 of 28 listed entity classes are augmented today.** `packages/blue-sdk-viem/src/augment/` covers exactly 12 classes (`Holding`, `Market`, `MarketParams`, `Position`, `Token`, `User`, `Vault`, `VaultConfig`, `VaultMarketAllocation`, `VaultMarketConfig`, `VaultMarketPublicAllocatorConfig`, `VaultUser` — verified by `ls`). The TIB `/entities` list at L148–156 enumerates **28** classes including `VaultV2`, `AccrualVaultV2`, `PreLiquidationParams`/`PreLiquidationPosition`, and the six VaultV2-adapter classes — none of which have augmentation today. Consumers attempting `VaultV2.fetch(...)` or `PreLiquidationPosition.fetch(...)` via `@morpho-org/morpho-sdk/entities` will get `undefined` at runtime with no TS error — exactly the silent regression the L289 binding test guards against. Either extend `blue-sdk-viem/src/augment/` to cover the remaining classes before re-export, or state explicitly which entity classes get augmented static fetch methods and which do not.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+not preserving a specific augmentation mechanism for its own sake.
+
+### Reallocation Simplification
+
+Keep the current friendly-then-aggressive public allocator algorithm, but remove the broad
+simulation engine from `morpho-sdk`:
+
+- replace `SimulationState` inputs with a narrow local reallocation data shape containing only the
+  market, vault, vault market config, position, holding, and block data needed by the algorithm;
+- extract small helper functions for candidate withdrawal computation, reallocation state updates,
```
0xbulma, line 265: **[MEDIUM] Per-helper unit tests are missing for the four extracted reallocation helpers.** L300–301 introduces candidate-withdrawal computation, reallocation-state updates, target-market capacity checks, and vault grouping as new helpers. The TIB only requires end-to-end parity (L363–365). Without per-helper unit tests, a regression localized to one helper may cancel out at end-to-end parity (errors offset), or surface only as an obscure failure far from the bug site. `packages/morpho-sdk/CLAUDE.md`: *"Security invariants are part of the SDK contract; test routing, authorization, `chainId`, builder/signer equality, LLTV buffers, and accounting behavior."* State per-helper unit-test deliverables in Phase 5 (L424–426).

docs/tibs/TIB-0002-consolidate-sdk-packages.md
0xbulma: **[FOLLOWUP L364] Moving tests preserves coverage but does not establish cross-package byte parity.** Your reply ("the file will just be moved along with the tests") preserves existing coverage of the pre-extraction encoder, but does not establish byte-identical parity between `bundler-sdk-viem` output (pre-extraction) and `morpho-sdk` output (post-extraction) for the same `ActionArgs` fixture. The deprecation contract for `bundler-sdk-viem` consumers depends on the second guarantee — without it, integrators upgrading from `bundler-sdk-viem` to `morpho-sdk/actions` get encoded-calldata drift with no test signal. Add a cross-package parity test that imports the encoder from BOTH packages at the deprecation tag, encodes the same fixture, and asserts `Hex` equality. Distinct from the in-place test migration.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+- Source and tests are changed under `packages/*/src` and `packages/*/test`; generated `lib/`
+  output is not edited.
+
+## Dependencies
+
+- `viem` remains a peer dependency for consumer-facing Ethereum client and ABI functionality.
+- `@morpho-org/morpho-sdk` depends directly on `@morpho-org/morpho-ts`,
+  `@morpho-org/blue-sdk`, and `@morpho-org/blue-sdk-viem`.
+- `@morpho-org/blue-sdk-viem` keeps `@morpho-org/blue-sdk` as a peer dependency.
+- Packages other than `morpho-sdk` keep `morpho-ts`, `blue-sdk`, and `blue-sdk-viem` as peer
```
0xbulma, line 471: **[MEDIUM] Phase-1 direct-dep flip risks two-class-identity for mixed consumers.** After Phase 1 `morpho-sdk` declares `blue-sdk`/`blue-sdk-viem`/`morpho-ts` as direct deps, while `liquidation-sdk-viem`, `liquidity-sdk-viem`, `migration-sdk-viem` keep them as peers (L500–501). An app that installs `morpho-sdk` + `liquidation-sdk-viem` will satisfy the peer transitively, but pnpm may resolve a different `blue-sdk` version for the peer slot than `morpho-sdk` pinned directly. Result: `instanceof Market` checks across the boundary silently break (two distinct class identities). Require either a single-version `pnpm-workspace.yaml` `overrides` entry, or commit every workspace package to a single release train so direct-dep and peer-dep ranges cannot diverge.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,546 @@
...
+  dependencies when they need those packages.
+- Published packages do not bundle internal dependencies; they publish dependency ranges. A
```
0xbulma: **[HIGH] Cross-package version-bump coordination is unspecified.** L502–507 describes when a dependent package "needs a new release" in prose but names no owner and no enforcement. After Phase 4 extraction, when `blue-sdk` (re-exported through `morpho-sdk/entities`) ships a new class, `migration-sdk-viem`, `liquidity-sdk-viem`, `liquidation-sdk-viem` must be bumped in lockstep — otherwise downstream lockfiles silently pin a stale transitive `blue-sdk`. Mandate either (a) a changeset rule (`morpho-sdk` minor → enumerated peer-dep packages get a patch in the same changeset) or (b) a CI check that diffs the cascade graph and fails when a dependent release is missing. Distinct from the already-posted Phase 7 timing concern at L432; this is about the *content* of the cascade rule, not its phase order.

0xbulma (commented):
Second-pass review: 7 findings deduped against the 27 already-posted comments. Mix of new gaps (`./helpers` subpath inconsistency, `FetchParameters` shape conflict, cross-package version-bump coordination, pnpm two-class-identity, per-helper test coverage) and follow-ups on author replies that left the underlying concern open (L289 augmentation coverage, L364 parity-vs-coverage).

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: There are no clashing error names currently

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: There are no collisions?

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Because there are no other collisions

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: None breaks

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Implementation detail

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: The reader knows

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Makes me think parity tests are overkill. Review is enough

Rubilmax: Not too generic: new implementation will be unit tested and integration tested

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: No package depends on morpho-sdk

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Fine to go without for now IMO

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Fine to go without for now IMO

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Yes but really i think parity tests are overkill, algorithm will be really, really similar and can be reviewed identical

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: To be studied

docs/tibs/TIB-0002-consolidate-sdk-packages.md
Rubilmax: Implementation will be thorough on this

PR #567, opened 2026-04-30
fix(client): validate marketParams.id against tuple at factory boundary

Foulks-Plb:
## Issue (Cantina SDK-107 / MORP2-61)

`MorphoClient.marketV1()` stores the caller-supplied `marketParams` as-is. Reads use `marketParams.id`; writes encode the raw tuple. The SDK never checks that `id` matches the tuple.

## Exploit

Pass a forged plain object where `id = marketA.id` but `loanToken/collateralToken/oracle/irm/lltv` describe market B.

- `getPositionData()` → fetches A (health looks fine).
- User sees A's UI, approves A's token.
- `borrow()` / `repay()` encodes B → transaction lands on the wrong market.

Works silently when A and B share a token (e.g. same loan token → `repay` hits the wrong debt).

## Fix

Recompute the canonical id from the tuple via `MarketUtils.getMarketId` and throw on mismatch at the factory boundary. Genuine `MarketParams` instances from `@morpho-org/blue-sdk` always have a self-consistent `id` (derived in the constructor), so existing callers are unaffected.

## Migration of [morpho-sdk#129](https://github.com/morpho-org/morpho-sdk/pull/129)

Same fix as the standalone-repo PR, adapted for `packages/morpho-sdk/src/client/morphoClient.ts`. Linear: [SDK-107](https://linear.app/morpho-labs/issue/SDK-107/morp2-61medium-morphoclientmarketv1-can-fetch-one-market-while).

cc @Rubilmax

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/567" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/morpho-sdk/src/client/morphoClient.ts
```diff
@@ -38,6 +42,10 @@ export class MorphoClient implements MorphoClientType {
   }
 
   public marketV1(marketParams: MarketParams, chainId: number) {
+    const derivedId = MarketUtils.getMarketId(marketParams);
+    if (marketParams.id !== derivedId) {
+      throw new MarketIdMismatchError(marketParams.id, derivedId);
+    }
```
Rubilmax, line 49: At this point if `marketParams.id` is overridden we could argue it's the caller's fault... It is statically typed `MarketParams`

Or just use `MarketUtils.getMarketId` everywhere down the road but im not sure we need to protect from inconsistent input

Rubilmax: Can happen with hardcoded, agent-written flows

packages/morpho-sdk/src/client/morphoClient.ts
```diff
@@ -38,6 +42,10 @@ export class MorphoClient implements MorphoClientType {
   }
 
   public marketV1(marketParams: MarketParams, chainId: number) {
+    const derivedId = MarketUtils.getMarketId(marketParams);
+    if (marketParams.id !== derivedId) {
```
Rubilmax, line 47:
```suggestion
    // Can happen with one-time/hardcoded/agent-written possibly inconsistent input market params.
    if (marketParams.id !== derivedId) {
```

Rubilmax (approved):

packages/morpho-sdk/src/client/morphoClient.ts
```diff
@@ -38,6 +42,10 @@ export class MorphoClient implements MorphoClientType {
...
+      throw new MarketIdMismatchError(marketParams.id, derivedId);
+    }
     return new MorphoMarketV1(this, marketParams, chainId);
```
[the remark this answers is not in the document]
Foulks-Plb: let's shallow it

PR #569, opened 2026-04-30
fix(marketV1): refetch target market in getReallocationData

Foulks-Plb:
## Issue (Cantina SDK-102 / MORP2-72)

`MorphoMarketV1.getReallocationData()` is documented as fetching all on-chain data needed to plan shared-liquidity reallocations, but it never actually fetched the target market — it accepted an external `market` argument, validated only `market.id`, then assembled that caller-owned snapshot with fresh vault / config / position data fetched at `block.number`. `computeReallocations()` then treats the injected target market as authoritative when deciding whether a reallocation is needed and how much liquidity to source, and `borrow()` blindly prepends the resulting `reallocateTo` actions and sets `tx.value` to the summed PublicAllocator fees.

## Exploit

An attacker controlling a PublicAllocator-enabled vault with a high reallocation fee can temporarily depress the target market's apparent liquidity between the caller's market fetch and `getReallocationData()`, the planner believes a reallocation is required and builds a `reallocateTo` against the attacker's vault, the attacker restores liquidity but the bundle still executes the fee-bearing call and the native fee lands on the attacker's vault. Since the reallocation call itself remains valid on-chain, there is no revert: the victim simply overpays native token.

## Fix

Drop the `market` parameter from `getReallocationData`. The target market is now refetched internally at `block.number` in the same `Promise.all` as the source vaults, so every row of the resulting `SimulationState` comes from the same epoch.

## DevEx

Breaking signature change on the entity: consumers must drop the `market` field from `getReallocationData()` calls. Typical migration is a single-line deletion. `getMarketData()` / `getPositionData()` remain unchanged.

## Migration of [morpho-sdk#132](https://github.com/morpho-org/morpho-sdk/pull/132)

Same fix as the standalone-repo PR, adapted to `packages/morpho-sdk/`. Linear: [SDK-102](https://linear.app/morpho-labs/issue/SDK-102/morp2-72medium-caller-supplied-target-market-snapshots-can-force).

cc @Rubilmax

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/569" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -944,13 +942,18 @@ export class MorphoMarketV1 implements MarketV1Actions {
...
-    );
-
-    // Collect unique market IDs from all vault withdrawQueues + target market.
     const targetMarketId = this.marketParams.id;
+
+    // Phase 1: Fetch the target market and all vaults at `block.number` in
+    // parallel so every row of the resulting state comes from the same epoch
+    // and the planner never trusts a caller-owned target-market snapshot.
+    const [targetMarket, vaults] = await Promise.all([
+      fetchMarket(targetMarketId, client, fetchParams),
```
Rubilmax, line 951: Do we need the full market here? Not only market params?

Rubilmax (approved):

Rubilmax (approved):

PR #570, opened 2026-04-30
fix(entities): validate userAddress on all vault methods

Foulks-Plb:
## Summary

Extends the builder = executor guard to every vault entity method that accepts a `userAddress`, addressing the deferred `[CRITICAL]` / `[HIGH]` findings from the morpho-sdk #122 / #123 reviews.

Before this PR, `validateUserAddress` was wired into the six `MorphoMarketV1` methods (#122) and `MorphoVaultV1.migrateToV2` (#123). The vault deposit / withdraw / redeem paths were still permissive, even though `VaultV1.deposit` / `VaultV2.deposit` share the **exact same mixed-account bundle shape** as the SDK-100/SDK-101 cases — a victim with standing GA1 approval signing an attacker-built tx (`userAddress = attacker`) would have minted shares to the attacker.

## Changes

- `MorphoVaultV1`: `deposit`, `withdraw`, `redeem` now call `validateUserAddress`.
- `MorphoVaultV2`: `deposit`, `withdraw`, `redeem`, `forceWithdraw`, `forceRedeem` now call `validateUserAddress`.
- Consolidated the inline `chainId` checks on the `validateChainId` helper for uniformity with `migrateToV2` / MarketV1.
- Regression tests (`AddressMismatchError`) for each method.

`migrateToV2` is intentionally untouched — it is already covered.

## Migration of [morpho-sdk#133](https://github.com/morpho-org/morpho-sdk/pull/133)

Same fix as the standalone-repo PR, adapted to `packages/morpho-sdk/`.

cc @Rubilmax

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/570" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


0xbulma (commented):

**Reviewed commit:** `753bafa1`

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 1 |
| 🔵 Medium | 1 |
| 🔵 Minor | 0 |

---

### 🟡 Important — `examples/example.ts` will throw post-PR

**File:** [`packages/morpho-sdk/examples/example.ts`](https://github.com/morpho-org/sdks/blob/Foulks-Plb/vault-validate-user/packages/morpho-sdk/examples/example.ts#L29-L36)

The canonical SDK example creates a wallet client without an `account`:

```ts
const walletClient = createWalletClient({
  chain: mainnet,
  transport: http(MAINNET_RPC_URL),
});
const morpho = new MorphoClient(walletClient);
// ...
const USER_ADDRESS: Address = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266";
const deposit = vault.deposit({ amount, userAddress: USER_ADDRESS, accrualVault }); // line 51
const withdraw = vault.withdraw({ amount, userAddress: USER_ADDRESS });            // line 75
const redeem = vault.redeem({ shares, userAddress: USER_ADDRESS });                // line 90
```

With this PR, every `vault.*` call routes through `validateUserAddress(this.client.viemClient.account?.address, userAddress)`. Because `walletClient.account` is undefined, the helper throws `MissingClientPropertyError("account")` (`packages/morpho-sdk/src/helpers/validate.ts:61-63`) on the first call (`example.ts:51`) — the example becomes unrunnable.

**This is correct enforcement of the documented `builder = signer` invariant** (see `packages/morpho-sdk/README.md`'s “Integration invariant — builder = signer”), but the canonical example then needs to demonstrate the supported pattern:

```ts
import { privateKeyToAccount } from "viem/accounts";
const account = privateKeyToAccount(PRIVATE_KEY); // or generatePrivateKey() with a comment
const walletClient = createWalletClient({ account, chain: mainnet, transport: http(MAINNET_RPC_URL) });
// then: userAddress: account.address
```

Fix the example as part of this PR to keep the docs/behavior consistent. Same concern applies to the README usage snippet (`packages/morpho-sdk/README.md` near “Deposit with native token wrapping”) which shows `userAddress: "0xUser..."` without showing the matching client setup — worth a note there too.

---

### 🔵 Medium — Missing changeset

No changeset under `.changeset/` for this PR. Per [`CLAUDE.md` §7](https://github.com/morpho-org/sdks/blob/main/CLAUDE.md#7-releases--versioning): *“every user-visible change ships with a changeset.”*

This change is **observable at runtime by integrators**: code paths that previously built txs successfully (account-less viem client, or `userAddress !== client.account.address`) now throw `MissingClientPropertyError` / `AddressMismatchError`. Even though the intent is a security fix, integrators relying on the prior permissive behavior need a release note to spot the breakage.

Add e.g. `.changeset/<slug>.md`:

```md
---
"@morpho-org/morpho-sdk": patch
---

fix(entities): vault deposit/withdraw/redeem/forceWithdraw/forceRedeem now enforce `userAddress === client.account.address` (builder = signer invariant). Throws `MissingClientPropertyError` when the client has no account, or `AddressMismatchError` on divergence — closes the SDK-100/SDK-101 mixed-account-bundle vector on the vault paths.
```

Whether this should be `patch` or `minor` depends on how the team weighs *security fix* vs. *behavior change* — defer to maintainers, but don't ship without one.

---

### Notes (not findings)

- **Pattern matches `MorphoMarketV1`** — `validateChainId` + `validateUserAddress` ordering is identical to `marketV1.ts` (e.g. `supplyCollateral`, `borrow`, `repay`). ✅
- **`getData()` intentionally not consolidated** — both `MorphoVaultV1.getData()` and `MorphoVaultV2.getData()` keep the inline `if (chain?.id && chain?.id !== this.chainId) throw new ChainIdMismatchError(...)` pattern, which is *more lenient* than `validateChainId` (allows clients with no chain configured). The PR description says the inline checks were “consolidated… for uniformity with `migrateToV2` / MarketV1” but `getData` is not. That's defensible (read paths permit no-chain public clients) but worth either a comment on the helper or a follow-up to align — `MorphoMarketV1.getMarketData` uses the strict helper. Calling it out so it's a deliberate choice and not an oversight.
- **`ChainIdMismatchError` import retained correctly** — still referenced by `getData()` in both files, so it's not an unused import.
- **Existing tests unaffected** — every call site in `packages/morpho-sdk/test/actions/vaultV{1,2}/*.test.ts` and `wrap.test.ts` already uses `userAddress: client.account.address`, so no regressions.
- **Test coverage** — the 8 new `AddressMismatchError` tests cover the full surface called out in the PR (V1: deposit/withdraw/redeem; V2: deposit/withdraw/redeem/forceWithdraw/forceRedeem). `migrateToV2` is intentionally untouched and is already covered upstream. ✅

_This is an automated parallel review on commit `753bafa1`._

PR #571, opened 2026-04-30
fix(blue-sdk-viem): reject VaultV1 liquidity adapter with non-empty data (SDK-153)

Foulks-Plb:
## Summary

Fix [SDK-153](https://linear.app/morpho-labs/issue/SDK-153/morp2-6low-vaultv2-fetchers-model-vaultv1-liquidity-adapters-as) (Cantina audit, Low).

A VaultV2 using a `MorphoVaultV1Adapter` as `liquidityAdapter` requires `liquidityData == "0x"` onchain — the adapter rejects any non-empty bytes with `InvalidData()`. The fetchers were assembling a fully usable `VaultV2` / `AccrualVaultV2` in that broken state, so downstream tooling reported normal caps/liquidity even though every deposit (auto-allocating through the liquidity adapter) and every withdrawal that needs deallocation would revert.

`fetchVaultV2` now throws `UnsupportedVaultV2AdapterError(liquidityAdapter)` when the liquidity adapter is a `MorphoVaultV1Adapter` and `liquidityData !== "0x"`, in both the deployless and multicall paths.

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1777906611026089)


packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -63,6 +64,20 @@ export async function fetchVaultV2(
           ],
         });
 
+      if (
+        vault.liquidityData !== "0x" &&
+        morphoVaultV1AdapterFactory != null &&
+        vault.liquidityAdapter !== zeroAddress &&
+        (await readContract(client, {
```
0xbulma: Not sure we want this kind of syntax no?

0xbulma: await in the if condition should not be inlined.

Foulks-Plb: Right

Rubilmax (approved):

PR #577, opened 2026-04-30
fix(simulation-sdk): align Blue_Borrow slippage with onchain minSharePrice

Foulks-Plb:
Fixes [SDK-119](https://linear.app/morpho-labs/issue/SDK-119/morp2-64low-blue-borrow-simulation-underestimates-allowed-slippage-and)

## Summary

- The `Blue_Borrow` simulator approximated post-slippage state with `expectedShares * (WAD + slippage)` (asset-mode) and `expectedAssets / (WAD + slippage)` (share-mode), which is strictly more favorable than the `minSharePrice = (WAD - slippage)` floor enforced by `GeneralAdapter1.morphoBorrow`. For any positive slippage, real onchain borrows can mint up to `expectedShares / (WAD - slippage)` shares (or return as few as `expectedAssets * (WAD - slippage)` assets), so the pre-fix simulator overstated borrower health and credited liquidity, letting downstream planners greenlight bundles that revert onchain within the configured slippage window.
- Mirror the onchain bound: book the worst-case shares/assets that `minSharePrice` still permits, using `MathLib.wDivUp` / `MathLib.wMulDown` for conservative rounding.
- Reject `slippage >= WAD` locally (which would otherwise divide by zero) via `SimulationErrors.InvalidInput`.

## Concrete example (s = 20%, expected = 100 shares)

| | Onchain (real) | Pre-fix simulation | Post-fix simulation |
|---|---|---|---|
| Shares booked | up to **125** | 120 | **125** ✓ |


packages/simulation-sdk/src/handlers/blue/borrow.ts
```diff
@@ -31,6 +31,9 @@ export const handleBlueBorrowOperation: OperationHandler<
   if ((assets === 0n) === (shares === 0n))
     throw new BlueErrors.InconsistentInput(assets, shares);
 
+  if (slippage >= MathLib.WAD)
```
0xbulma: **[Important]** This local guard duplicates (and partially overlaps) the central one in `handlers/blue/index.ts:33-37`:

```ts
if (slippage < 0n || slippage > MathLib.WAD)
  throw new SimulationErrors.InvalidInput({ slippage });
```

The only case the central guard misses is `slippage === MathLib.WAD`, which post-fix would make `WAD - slippage = 0` and trip `mulDiv` with `DIVISION_BY_ZERO` (a generic `Error`, not a typed simulation error). That's exactly the case this local guard catches.

Suggestion: tighten the central check to `slippage < 0n || slippage >= MathLib.WAD` and drop this local one — single source of truth for the slippage domain across all Blue ops. As-is, two separate guards can drift independently.

packages/simulation-sdk/src/handlers/blue/borrow.ts
```diff
@@ -31,6 +31,9 @@ export const handleBlueBorrowOperation: OperationHandler<
...
+    throw new SimulationErrors.InvalidInput({ slippage });
```
0xbulma: **[Important]** The same `WAD + slippage` pattern this PR fixes for borrow is still present in sibling `packages/simulation-sdk/src/handlers/blue/withdraw.ts` (L51-58). Onchain, `GeneralAdapter1.morphoWithdraw` enforces `minSharePrice = WAD - slippage` (see `bundler-sdk-viem/src/actions.ts:660-664` and `morpho-sdk/src/helpers/slippage.ts`), so the withdraw simulator currently understates worst-case shares burned and overstates worst-case assets received — the same class of bug the audit flagged for borrow.

Suggestion: fix `withdraw.ts` in this PR using the symmetric transform (`wDivUp`/`wMulDown` against `WAD - slippage`), or open a follow-up issue with the same audit reference and link it from the changeset so the divergence is intentional and tracked.

packages/simulation-sdk/src/handlers/blue/borrow.ts
```diff
@@ -44,13 +47,13 @@ export const handleBlueBorrowOperation: OperationHandler<
   if (market.price == null) throw new BlueErrors.UnknownOraclePrice(id);
 
   if (shares === 0n)
-    shares = MathLib.wMulUp(
+    shares = MathLib.wDivUp(
```
0xbulma, line 61: **[Important]** Tests for this audit fix existed and were explicitly deleted in commit `f6660fb1` ("chore: drop borrow slippage explainer comment and tests", -119 lines in `borrow.test.ts`). The PR now ships the math change without coverage. Per CLAUDE.md §5 'Security invariants are tests' and §10 'PRs that ship without their tests, JSDoc, and changeset' don't land.

Suggestion: restore the three deleted tests (or equivalents):
1. `should book worst-case shares from minSharePrice floor in asset-mode with positive slippage` — asserts `wDivUp(toBorrowShares, WAD - slippage)`, with the explicit comparison `expect(expectedShares).toBeGreaterThan(optimisticShares)` so a sign-flip back to `WAD + slippage` fails the suite.
2. `should book worst-case assets from minSharePrice floor in share-mode with positive slippage` — same idea for L55.
3. `should throw on slippage equal to WAD` — asserts the new `InvalidInput` by class identity (per §5: 'Errors asserted by class identity, not message string'), e.g. `expect(() => simulateOperation(...)).toThrow(SimulationErrors.InvalidInput)`.

0xbulma (commented):

**Reviewed commit:** `e5c1696c`

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 3 |
| 🔵 Medium | 3 |
| 🔵 Minor | 0 |

### Math is correct
The new asset-mode `wDivUp(toBorrowShares, WAD - slippage)` and share-mode `wMulDown(shares, WAD - slippage)` correctly mirror `GeneralAdapter1.morphoBorrow`'s `minSharePrice = WAD - slippage` floor (verified against `bundler-sdk-viem/src/actions.ts` L660–700 and `morpho-sdk/src/helpers/slippage.ts:22`).

### File-level findings (cannot be inlined; lines not in diff hunks)

**🟡 Important — Missing changeset.** This is a user-visible behaviour change in published `simulation-sdk` (sign flip on slippage encoding plus a new typed-error path). No file added under `.changeset/`. Per CLAUDE.md §7 every user-visible change ships with a changeset.

**🔵 Medium — Missing JSDoc on touched export.** `handleBlueBorrowOperation` (L10 in `borrow.ts`) is exported but has no JSDoc. The PR introduces a new failure mode (`SimulationErrors.InvalidInput` for `slippage >= WAD`) that integrators may pattern-match on. Per CLAUDE.md §6 every exported symbol needs JSDoc with `@throws` for each typed error class, and §9 'on refactor, adopt the convention' applies on touch. Suggestion: add a short JSDoc covering `@param`, `@returns`, and `@throws` for each typed error (`SimulationErrors.InvalidInput`, `BlueErrors.InconsistentInput`, `BlueErrors.UnknownOraclePrice`, `BlueErrors.InsufficientLiquidity`, `BlueErrors.InsufficientCollateral`, `BlueSimulationErrors.UnauthorizedBundler`).

**🔵 Medium — Stale convention example.** `packages/simulation-sdk/CLAUDE.md` line 13 reads:
> Encode slippage as WAD-scaled `bigint`, e.g. `MathLib.WAD + slippage`.

This is now contradicted by the touched handler. Update the example to reflect that borrow (and withdraw onchain) uses `WAD - slippage` to mirror `minSharePrice`, while supply/repay use `WAD + slippage` to mirror `maxSharePrice`. Or generalize the wording so it doesn't pin a sign.

_This is an automated parallel review. See inline comments below for line-anchored findings._

Rubilmax (commented):
There is no slippage onchain, so there's no unique way to define the slippage. Here it seems not worthy to chnge the definition of slippage in these cases considering this package will get deprecated anyway

IMO not a bug, worth closing

PR #578, opened 2026-04-30
fix(marketV1): accrue market before sizing share-mode repay

Foulks-Plb:
Fixes [SDK-121](https://linear.app/morpho-labs/issue/SDK-121/morp2-59low-share-mode-repay-transferamount-is-computed-from-unaccrued)

## Summary

- Share-mode `repay({ shares })` and `repayWithdrawCollateral({ shares })` derived `transferAmount` and `maxSharePrice` from `positionData.market.toBorrowAssets(shares, "Up")` without first accruing the market. Because `fetchAccrualPosition` returns the raw `lastUpdate` snapshot, on-chain `morphoRepay` required more assets than the bundler had pre-pulled and the supposedly accrual-immune full-repay reverted on dormant markets.
- Accrue `positionData.market` to `Time.timestamp() + 10 min` (mirroring the existing `repayWithdrawCollateral` health-check buffer) and size both `transferAmount` and `maxSharePrice` from the accrued market. Residual loan tokens are skimmed back to the receiver, so over-buffering is safe.
- Adds dormant-market regression tests for both entry points: borrow, advance the chain by 30 days, run the SDK repay flow end-to-end, and assert the position closes.

## Why

The SDK explicitly markets share-mode full-repay as immune to interest accrual. Without this fix, a user trying to close their position to escape liquidation on a market whose `lastUpdate` lags meaningfully behind the current block sees the transaction revert at `morphoRepay` even with default slippage and immediate signing.

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/578" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


0xbulma (commented):

**Reviewed commit:** `1d959c6a`

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 5 |
| 🔵 Medium | 5 |
| 🔵 Minor | 2 |

### Out-of-hunk findings

The GitHub review API rejects inline comments outside diff hunks, so the four findings below are surfaced here in the body. They all sit within the public surface this PR changes the behavior of.

**🟡 Important — Missing changeset (`.changeset/`)**
Per root `CLAUDE.md` §7 (and §2.10) every user-visible change ships with a changeset. This PR changes the on-chain calldata produced by `repay({ shares })` and `repayWithdrawCollateral` (transferAmount sized from a 2h-forward-accrued market; `maxSharePrice` derived from the same projection with `slippageTolerance` zeroed) and tightens the assets-mode `repayWithdrawCollateral` health check (10 min → 2 h buffer). Add a `patch` or `minor` changeset describing the behavior change.

**🟡 Important — Stale JSDoc on `repay` (`packages/morpho-sdk/src/entities/marketV1/marketV1.ts:161`)**
The bullet *“**Shares mode:** `slippageTolerance` also caps `transferAmount`.”* is now factually wrong. After this PR, shares-mode `transferAmount` is sized purely from a 2h-forward-accrued market projection and `sharePriceSlippage` is forced to `0n`, so `slippageTolerance` is silently ignored in shares mode. Replace the bullet with the new contract: *“Shares mode: `transferAmount` and `maxSharePrice` are sized from a market projected 2 h forward (upper-bounding on-chain accrual); the bundle skims residual loan tokens back to the receiver. `slippageTolerance` is not applied in shares mode.”* Also missing per CLAUDE.md §6: `@throws` enumerating typed errors and an `@example` block.

**🔵 Medium — Incomplete JSDoc on `repayWithdrawCollateral` (`packages/morpho-sdk/src/entities/marketV1/marketV1.ts:215`)**
The public docs don't mention the new shares-mode 2h projection, the bundle skim, or that the post-repay health check now evaluates against a market accrued +2 h forward (was +10 min). Add a `**Shares mode**` bullet matching the new `repay` doc, call out the stricter health check (so assets-mode users have warning that previously borderline-healthy `repayWithdrawCollateral({ assets, withdrawAmount })` calls may now throw `WithdrawMakesPositionUnhealthyError`), and add `@throws` / `@example` per §6.

**🟡 Important — `slippageTolerance` silently ignored in shares mode (`packages/morpho-sdk/src/entities/marketV1/marketV1.ts:506` and the parallel destructure in `repayWithdrawCollateral`)**
In shares mode, the user-supplied `slippageTolerance` is validated by `validateSlippageTolerance` and then silently overridden to `0n` (`sharePriceSlippage = 0n`). A caller passing a deliberate non-default `slippageTolerance` with `shares` sees no effect and no warning. Either (a) document the shares-mode no-op on the `slippageTolerance` JSDoc and on the `repay` description, or (b) throw a typed error when `slippageTolerance` is provided together with `shares`. Pick one — silent ignore of a parameter the caller explicitly set is the worst of both worlds.

_Inline comments below cover findings inside diff hunks._

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -536,6 +536,8 @@ export class MorphoMarketV1 implements MarketV1Actions {
     let assets: bigint;
     let shares: bigint;
     let transferAmount: bigint;
+    let marketForRepay: Market;
+    let sharePriceSlippage = slippageTolerance;
```
0xbulma: **[🔵 Minor]** `let marketForRepay: Market;` is declared without an initializer and only assigned inside the if/else; `let sharePriceSlippage = slippageTolerance;` is initialized then overwritten to `0n` in shares mode. Both patterns are fragile to a future refactor (an early return / new branch silently leaves `marketForRepay` undefined or `sharePriceSlippage` mis-set) and harder to scan than a discriminated assignment.

Suggestion: compute both via ternary after the branch sets `transferAmount`:

```ts
const marketForRepay = isSharesMode
  ? positionData.market.accrueInterest(accrualTimestamp)
  : positionData.market;
const sharePriceSlippage = isSharesMode ? 0n : slippageTolerance;
```

Apply the same shape to `repayWithdrawCollateral` (lines 722–723).

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -545,16 +547,14 @@ export class MorphoMarketV1 implements MarketV1Actions {
...
-        shares,
-        "Up",
-      );
-      transferAmount = MathLib.wMulUp(
-        baseTransferAmount,
-        MathLib.WAD + slippageTolerance,
-      );
+      // 2h forward accrual upper-bounds the on-chain repay price; bundle
+      // skims residual back to the receiver.
+      const accrualTimestamp =
```
0xbulma, line 545: **[🔵 Medium]** The shares-mode block here (lines 550–557) is duplicated almost verbatim in `repayWithdrawCollateral` (lines 725–741), and the `Time.s.from.h(2n)` literal appears in two places. Per root CLAUDE.md §1 (single source of truth) and the helpers convention in `packages/morpho-sdk/src/helpers/CLAUDE.md`, extract a small pure helper into `helpers/`:

```ts
// helpers/repayAccrual.ts
export const REPAY_ACCRUAL_BUFFER = Time.s.from.h(2n);

export function projectMarketForRepay(market: Market): {
  marketForRepay: Market;
  accrualTimestamp: bigint;
} {
  const accrualTimestamp =
    MathLib.max(Time.timestamp(), market.lastUpdate) + REPAY_ACCRUAL_BUFFER;
  return {
    marketForRepay: market.accrueInterest(accrualTimestamp),
    accrualTimestamp,
  };
}
```

This (a) removes the `let marketForRepay` boilerplate, (b) gives the 2h constant a documented name, and (c) makes the uniformity between the two call sites explicit (today the buffer happens to be 2h in both places by convention, not by construction).

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -545,16 +547,14 @@ export class MorphoMarketV1 implements MarketV1Actions {
...
+        MathLib.max(Time.timestamp(), positionData.market.lastUpdate) +
+        Time.s.from.h(2n);
+      marketForRepay = positionData.market.accrueInterest(accrualTimestamp);
+      transferAmount = marketForRepay.toBorrowAssets(shares, "Up");
+      sharePriceSlippage = 0n;
```
0xbulma: **[🔵 Minor]** Forcing `sharePriceSlippage = 0n` removes the additive safety margin on `maxSharePrice` that the pre-PR code applied (`WAD + slippageTolerance`). With `slippageTolerance: 0n` and `market = marketForRepay`, `computeMaxRepaySharePrice` just round-trips: `maxSharePrice = ceil((toBorrowAssets(shares,'Up') * RAY) / shares)` — i.e. exactly the projected share price.

Failure mode is benign (revert, not fund loss): if execution timestamp exceeds the 2h projection, `transferAmount` already underfunds the bundler so the bundle reverts before `maxSharePrice` is checked. But on a high-utilization market with intervening accruals, the on-chain post-repay share price can rise slightly above the single-step projection (see `AdaptiveCurveIrmLib`'s underestimation note); a thin additive buffer on `maxSharePrice` only would prevent a needless revert in that window. Consider keeping `slippageTolerance` (or a small named constant) on the `maxSharePrice` computation while still using the 2h projection as the source of truth for the assets bound.

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -718,6 +719,14 @@ export class MorphoMarketV1 implements MarketV1Actions {
     let assets: bigint;
     let shares: bigint;
     let transferAmount: bigint;
+    let marketForRepay: Market;
+    let sharePriceSlippage = slippageTolerance;
+
+    // 2h forward accrual upper-bounds the on-chain repay price (shares
+    // mode) and the post-repay health check; bundle skims residual back.
+    const accrualTimestamp =
```
0xbulma, line 713: **[🔵 Medium]** The accrual-timestamp buffer for `repayWithdrawCollateral` widened from `+10 min` (the previous health-check buffer at the now-deleted lines 758–761) to `+2 h`, and is now applied uniformly to **both** the shares-mode `transferAmount`/`maxSharePrice` sizing **and** the post-repay health check (line 762: `positionData.repay(assets, shares, accrualTimestamp)`).

In assets mode, `transferAmount = params.assets` is fixed by the user — only the post-repay health check sees the wider buffer. `validatePositionHealthAfterWithdraw` will now evaluate against a market 2 h forward (more accrued debt → higher residual borrow → stricter LLTV check), so an assets-mode `repayWithdrawCollateral({ assets, withdrawAmount })` that sat just under the LLTV+buffer threshold under the old 10-min projection can now throw `BorrowExceedsSafeLtvError`/`WithdrawMakesPositionUnhealthyError`.

The shares-mode bug fix did not require widening the assets-mode health-check buffer. Two paths:

1. **Preserve assets-mode behavior**: keep `+10 min` for the post-repay health check and apply `+2 h` only when sizing the shares-mode `marketForRepay`. Easier to do once the helper is extracted (see line 552 comment).
2. **Accept the stricter check**: document the change in JSDoc (`@throws WithdrawMakesPositionUnhealthyError`) and the changeset, and add an assets-mode regression test exercising a borderline-healthy position to lock in the new safety boundary.

Either is fine; silent widening is not.

packages/morpho-sdk/test/actions/marketV1/repay.test.ts
```diff
@@ -265,6 +266,81 @@ describe("RepayMarketV1", () => {
     );
   });
 
+  // Regression: repay-by-shares used to size transferAmount from the stale
```
0xbulma, line 273: **[🔵 Medium]** Per root CLAUDE.md §5 (`morpho-sdk` is wired for colocated tests via `vitest.config.ts`) and §9 ("on refactor, adopt the convention"), this PR refactors `entities/marketV1/marketV1.ts` and the new regression should land colocated next to source (e.g. `packages/morpho-sdk/src/entities/marketV1/marketV1.test.ts`).

The existing legacy file is acknowledged in `packages/morpho-sdk/CLAUDE.md` ("migrate them next to source on refactor"), so the migration of these tests is in scope for this PR. If a full file move is too large to bundle here, at minimum document the deferred migration in the package `AGENTS.md` so the next refactor picks it up.

packages/morpho-sdk/test/actions/marketV1/repay.test.ts
```diff
@@ -265,6 +266,81 @@ describe("RepayMarketV1", () => {
...
+          throw new Error("Approval requirement not found");
+        }
+        await client.sendTransaction(approval);
+
+        const tx = repay.buildTx();
+        await client.sendTransaction(tx);
+      },
+    });
+
+    expect(finalState.position.borrowShares).toBe(0n);
```
0xbulma, line 349: **[🟡 Important]** Test asserts only `borrowShares == 0n` and `collateral == collateralAmount` — it does **not** verify the PR's central safety claim that *“residual loan tokens are skimmed back to the receiver, so over-buffering is safe.”* With a 30-day forward accrual and a 100,000-unit pre-fund, this test would still pass if the bundle silently kept the over-pulled tokens in `GeneralAdapter1` (extractable by anyone, or just stuck).

Add an assertion that the user got back the unused buffer. The cleanest invariant is `userLoanTokenDelta == -morphoLoanTokenDelta`:

```ts
expect(initialState.userLoanTokenBalance - finalState.userLoanTokenBalance)
  .toEqual(finalState.morphoLoanTokenBalance - initialState.morphoLoanTokenBalance);
```

Same fix needed in `repayWithdrawCollateral.test.ts` line 320.

packages/morpho-sdk/test/actions/marketV1/repayWithdrawCollateral.test.ts
```diff
@@ -249,6 +250,80 @@ describe("RepayWithdrawCollateralMarketV1", () => {
     );
   });
 
+  // Regression: same accrual bug as repay({ shares }) — transferAmount used
```
0xbulma, line 255: **[🔵 Medium]** Same colocation concern as the sibling repay test (see comment on `repay.test.ts:269`). New test added to legacy `test/` location during a refactor of `entities/marketV1/marketV1.ts`; per CLAUDE.md §5 and §9 the file should land colocated with source, or the deferred migration should be documented in the package `AGENTS.md`.

packages/morpho-sdk/test/actions/marketV1/repayWithdrawCollateral.test.ts
```diff
@@ -249,6 +250,80 @@ describe("RepayWithdrawCollateralMarketV1", () => {
...
+            await client.sendTransaction(req);
+          }
+        }
+
+        const tx = action.buildTx();
+        await client.sendTransaction(tx);
+      },
+    });
+
+    expect(finalState.position.borrowShares).toBe(0n);
```
0xbulma, line 327: **[🟡 Important]** Same gap as `repay.test.ts:340` — asserts position closure and collateral receipt but never checks that residual loan tokens were skimmed back to the receiver. The PR's safety claim is unverified; a regression where `GeneralAdapter1` stops skimming would not fail this test.

Add a delta-balanced loan-token assertion, e.g.:

```ts
expect(initialState.userLoanTokenBalance - finalState.userLoanTokenBalance)
  .toEqual(finalState.morphoLoanTokenBalance - initialState.morphoLoanTokenBalance);
```

Also worth adding a sibling assets-mode regression on a borderline-healthy position to exercise the widened 10 min → 2 h post-repay health-check buffer (see `marketV1.ts:727` comment).

Rubilmax (dismissed):
No changesets

packages/morpho-sdk/test/actions/marketV1/repay.test.ts
Foulks-Plb: This is already the case in testInvariant.

packages/morpho-sdk/test/actions/marketV1/repayWithdrawCollateral.test.ts
Foulks-Plb: This is already the case in testInvariant.

Rubilmax (dismissed):

Rubilmax (approved):

PR #579, opened 2026-05-03
ci: harden release workflows

tarikbellamine:
Summary:
- narrow reusable workflow permissions and replace broad secret inheritance with explicit secrets
- harden release checkouts, tag-push credentials, Changesets version commit staging, release-branch pushes, and release PR creation
- expand release-critical CODEOWNERS, fix Dependabot target branch, add GitHub Actions updates, and delay consumption of newly published upstream packages by 3 days

Deferred:
- npm-publish environment gate and npm trusted-publisher environment matching are intentionally left for a later PR coordinated with repo/npm settings.

Verification:
- pnpm install --frozen-lockfile
- pnpm lint
- pnpm build
- Ruby YAML parse for edited workflows and Dependabot config
- git diff --check

Note:
- A local pnpm test attempt failed because MAINNET_RPC_URL was unset and anvil is not installed in this environment.

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/579" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


.github/CODEOWNERS
```diff
@@ -1,2 +1,9 @@
-/.github/CODEOWNERS @morpho-org/security
-/.github/workflows/ @morpho-org/security
+/.github/ @morpho-org/security
+/.changeset/ @morpho-org/security
+/package.json @morpho-org/sdk-engineers
+/packages/*/package.json @morpho-org/sdk-engineers
```
0xbulma, line 5: **[LOW]** `/packages/*/package.json` does not cross slashes in CODEOWNERS, so a future nested workspace package (e.g. `packages/foo/sub/package.json`) would not be owned by `@morpho-org/sdk-engineers` and could merge without their review. Currently flat — verified — so coverage is complete today.

_Suggestion: Use `/packages/**/package.json` (CODEOWNERS does support `**` for slash recursion), or document the flat-layout assumption._

.github/dependabot.yml
```diff
@@ -4,4 +4,10 @@ updates:
     directory: /
     schedule:
       interval: weekly
-    target-branch: dev
+    target-branch: main
```
0xbulma, line 7: **[LOW]** Target branch flipped from `dev` to `main`. Verified `dev` does not exist on origin (only `main` and `next`), so the change is correct. However, `next` (the prerelease branch) is now uncovered by dependabot.

_Suggestion: Add a third entry with `target-branch: next` if prereleases should receive automated dependency updates, or document the merge-from-main flow._

.github/workflows/publish.yml
```diff
@@ -23,8 +23,9 @@ jobs:
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
-          ref: ${{ github.ref_name }}
+          ref: ${{ github.sha }}
```
0xbulma, line 26: **[LOW]** `ref: ${{ github.sha }}` is the security fix from `github.ref_name`, but lacks an inline comment explaining the rationale. Future maintainers may revert this as 'redundant'.

_Suggestion: One-line comment such as `# Pin to the triggering commit so a mid-run branch update cannot inject code into the publish job.`_

.github/workflows/publish.yml
```diff
@@ -75,31 +76,47 @@ jobs:
...
           publish_tag="latest"
           if [ "${GITHUB_REF_NAME}" = "next" ]; then
             publish_tag="next"
           fi
 
           before_tags="$(mktemp)"
           after_tags="$(mktemp)"
-          new_tags="$(mktemp)"
-          trap 'rm -f "$before_tags" "$after_tags" "$new_tags"' EXIT
+          new_tags="${RUNNER_TEMP}/new-package-tags.txt"
```
0xbulma, line 92: **[LOW]** `new_tags="${RUNNER_TEMP}/new-package-tags.txt"` is intentionally outside the `EXIT` trap so it survives across steps (consumed by `Push package tags` at L106). This is correct but isn't documented.

_Suggestion: One-line comment explaining the cross-step handoff, plus `rm -f "$new_tags"` at the start of this step for hygiene against re-run edge cases (`RUNNER_TEMP` can survive across job re-runs in some failure modes)._

.github/workflows/publish.yml
```diff
@@ -75,31 +76,47 @@ jobs:
...
-            refs=()
+            echo "has_tags=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_tags=false" >> "$GITHUB_OUTPUT"
+          fi
 
-            for tag in "${tags[@]}"; do
-              refs+=("refs/tags/${tag}")
-            done
+      - name: Push package tags
```
0xbulma, line 106: **[MEDIUM]** Splitting tag detection (`Publish packages` at L78) from tag push (`Push package tags` here at L106) widens the window in which npm artifacts are published without origin tags. If the runner is cancelled, evicted, or the push step fails after `pnpm release` succeeds, tags exist only locally and are reaped with the runner — leaving a partial release on npm.

_Suggestion: Push tags inside the same step as `pnpm release` (e.g. `git push --atomic --follow-tags origin HEAD` immediately after publish returns 0), or accept the failure mode and document the manual reconciliation procedure in CONTRIBUTING.md._

.github/workflows/publish.yml
```diff
@@ -75,31 +76,47 @@ jobs:
...
-            git push --atomic origin "${refs[@]}"
-          fi
+          mapfile -t tags < "${RUNNER_TEMP}/new-package-tags.txt"
+          refs=()
+
+          for tag in "${tags[@]}"; do
+            refs+=("refs/tags/${tag}")
+          done
+
+          git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```
0xbulma, line 121: **[LOW]** `git remote set-url origin "https://x-access-token:${GH_TOKEN}@..."` writes the workflow `GITHUB_TOKEN` into `.git/config` on the runner. Safe today (ephemeral GitHub-hosted runner; no `.git` directory is cached by `actions/cache`), but the token persists in the workspace until job teardown.

_Suggestion: Use `git -c http.extraheader='AUTHORIZATION: bearer ${GH_TOKEN}' push ...` instead, mirroring what `actions/checkout` does internally — avoids persisting the credential on disk._

.github/workflows/publish.yml
```diff
@@ -75,31 +76,47 @@ jobs:
...
+          git push --atomic origin "${refs[@]}"
```
0xbulma, line 122: **[LOW]** `git push --atomic origin "${refs[@]}"` fails entirely if any single tag already exists on origin (e.g. from a previous partially-succeeded run). Combined with the npm-published-without-tags scenario flagged at L106, recovery requires manual reconciliation and the workflow stays red.

_Suggestion: Drop `--atomic` so unrelated tag successes still land, or pre-filter against `git ls-remote --tags origin` before constructing the ref list._

.github/workflows/version-pr.yml
```diff
@@ -88,13 +93,72 @@ jobs:
...
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
+          set -euo pipefail
+
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
-          git add .
+
+          is_allowed_version_path() {
```
0xbulma, line 106: **[LOW]** `is_allowed_version_path()` is the security boundary that prevents the bot identity from committing arbitrary files during versioning, but it has no header comment explaining its intent.

_Suggestion: One- or two-line comment above the function, e.g. `# Allowlist of paths that 'pnpm run version' may modify; anything outside this set aborts the release as a sign of misconfigured tooling or supply-chain anomaly.`_

.github/workflows/version-pr.yml
```diff
@@ -88,13 +93,72 @@ jobs:
...
+            case "$1" in
+              packages/*/package.json|packages/*/CHANGELOG.md|.changeset/*.md|.changeset/pre.json)
```
0xbulma: **[MEDIUM]** Two issues with this allowlist:

1. **`packages/*/package.json` does not cross slashes** in bash `case` patterns. The repo's 15 workspaces are flat today (verified), but a future nested package (`packages/foo/sub/package.json`) would silently fail the version commit. Future-proof by replacing the `case` with a regex: `[[ "$1" =~ ^(packages/[^/]+/(package\.json|CHANGELOG\.md)|\.changeset/[^/]+\.md|\.changeset/pre\.json)$ ]]`, or document the flat-layout assumption.

2. **`packages/*/CHANGELOG.md` is dead.** `.changeset/config.json` sets `"changelog": false` and `CONTRIBUTING.md` L99 documents that no per-package CHANGELOGs are kept in source. Drop the entry, or add a comment reserving it for a future opt-in.

Also worth considering: a smoke-test job that runs `pnpm run version` against a synthetic changeset and asserts the produced file set ⊆ allowlist. Otherwise, future config drift only surfaces as a runtime `::error::` during the next release.

.github/workflows/version-pr.yml
```diff
@@ -88,13 +93,72 @@ jobs:
...
+            if is_allowed_version_path "$path"; then
+              allowed_changed+=("$path")
+            fi
+          done < <(git ls-files --others --exclude-standard -z)
+
+          if [ "${#allowed_changed[@]}" -gt 0 ]; then
+            git add -- "${allowed_changed[@]}"
+          fi
+
+          staged_violations=()
```
0xbulma, line 129: **[LOW]** Triple validation is logically belt + suspenders + spare suspenders. Step A (L117–132) only stages allowlisted paths; this Step B then asserts no staged path is unauthorized — by construction unreachable. Step C (L147–161) re-walks the worktree.

_Suggestion: Pick one model. Either drop Step B (this block), or simplify Step A to `git add -A` and rely solely on Step B + C as the gate. Current code is hard to follow without all three being aware of each other._

.github/workflows/version-pr.yml
```diff
@@ -103,15 +167,28 @@ jobs:
           fi
 
           git commit -m "chore: version packages"
-          git push --force origin "HEAD:refs/heads/${RELEASE_BRANCH}"
+
+          remote_release_ref="refs/heads/${RELEASE_BRANCH}"
+          local_release_ref="refs/remotes/origin/${RELEASE_BRANCH}"
+          if git ls-remote --exit-code --heads origin "${RELEASE_BRANCH}" >/dev/null 2>&1; then
```
0xbulma, line 168: **[MEDIUM]** TOCTOU between `git ls-remote --exit-code --heads origin` here and `git fetch ...` on the next line. If the release branch is created/deleted/rewritten in the gap, the workflow takes the wrong branch (force-with-lease vs. plain push). Mitigated in practice because `concurrency: cancel-in-progress: false` keeps this workflow the only writer, but not robust to manual branch edits or App-token compromise.

_Suggestion: Always attempt `git fetch origin '+refs/heads/${RELEASE_BRANCH}:refs/remotes/origin/${RELEASE_BRANCH}' || true`, then branch on local existence — that removes the race entirely._

.npmrc
```diff
@@ -3,3 +3,4 @@ link-workspace-packages=deep
 prefer-workspace-packages=true
 
 ignore-workspace-cycles=true
+minimum-release-age=4320
```
0xbulma, line 6: **[LOW]** `minimum-release-age=4320` is a magic number. The unit is **minutes** (= 72 hours / 3 days), and the directive requires pnpm ≥ 10.4 (repo on 10.27, fine).

_Suggestion: Add a one-line comment above this line explaining the unit and the rationale (supply-chain mitigation: refuse to install package versions newer than 72h to dodge fresh compromises)._

0xbulma (commented):

**Reviewed commit:** `a1451919`

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 6 |
| Low | 14 |

### Bottom line

Solid hardening pass — SHA-pinned actions, least-privilege `permissions:` per job, `persist-credentials: false`, `ref: github.sha`, secret enumeration replacing `secrets: inherit`, allowlist on the version commit, force-with-lease on the release branch, trusted publishing with `id-token: write`, and `minimum-release-age` in `.npmrc`. See inline comments for actionable nits and two operational risks worth addressing before merge:

1. **Tag/publish atomicity** (publish.yml L106 + L122) — splitting tag detection from tag push, plus `--atomic`, can leave npm-published-without-tags after partial failure.
2. **TOCTOU on the release branch** (version-pr.yml L173) — `ls-remote` then `fetch` is racy in the rare case of concurrent writers.

### Additional context-only findings (not anchored inline)

- **`publish.yml` L39 (pre-existing) — stale `Node 22.22.2` reference.** `.nvmrc` only pins major `22`, so `actions/setup-node` floats with whatever 22.x ships. The named patch will keep drifting. Consider rewording to `Node 22's bundled npm 10.x` and adding a tracking link/issue so the `npm@11.11.0` pin can be removed once upstream ships npm ≥ 11.5.1 by default.
- **`publish.yml` L42 (pre-existing) — `npm@11.11.0` + trusted publishing not validated end-to-end.** Prior churn (b22f32f5, 2aa6b4ff, 0a3b8fd8) but no record of a successful publish run with `id-token: write` against the npm registry. The first real release after merge becomes the validation event. Consider a `--dry-run --provenance` against a scratch package on a release-trial branch.
- **`version-pr.yml` L20 — `next` branch path unverified.** `origin/next` exists but is severely stale (last activity ~PR #447; main is 367 commits ahead, zero commits unique to `next`). `pnpm changeset pre enter next` and the `.changeset/pre.json` handling have not been exercised on the hardened workflow. Either delete `origin/next` (making the path dead until needed) or refresh and dry-run a prerelease cycle.
- **`publish.yml` L46 / `version-pr.yml` L44 — `Detect pending changesets` is byte-identical across both files.** Two copies of release-gating logic invite drift. Extract to a composite action (`.github/actions/detect-pending-changesets`).
- **`publish.yml` L72 / `version-pr.yml` L73 — `Refuse prerelease mode on main` duplicated verbatim.** Same composite-action recommendation.
- **`publish.yml` L67 — `Skip until version PR is merged` step name is misleading.** It only echoes a message; the actual gate is the `if:` on `Publish packages` at L80. Consider renaming to `Note: pending changesets, publish deferred` and emitting `::warning::` so it surfaces in the check summary.
- **`publish.yml` L48 / `version-pr.yml` L46 — `shell: bash` and `set -euo pipefail` repeated on every script step.** Move to `defaults.run.shell: bash -euo pipefail {0}` at job level for parity with the hardened steps.
- **`version-pr.yml` L46 / `publish.yml` L46 — `Detect pending changesets` lacks an explicit `set -euo pipefail`.** Relies on GHA defaults (no `-u`).
- **`version-pr.yml` L52 / `publish.yml` L54 — README.md skip is case-sensitive.** Stray `README.MD` would be treated as a real changeset on Linux. `shopt -s nocasematch` or `${name,,}` fixes it.
- **`version-pr.yml` L104 — `git remote set-url` writes the GitHub App token (broader trust scope than `GITHUB_TOKEN`) into `.git/config`.** Same recommendation as publish.yml L121 inline comment: prefer `git -c http.extraheader='AUTHORIZATION: bearer ${GH_TOKEN}' push ...`. Pre-existing line, not changed by this PR.
- **`publish.yml` L16 — job-level `if: github.ref_name == 'main' || github.ref_name == 'next'` duplicates the caller-side gate at `push.yml` L54.** Defense-in-depth or accidental? Document the dual gate as intentional, or remove one.
- **`build.yml` / `lint.yml` / `test.yml` / `publish.yml` / `version-pr.yml` — checkout + `pnpm/action-setup` + `setup-node` + `pnpm install --frozen-lockfile` is repeated verbatim 5×.** Composite action for follow-up; tradeoff: less per-workflow visibility of action SHAs.
- **`publish.yml` L110 — tag push uses `GITHUB_TOKEN` (`contents: write`).** If repo enables Tag protection rules in Settings, GITHUB_TOKEN tag creation can be blocked. Either carve out the `github-actions` bot or use the same `VERSION_APP_*` token as version-pr.yml.

_This is an automated parallel review using the same agents as `/ben-local-review`._

.github/CODEOWNERS
```diff
@@ -1,2 +1,9 @@
-/.github/CODEOWNERS @morpho-org/security
-/.github/workflows/ @morpho-org/security
+/.github/ @morpho-org/security
+/.changeset/ @morpho-org/security
```
Foulks-Plb: **[Important]** Owning the entire `/.changeset/` directory by `@morpho-org/security` forces every PR adding a changeset markdown file (per CLAUDE.md §2.10, that's effectively every user-visible PR) to wait on a security review.

Suggestion: scope security ownership to release config only (e.g. `/.changeset/config.json` and `/.changeset/pre.json`) and leave per-PR changeset `*.md` files unowned so any engineer can merge them. The protection you actually want — guarding the *publication* path — is already covered by ownership of `/.github/workflows/` plus the version-PR allowlist.

.github/CODEOWNERS
```diff
@@ -1,2 +1,9 @@
-/.github/CODEOWNERS @morpho-org/security
-/.github/workflows/ @morpho-org/security
+/.github/ @morpho-org/security
```
Foulks-Plb, line 1: **[Minor]** `/.github/` (with trailing slash) recursively covers everything under `.github`, including issue templates and PR templates that aren't security-sensitive. The previous two-line form explicitly listed CODEOWNERS and workflows; this form silently widens the scope.

Suggestion: confirm intentional. If non-security DX files (issue/PR templates) should remain unowned, list `/.github/workflows/`, `/.github/CODEOWNERS`, and `/.github/dependabot.yml` explicitly instead of the whole directory.

.github/dependabot.yml
```diff
@@ -4,4 +4,10 @@ updates:
...
+
+  - package-ecosystem: github-actions
```
Foulks-Plb, line 9: **[Minor]** Adding the `github-actions` ecosystem with `interval: weekly` will produce one PR per action per week. Each PR rewrites pinned SHAs and runs the full lint/build/test/version-pr matrix.

Suggestion: add a `groups:` entry so Dependabot batches all action bumps into a single weekly PR — reduces review and CI load while keeping the security benefit.

.github/workflows/publish.yml
```diff
@@ -75,31 +76,47 @@ jobs:
...
+          new_tags="${RUNNER_TEMP}/new-package-tags.txt"
+          trap 'rm -f "$before_tags" "$after_tags"' EXIT
 
-          git tag --list | sort > "$before_tags"
+          git tag --list | LC_ALL=C sort > "$before_tags"
           pnpm release --tag "$publish_tag"
-          git tag --list | sort > "$after_tags"
+          git tag --list | LC_ALL=C sort > "$after_tags"
 
           comm -13 "$before_tags" "$after_tags" > "$new_tags"
```
Foulks-Plb: **[Medium]** Inputs are sorted with `LC_ALL=C sort` (good), but `comm` itself runs under the ambient `LC_COLLATE` (UTF-8 on `ubuntu-latest`). When `comm`'s collation disagrees with the input order, it silently emits wrong differences — which here would mean publishing the wrong tag set (or none).

Suggestion: pin the comparison to the same locale as the sort:
```bash
LC_ALL=C comm -13 "$before_tags" "$after_tags" > "$new_tags"
```

.github/workflows/publish.yml
```diff
@@ -75,31 +76,47 @@ jobs:
...
+          fi
 
-            for tag in "${tags[@]}"; do
-              refs+=("refs/tags/${tag}")
-            done
+      - name: Push package tags
+        if: steps.publish-packages.outputs.has_tags == 'true'
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
```
Foulks-Plb, line 110: **[Medium]** This step pushes `refs/tags/*` using the default `${{ github.token }}`. If a repository ruleset or branch/tag protection blocks `GITHUB_TOKEN` from creating tags (a common hardening default), the publish will fail here even though the previous step succeeded — leaving npm and git out of sync.

Suggestion: either confirm the ruleset allows `GITHUB_TOKEN` to create `refs/tags/*`, or use the same GitHub App token already wired through `version-pr.yml` (`VERSION_APP_ID` / `VERSION_APP_PRIVATE_KEY`) so all release-bot writes go under one allowlisted identity.

.github/workflows/publish.yml
```diff
@@ -75,31 +76,47 @@ jobs:
...
           if [ "${GITHUB_REF_NAME}" = "next" ]; then
             publish_tag="next"
           fi
 
           before_tags="$(mktemp)"
           after_tags="$(mktemp)"
-          new_tags="$(mktemp)"
-          trap 'rm -f "$before_tags" "$after_tags" "$new_tags"' EXIT
+          new_tags="${RUNNER_TEMP}/new-package-tags.txt"
+          trap 'rm -f "$before_tags" "$after_tags"' EXIT
```
Foulks-Plb, line 93: **[Minor]** The trap removes `before_tags` and `after_tags` but not the `new_tags` file at `${RUNNER_TEMP}/new-package-tags.txt`. `RUNNER_TEMP` is cleaned at end-of-job so this isn't a leak, but the inconsistency invites future bugs (someone moves the consuming step out of this job and the file leaks until the runner is recycled).

Suggestion: either add a comment that `new_tags` is intentionally retained for the next step in the same job, or move `new_tags` to a job-scoped artifact / step output that's explicit about lifetime.

.github/workflows/publish.yml
Foulks-Plb, line 121: **[Minor]** `git remote set-url` persists the token inside `.git/config`. GitHub Actions masks `${{ github.token }}` in logs, but any future step in this job that reads `git config --get remote.origin.url` (or runs `git remote -v`) would print the token; it also persists for the rest of the job.

Suggestion: avoid persisting credentials by passing them per-invocation, e.g.:
```bash
git -c http.extraheader="Authorization: Bearer ${GH_TOKEN}" push --atomic origin "${refs[@]}"
```
Same pattern applies in `version-pr.yml`.

.github/workflows/test.yml
```diff
@@ -5,13 +5,22 @@ on:
     secrets:
       MAINNET_RPC_URL:
         required: true
+      BASE_RPC_URL:
+        required: true
```
Foulks-Plb, line 9: **[Important]** Marking `BASE_RPC_URL` and `ARBITRUM_RPC_URL` as `required: true` will hard-fail the entire push pipeline (`lint`/`build`/`test`/`version-pr`/`publish` all transitively depend on `test`) the first time this lands if either secret isn't configured. Previously `secrets: inherit` silently passed empty values, so missing secrets only manifested as failing tests, not aborted workflow startup.

Suggestion: verify both secrets exist at repo level before merge, or downgrade to `required: false` and let the tests that consume them fail individually with clearer locality.

.github/workflows/version-pr.yml
Foulks-Plb: **[Medium]** The allowlist permits only `packages/*/package.json`, `packages/*/CHANGELOG.md`, `.changeset/*.md`, `.changeset/pre.json`. Today this is fine: `.changeset/config.json` has `commit: false` and `updateInternalDependencies: patch` doesn't regenerate the lockfile (changesets doesn't run `pnpm install`).

But if anyone adds a `pnpm install --lockfile-only` step here in the future (or changesets ever bumps something that requires lockfile resolution), the post-version sanity check at lines 157-161 will hard-fail every release with a confusing error. Either add `pnpm-lock.yaml` proactively or add a comment naming the assumption ("lockfile is intentionally excluded; do not add `pnpm install` here without widening the allowlist").

.github/workflows/version-pr.yml
```diff
@@ -88,13 +93,72 @@ jobs:
...
         if: steps.pending-changesets.outputs.has_changesets == 'true'
+        shell: bash
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
+          set -euo pipefail
+
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```
Foulks-Plb, line 104: **[Minor]** Same token-in-remote-URL pattern as `publish.yml`: the app token persists in `.git/config` for the rest of the job (subsequent "Open or update release PR" inherits it via `gh`, which is fine; but any added step that prints the remote URL would surface the token).

Suggestion: prefer `git -c http.extraheader="Authorization: Bearer ${GH_TOKEN}" push ...` and unset/leave the remote URL untouched.

Foulks-Plb (commented):

**Reviewed commit:** `a1451919`

This PR is well-targeted CI hardening: tight per-job permissions, `persist-credentials: false` on every checkout, pinned `${{ github.sha }}` for release-critical refs, `secrets: inherit` replaced with explicit per-secret passing, an allowlist that fails-loud if `pnpm version` writes outside the release surface, and `--force-with-lease` on the release-branch push. The shell hardening (`set -euo pipefail`, null-delimited file lists, `LC_ALL=C sort`, `// empty` jq fallback) is solid.

The most significant findings worth resolving before merge:
- **CODEOWNERS gates every changeset on security review** — likely unintended scope.
- **`test.yml` marks two new secrets as required** — pipeline hard-fails if they aren't already configured.
- **`comm` runs under ambient locale** while inputs are `LC_ALL=C sort`-ed — minor collation hazard.
- **Tag push uses default `GITHUB_TOKEN`** — confirm branch/tag protection allows it; the version-PR step already uses a dedicated GitHub App token, consistency would be better.
- **PR description acknowledges trusted-publisher / npm env gate are deferred** — without them, the publish step has no npm credentials and will fail at runtime; that's pre-existing on `main`, but worth tracking with the PR.

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 2 |
| 🔵 Medium | 3 |
| 🔵 Minor | 9 |

.github/workflows/version-pr.yml
```diff
@@ -88,13 +93,72 @@ jobs:
...
+          fi
+
+          staged_violations=()
+          while IFS= read -r -d '' path; do
+            if ! is_allowed_version_path "$path"; then
+              staged_violations+=("$path")
+            fi
+          done < <(git diff --cached --name-only -z)
+
+          if [ "${#staged_violations[@]}" -gt 0 ]; then
```
Foulks-Plb, line 136: **[Minor]** `staged_violations` is effectively dead code: the only `git add` happens at line 131 and only adds files that already passed `is_allowed_version_path`. With `actions/checkout` providing a clean tree, nothing else is staged.

Suggestion: drop the block, or keep it and add a one-line comment marking it as a defense-in-depth assertion (so the next reader doesn't conclude it's load-bearing and replicate the pattern). The unstaged/untracked check at lines 147-161 is the actual guard.

.github/workflows/version-pr.yml
```diff
@@ -88,13 +93,72 @@ jobs:
...
+          done < <(git diff --name-only -z)
+
+          untracked=()
+          while IFS= read -r -d '' path; do
+            untracked+=("$path")
+          done < <(git ls-files --others --exclude-standard -z)
+
+          if [ "${#unstaged[@]}" -gt 0 ] || [ "${#untracked[@]}" -gt 0 ]; then
+            echo "::error::Versioning produced files outside the release allowlist:"
+            printf '  %s\n' "${unstaged[@]}" "${untracked[@]}"
```
Foulks-Plb, line 154: **[Minor]** Under `set -u`, expanding two arrays where either may be empty is safe in Bash 5 (which `ubuntu-latest` ships), but breaks in Bash 4. Not a current bug, just a sharp edge.

Suggestion (optional): use the empty-safe expansion to harden against future runner changes:
```bash
printf '  %s\n' "${unstaged[@]+"${unstaged[@]}"}" "${untracked[@]+"${untracked[@]}"}"
```

.github/workflows/version-pr.yml
Foulks-Plb, line 168: **[Minor]** `git ls-remote --exit-code` errors are silenced via `>/dev/null 2>&1`. A transient network failure is treated as "branch does not exist" and falls through to a non-force `git push`, which fails fast if the branch actually exists remotely (so behavior is safe). Worth a one-line comment to document that contract so the next maintainer doesn't "fix" the silencing into a `|| true`.

.npmrc
Foulks-Plb, line 6: **[Medium]** `minimum-release-age=4320` (minutes; 72 h) is a strong supply-chain mitigation, but it cuts both ways:

1. `pnpm install --frozen-lockfile` will refuse to resolve any version published within the last 72 h. If the lockfile already references a fresh version, install fails locally and in CI. Sanity-check `pnpm-lock.yaml` doesn't currently pin anything published in the last 3 days.
2. Emergency upgrades (e.g. `viem` security patch — the only `morpho-sdk` peer dep) need a documented override. Add a CONTRIBUTING.md note covering `--config.minimum-release-age=0` for one-off bumps, or a separate `.npmrc.emergency`.
3. The directive is silently ignored by pnpm < 10.16. The repo pins `pnpm@10.27.0` so this is fine today, but worth a note on the `engines.pnpm` field or the corepack pin to prevent regressions.

.github/workflows/version-pr.yml
```diff
@@ -88,13 +93,72 @@ jobs:
...
+
+          is_allowed_version_path() {
+            case "$1" in
+              packages/*/package.json|packages/*/CHANGELOG.md|.changeset/*.md|.changeset/pre.json)
+                return 0
+                ;;
+              *)
+                return 1
+                ;;
+            esac
```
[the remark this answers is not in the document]
Foulks-Plb: Fixed in 71840131 — switched to anchored regex with `[^/]+` so each component is bounded to one path level.

.github/CODEOWNERS
Foulks-Plb: Skipping — workspace layout is flat today; not future-proofing speculatively. Will revisit if nested workspaces are introduced.

.github/workflows/publish.yml
Foulks-Plb: Skipping — pure doc nit; the security rationale belongs in the PR description / CONTRIBUTING, not inline.

.github/workflows/publish.yml
Foulks-Plb: Skipping — comment-only; the cross-step handoff is visible from the variable name and the consuming step.

.github/workflows/publish.yml
Foulks-Plb: Skipping for this PR — real concern but invasive (changesets owns the publish step; injecting `git push --tags` into it requires a refactor). Window is narrow and recovery is manual today; tracked separately.

.github/workflows/publish.yml
Foulks-Plb: Skipping — runner is ephemeral on GitHub-hosted Ubuntu; `.git/config` is wiped at job teardown and no `actions/cache` step caches `.git`. Not worth the rewrite.

.github/workflows/publish.yml
Foulks-Plb: Skipping — partial-success recovery is rare and easier to handle manually than to teach the workflow to filter pre-existing tags. `--atomic` is intentional to avoid half-pushed tag sets.

.github/workflows/version-pr.yml
Foulks-Plb: Skipping — comment-only. The function name plus surrounding step name (`Push release branch`) already reads as the gate.

.github/workflows/version-pr.yml
Foulks-Plb: Both addressed in 71840131: switched to anchored regex (`[^/]+`) and dropped the dead `packages/*/CHANGELOG.md` entry (changesets has `changelog: false`).

.github/workflows/version-pr.yml
Foulks-Plb: Skipping — the three layers are defense-in-depth on a security boundary. Step A is the additive path, Step B is the unauthorized-stage assertion, Step C catches anything pnpm wrote outside the staged set. Cheap to keep; expensive to reconstruct after a regression.

.github/workflows/version-pr.yml
Foulks-Plb: Skipping — mitigated by `concurrency: cancel-in-progress: false` (this workflow is the sole writer). Manual branch edits during a release are explicitly out-of-band and would surface as a clean failure, not silent corruption.

.npmrc
Foulks-Plb: Skipping — comment-only. Default value lives in `.npmrc` documentation; if anyone touches the line they'll search the directive name.

.github/CODEOWNERS
Foulks-Plb: Fixed in 71840131 — scoped to `/.changeset/config.json` and `/.changeset/pre.json` so per-PR `*.md` changesets don't gate on security review.

.github/CODEOWNERS
Foulks-Plb: Skipping — broader scope is intentional. `/.github/` content is small (workflows, CODEOWNERS, dependabot, issue/PR templates) and security review on template changes is low-cost.

.github/dependabot.yml
Foulks-Plb: Skipping for now — QoL improvement, not urgent. Will revisit after we observe actual PR volume from the new `github-actions` ecosystem entry.

.github/workflows/publish.yml
Foulks-Plb: Fixed in 71840131 — pinned `LC_ALL=C` on `comm` to match the locale used for sorting the inputs.

.github/workflows/publish.yml
Foulks-Plb: Skipping for this PR — no tag protection ruleset is configured today, so `GITHUB_TOKEN` works. If/when a tag rule is added, this will fail loudly and we can switch to the `VERSION_APP_*` token at that point.

.github/workflows/publish.yml
Foulks-Plb: Skipping — comment-only. Same scope as the `0xbulma` thread above.

.github/workflows/publish.yml
Foulks-Plb: Skipping — ephemeral runner; same rationale as the publish.yml thread above.

.github/workflows/push.yml
```diff
@@ -13,34 +13,54 @@ concurrency:
...
   build:
+    permissions:
+      contents: read
+
     uses: ./.github/workflows/build.yml
 
   test:
+    permissions:
+      contents: read
+
```
Foulks-Plb, line 30: Skipping — secrets configuration is a repo-settings check, not a code change. Will verify in repo settings before merge.

.github/workflows/test.yml
Foulks-Plb: Skipping — same as above; this is a settings-side verification. Once confirmed configured, `required: true` is the correct stance (fail loud).

.github/workflows/version-pr.yml
Foulks-Plb: Skipping — speculative. `pnpm run version` doesn't run `pnpm install`; if anyone adds `--lockfile-only` later, the sanity check at L157–161 will fail loudly with a clear path-list message.

.github/workflows/version-pr.yml
Foulks-Plb: Skipping — ephemeral runner; same rationale as publish.yml.

.github/workflows/version-pr.yml
Foulks-Plb: Skipping — intentional defense-in-depth. See reply on the `0xbulma` triple-validation thread.

.github/workflows/version-pr.yml
Foulks-Plb: Skipping — `ubuntu-latest` ships Bash 5.x; not optimizing for a runner downgrade that GitHub doesn't offer.

.github/workflows/version-pr.yml
Foulks-Plb: Skipping — comment-only. Behavior on transient `ls-remote` failure is safe (falls through to non-force push, which itself fails loud).

.npmrc
Foulks-Plb: Skipping for this PR — supply-chain notes belong in CONTRIBUTING, not `.npmrc`. Will follow up with a CONTRIBUTING note covering the override pattern and pnpm version requirement.

.github/dependabot.yml
Foulks-Plb: Applied in 19617b54 — added npm + github-actions dependabot entries targeting `next`.

0xbulma (approved):

Foulks-Plb (approved):

.github/workflows/push.yml
```diff
@@ -13,34 +13,54 @@ concurrency:
...
     uses: ./.github/workflows/version-pr.yml
-    secrets: inherit
+    secrets:
+      VERSION_APP_ID: ${{ secrets.VERSION_APP_ID }}
+      VERSION_APP_PRIVATE_KEY: ${{ secrets.VERSION_APP_PRIVATE_KEY }}
 
   publish:
     if: github.ref_name == 'main' || github.ref_name == 'next'
 
+    permissions:
```
tarikbellamine, line 56: Medium: this publish job is still governed by the workflow-level `cancel-in-progress: true` above. On `main`/`next`, a later push can cancel an in-flight release after `pnpm release` has published packages but before tags are pushed, leaving npm ahead of git.

Consider disabling cancellation for release branches:

```yaml
cancel-in-progress: ${{ github.ref_name != 'main' && github.ref_name != 'next' }}
```

.github/workflows/publish.yml
```diff
@@ -75,31 +76,47 @@ jobs:
...
-            git push --atomic origin "${refs[@]}"
-          fi
+          mapfile -t tags < "${RUNNER_TEMP}/new-package-tags.txt"
+          refs=()
+
+          for tag in "${tags[@]}"; do
+            refs+=("refs/tags/${tag}")
+          done
+
+          git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```
tarikbellamine, line 121: Medium: this step runs after repo-controlled install/publish lifecycle code has already run, then exposes `GH_TOKEN` and writes it into `.git/config` before `git push`. If anything earlier creates `.git/hooks/pre-push`, Git will execute that hook here with the token available.

Prefer disabling hooks for the push and using an ephemeral auth header instead of mutating `origin`, for example:

```bash
git -c core.hooksPath=/dev/null \
  -c "http.https://github.com/.extraheader=AUTHORIZATION: bearer ${GH_TOKEN}" \
  push --atomic origin "${refs[@]}"
```

tarikbellamine (commented):
Severity-ranked re-review (P0/P1/P2). Inline comments cover findings tied to specific lines. The following are repo-settings or out-of-diff items, called out here instead:

**P0 — No branch protection on `main` or `next`.** `main` → 404 on `/protection`; `next` → `protected:true, enabled:false`. Until this lands, every CODEOWNERS line in this PR is advisory and direct push to `main` ships to npm. This is the prerequisite for the rest of the PR to mean anything. Repository ruleset enforcing required reviews + CODEOWNERS reviews + no force-push + no deletion + required status checks (lint/build/test) on both branches.

**P2 — `default_workflow_permissions: write` at repo level.** Flip to `read` so future workflows fail closed. Repo settings → Actions → Workflow permissions.

**P2 — `GITHUB_TOKEN` can approve PRs (`can_approve_pull_request_reviews: true`).** Privilege-escalation vector if combined with auto-merge. Repo settings → Actions → 'Allow GitHub Actions to create and approve pull requests' → off.

**P2 — Audit `pnpm.overrides` in root `package.json` (not modified by this PR).** Several pins look unfamiliar (`lodash: 4.18.1`, `axios: 1.15.0`, `js-yaml: 4.1.1`, `flatted: 3.4.2`). Supply-chain attacks often hide in overrides because they rarely get re-reviewed. One-time check against the npm registry.

P3 items (cosmetic / hygiene) omitted from the inline pass; see prior conversation for the full list.

PR #589, opened 2026-05-04
fix(morpho-sdk): refuse partial shared-liquidity plans on impossible borrows

Foulks-Plb:
## Summary

- `computeReallocations` now throws `InsufficientSharedLiquidityError` when the selected vault withdrawals do not fully cover the target market's borrow shortfall, instead of returning a fee-bearing partial plan whose downstream `morphoBorrow` would still revert onchain.
- Mirrors the feasibility check that `populateBundle` already enforces in `@morpho-org/bundler-sdk-viem` (it re-simulates the populated sub-bundle and throws on an impossible borrow); the consumer-side mirror was missing this last guard.
- Partial coverage of the supply-target utilization heuristic remains acceptable as long as the absolute borrow shortfall (`newBorrow > newSupply`) is fully covered — so existing legitimate paths (e.g. `supplyTargetUtilization = 0`, low-target overrides) continue to work unchanged.

Addresses Cantina audit finding **MORP2-28** (informational) — [SDK-166](https://linear.app/morpho-labs/issue/SDK-166/morp2-28informational-consumer-computereallocations-can-return-fee).


packages/morpho-sdk/src/types/error.ts
```diff
@@ -326,6 +326,18 @@ export class MissingPublicAllocatorConfigError extends Error {
...
+  constructor(params: {
+    market: string;
+    shortfall: bigint;
+    available: bigint;
+  }) {
+    super(
+      `Shared liquidity is insufficient to cover the borrow on market ${params.market}: shortfall "${params.shortfall}", available "${params.available}". Reduce the borrow amount or wait for additional vault liquidity.`,
+    );
+  }
+}
```
0xbulma, line 392: **[MEDIUM]** Typed errors are public API and integrators pattern-match on them — but here `market`, `shortfall`, and `available` are only embedded in the message string, with no instance fields exposed. Tests and integrators cannot assert the populated payload without parsing the message, which [`CLAUDE.md` §5](https://github.com/morpho-org/sdks/blob/main/CLAUDE.md) explicitly forbids (`Errors asserted by class identity, not message string — messages can change without notice; classes are public API`).

Suggestion: promote the params to `public readonly` fields so callers can pattern-match on the payload:

```ts
export class InsufficientSharedLiquidityError extends Error {
  constructor(
    public readonly params: {
      readonly marketId: MarketId;
      readonly shortfall: bigint;
      readonly available: bigint;
    },
  ) {
    super(
      `Shared liquidity is insufficient to cover the borrow on market ${params.marketId}: shortfall "${params.shortfall}", available "${params.available}". Reduce the borrow amount or wait for additional vault liquidity.`,
    );
  }
}
```

While here: rename `market` → `marketId` and tighten the type to `MarketId` (the call site already passes a `MarketId`). The reallocation-error cluster (`UnsortedReallocationWithdrawalsError`, `MarketIdMismatchError`) already uses `marketId`; this aligns the new class with that group. Worth a one-line JSDoc on the class too — the surrounding file lacks JSDoc on every error class so this widens the gap rather than closing it; per [§9](https://github.com/morpho-org/sdks/blob/main/CLAUDE.md) (`on refactor, adopt the convention`), seeding a JSDoc here is a good first step.

packages/morpho-sdk/src/helpers/computeReallocations.test.ts
```diff
@@ -555,6 +558,68 @@ describe("computeReallocations", () => {
...
+      });
+
+      expect(() =>
+        computeReallocations({
+          reallocationData: data,
+          marketId: targetParams.id,
+          borrowAmount,
+          options: { enabled: true },
+        }),
+      ).toThrow(InsufficientSharedLiquidityError);
```
0xbulma: **[MEDIUM]** This test only asserts class identity (`toThrow(InsufficientSharedLiquidityError)`); the populated `shortfall`/`available` payload is not asserted, so a regression that throws the right class with wrong numbers would slip through. Once the error exposes readonly fields (see comment on `error.ts`), wrap in try/catch and assert:

```ts
try {
  computeReallocations({ ... });
  expect.unreachable();
} catch (e) {
  expect(e).toBeInstanceOf(InsufficientSharedLiquidityError);
  expect((e as InsufficientSharedLiquidityError).params.shortfall).toBe(200n * MathLib.WAD);
  expect((e as InsufficientSharedLiquidityError).params.available).toBe(150n * MathLib.WAD);
}
```

Also consider adding a tighter just-below-boundary case (e.g. friendly 50 WAD + aggressive 149 WAD vs 200 WAD shortfall) to lock the strict `<` comparison at `computeReallocations.ts:127`.

packages/morpho-sdk/src/helpers/computeReallocations.test.ts
```diff
@@ -555,6 +558,68 @@ describe("computeReallocations", () => {
       );
     });
 
+    test("error: InsufficientSharedLiquidityError when reallocations cannot cover the borrow shortfall", () => {
```
0xbulma, line 561: **[LOW]** This throw test exercises only the phase-1+phase-2 path (the configured `friendlyTargetMarket` of supply 850 / borrow 500 with borrowAmount 500 fires the phase-2 trigger `1000 > 850`). The phase-1-only path with `absoluteShortfall > 0n` is not covered.

It is mathematically unreachable under the `getMarketPublicReallocations` invariant (a non-fired phase 2 implies friendly already covers `absoluteShortfall`, so `totalReallocated >= absoluteShortfall`) — but that invariant lives outside this helper and a future refactor of the simulation-sdk side could silently regress it. Add either (a) a phase-1-only no-throw fork test where `absoluteShortfall > 0n` and the friendly reallocation just covers it, or (b) a comment in `computeReallocations.ts` near line 127 documenting the relationship `requiredAssets >= absoluteShortfall` that makes the guard sound.

packages/morpho-sdk/src/helpers/computeReallocations.test.ts
```diff
@@ -555,6 +558,68 @@ describe("computeReallocations", () => {
...
+        computeReallocations({
+          reallocationData: data,
+          marketId: targetParams.id,
+          borrowAmount,
+          options: { enabled: true },
+        }),
+      ).toThrow(InsufficientSharedLiquidityError);
+    });
+
+    test("should not throw when shortfall is zero even if supply target is unmet", () => {
```
0xbulma: **[LOW]** Test naming: per [`CLAUDE.md` §5](https://github.com/morpho-org/sdks/blob/main/CLAUDE.md), non-default non-error tests use the `behavior: <case>` prefix. Suggestion:

```ts
test("behavior: does not throw when absolute shortfall is zero even if supply target is unmet", () => { ... });
```

The rest of this file uses the older `should …` form pervasively — per §9 (`on refactor, adopt the convention`), the new tests landing here are the natural seed; consider aligning the `should throw MissingPublicAllocatorConfigError …` test below (line 623) to `error: MissingPublicAllocatorConfigError when vault config is missing` in the same pass.

0xbulma (commented):

**Reviewed commit:** `ee0d939`

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 1 |
| 🔵 Medium | 2 |
| 🔵 Minor | 2 |

### Overview

Fix is correct: `absoluteShortfall = newTotalBorrowAssets − newTotalSupplyAssets` is the right Morpho Blue feasibility threshold (`totalBorrowAssets + borrow ≤ totalSupplyAssets + reallocated`), and the new throw correctly mirrors the `populateSubBundle` re-simulation guard in `@morpho-org/bundler-sdk-viem`. Mathematical trace confirms phase-1-only paths cannot reach the throw under the simulation-sdk consistency invariant (a non-fired phase 2 implies friendly already covers `absoluteShortfall`); phase-2 paths reset `requiredAssets` to exactly `absoluteShortfall`, so the comparison is sound.

### Important — outside the inline diff

**🟡 Missing changeset.** This PR adds a new public exported error class (`InsufficientSharedLiquidityError`) and changes the throw behavior of `computeReallocations` — both user-visible. Per [`CLAUDE.md` §7](https://github.com/morpho-org/sdks/blob/main/CLAUDE.md), every user-visible change ships with a changeset; only `.changeset/config.json` is currently present. Add a `.changeset/<name>.md` with a `@morpho-org/morpho-sdk` minor bump (additive throw on a previously partial-plan-returning path is observably a behavior change for external consumers calling `computeReallocations` directly).

See inline comments for the rest. _Automated parallel review._

packages/morpho-sdk/src/types/error.ts
Foulks-Plb: Addressed in 22f333df: promoted constructor params to `public readonly`, renamed `market` → `marketId` (typed as `MarketId`), and added a JSDoc on the class. Tests now assert on `err.params.{marketId,shortfall,available}`.

packages/morpho-sdk/src/helpers/computeReallocations.test.ts
Foulks-Plb: Done in 22f333df: switched to try/catch and assert on `err.params.shortfall` (200 WAD) and `err.params.available` (150 WAD). Added a tighter just-below-boundary case (149 WAD aggressive vs 200 WAD shortfall = 199 vs 200) that locks the strict `<` comparison.

packages/morpho-sdk/src/helpers/computeReallocations.test.ts
Foulks-Plb: Went with option (b): added a comment in `computeReallocations.ts` near the `absoluteShortfall` definition documenting the `requiredAssets >= absoluteShortfall` invariant and why both phases satisfy it (phase-1-only path is mathematically unreachable, phase-2 resets `requiredAssets` to exactly `absoluteShortfall`).

packages/morpho-sdk/src/helpers/computeReallocations.test.ts
Foulks-Plb: Renamed the new test to `behavior: does not throw when absolute shortfall is zero even if supply target is unmet` and aligned the existing `should throw MissingPublicAllocatorConfigError …` test to the `error: <ErrorClass>` form. The other older `should …` tests in this file were left untouched to keep the diff scoped — happy to do a separate sweep if you'd prefer.

Foulks-Plb:
Thanks for the parallel review @0xbulma — addressed in 22f333df. Summary:

**Accepted:**
- 🟡 Error class fields → promoted constructor params to `public readonly`, renamed `market` → `marketId` (typed as `MarketId`), added a JSDoc on the class.
- 🟡 Test only asserts class identity → switched to try/catch and assert on `err.params.{marketId,shortfall,available}`. Added a tighter just-below-boundary case (199 vs 200 WAD) that locks the strict `<` comparison.
- 🔵 Phase-1-only path soundness → went with option (b): added a comment in `computeReallocations.ts` documenting the `requiredAssets >= absoluteShortfall` invariant.
- 🔵 Test naming → renamed the new `should not throw …` test to `behavior: does not throw …` and aligned the existing `should throw MissingPublicAllocatorConfigError …` test to `error: …`. Left the other older `should …` tests in this file untouched to keep the diff scoped.

**Declined:**
- 🟡 Missing changeset → the repo no longer ships changesets per PR (see `f114c69a docs: drop empty changeset` and the `70cd6f66 chore: drop changeset` precedent on `main`). I dropped the changeset in d8bfda6e for the same reason. Happy to add one back if the convention has changed since.

Rubilmax (approved):

PR #596, opened 2026-05-04
test: maximize unit-test coverage of SDK packages

0xbulma:
Adds **TIB-2026-04-27** at `docs/tibs/TIB-2026-04-27-maximize-unit-test-coverage.md` and the full implementation across **10 in-scope packages** plus a follow-up phase covering MetaMorphoAction, MarketUtils, paraswap, flashbots, plus three rounds of review fixes (Codex P2 + Devin + two local-review rounds).

## What this TIB decides

The eight-then-ten in-scope SDK packages — `morpho-ts`, `blue-sdk`, `blue-sdk-viem`, `bundler-sdk-viem`, `liquidation-sdk-viem`, `liquidity-sdk-viem`, `morpho-test`, `test`, plus `morpho-sdk` and `evm-simulation` (added in Phase 9 addendum) — get pushed toward as-close-to-100% **unit-test** coverage as practical, via:

1. **Colocation** of new unit tests next to source (`src/Foo.test.ts` beside `src/Foo.ts`).
2. **viem transport-level mocking** with `viem.custom()` — required because source uses `viem/actions` named imports, so client-method spies are silently no-ops.
3. **`nock`** for HTTP and GraphQL.
4. **`*.integration.test.ts`** naming for fork-bound tests that stay.

A new `createMockClient` / `mockRead` / `expectReadCall` helper is added to `@morpho-org/test` (sub-export `./mock`). `coverage.exclude` is tightened to drop generated files (`abis.ts`, `api/types.ts`, `api/sdk.ts`) and barrel `index.ts` files from the denominator.

Out of scope: `simulation-sdk`, `simulation-sdk-wagmi`, `migration-sdk-viem`, `blue-sdk-wagmi`, `test-wagmi`. They keep their `test/` directories and current vitest project entries.

## Coverage (last full measurement on `d5ac4295`)

Captured via `pnpm test:coverage --project <each in-scope project> --coverage.reportOnFailure=true`, lcov merged. **Local fork-bound tests fail to public-RPC rate-limiting, so these numbers are a floor — CI numbers (with `MAINNET_RPC_URL`/`BASE_RPC_URL` secrets) are higher.**

**Combined (10 in-scope projects):** 945 / (1042 - 3 skipped) tests passing. Lines: **35.01%** (2086 / 5958), Branches **27.84%**, Functions **39.58%**, Statements **34.86%**.

| Package | Lines | Coverage | Notes |
| --- | --- | --- | --- |
| `morpho-ts` | 288/299 | **96.32%** | Phase 1 target met (≥95%) |
| `evm-simulation` | 326/344 | **94.77%** | Already at Phase target — no new tests required |
| `blue-sdk` | 644/894 | **72.04%** | Math libs + errors + token + holding + vault classes 100% |
| `blue-sdk-viem` | 265/395 | **67.09%** | augment, error, utils, MetaMorphoAction (332 LOC, was 0%) covered |
| `liquidity-sdk-viem` | 13/43 | 30.23% | Denominator shrank after `coverage.exclude` cleanup; loader still primarily fork-covered |
| `morpho-sdk` | 253/873 | 28.98% | Helpers (constants, encodeDeallocation, metadata) covered; entities + actions stay fork-covered |
| `liquidation-sdk-viem` | 69/350 | 19.71% | thresholds, helpers, sky, 1inch, flashbots covered; LiquidationEncoder remains the largest gap |
| `bundler-sdk-viem` | 95/833 | 11.40% | errors + ActionBundle covered; BundlerAction (2,346 LOC) is the largest deferred file |

Round-3 added a handful of inline snapshots (MetaMorphoAction privileged setters) and split `augment.test.ts` into a `test.each` (added `AccrualPosition.fetch`, `AccrualPosition.fetchPreLiquidation`, `AccrualVault.fetch`) — line coverage % is materially unchanged from `d5ac4295`.

Largest gaps still deferred (Phase 11+): `LiquidationEncoder.ts` (1,001 LOC), `BundlerAction.ts` (2,346 LOC), `morpho-sdk/src/entities/marketV1/marketV1.ts` (1,071 LOC), `morpho-sdk/src/actions/*` builders.

**Files moved from 0% → 100%** by this PR (selection): `morpho-ts/{utils,urls,format/array,format/string}.ts`, `blue-sdk/{holding/AssetBalances,token/ExchangeRateWrappedToken,vault/VaultUser,errors,preLiquidation}.ts`, the entire `blue-sdk/src/math/*` namespace, `blue-sdk-viem/MetaMorphoAction.ts` (332 LOC).

## Source bug fixed in passing

- `morpho-sdk/src/helpers/metadata.ts:42` — `metadata.origin.replace(/^(?!0x)/, "")` was a no-op (zero-width negative-lookahead replaced an empty match with empty), so a `0x`-prefixed origin was *not* normalized — it became `"0x0xcafe"`, failed `isHex`, and got silently dropped via `console.warn`. Fixed to `/^0x/` so `"0xcafe"` and `"cafe"` produce the same 4-byte appended origin. Behavioural change documented in `.changeset/morpho-sdk-metadata-origin-prefix.md`.

## Rollout

10 phases (1 foundation + 8 packages + 1 deferred-large-files extension), one commit per phase on this single PR. `morpho-ts` → `test` (`createMockClient`) → `blue-sdk` → `morpho-test` → `liquidity-sdk-viem` → `bundler-sdk-viem` → `blue-sdk-viem` → `liquidation-sdk-viem` → `morpho-sdk` (Phase 9 addendum) → cross-package gap-fill (Phase 10) → review fixes (round 1: Codex P2 + Devin) → review fixes (round 2: local review) → review fixes (round 3: second local-review pass) → `chore: merge main` (commit `c71d22af`).

## Reviewer attention

Three Open Questions in the TIB are flagged for discussion:

1. Are there in-scope source files that legitimately can't be unit-tested at all without a fork (pure RPC plumbing whose entire purpose is the round-trip)?
2. Does having `src/Foo.test.ts` (new unit) and `test/Foo.test.ts` (existing fork) for the same source file cause any review confusion?
3. (Now resolved.) AGENTS.md §2.6 / §5 carve-out for transport-level mocks via `createMockClient` — codified in this PR; §2.6 wording broadened in round 3 to explicitly enumerate fork-only paths (oracles, accruals, deployless aggregation, contract-revert behaviour).

## Test plan

- [x] **45 colocated test files**, **548+ colocated tests pass** locally on `pnpm test --project <each in-scope> --run packages/<each>/src` (incl. `packages/morpho-sdk/src/helpers` for the non-fork helpers).
- [x] **All 7 touched packages build cleanly** via `pnpm --filter @morpho-org/<each> build` (ESM + CJS).
- [x] **No test files reach `lib/`**: `find packages/<each>/lib -name "*.test.*"` is empty everywhere.
- [x] **Lint**: `pnpm lint` (= `biome check && lint:address`) passes — only pre-existing warnings on `codegen.ts` files and one info on `anvil.ts` from biome 2.4.14, all on `main`.
- [x] CI green on `0953e548` (round-3 commit): lint / build / test all pass.
- [x] Merged `main` (commit `c71d22af`) — pulled in pnpm v11 + viem 2.49.3 + vitest 4.1.6 + typescript 6.0.3 bumps and the AGENTS.md §2.10/§6/§7 wording refresh.


packages/test/src/mock.ts
```diff
@@ -0,0 +1,161 @@
...
+>(handle: MockClientHandle, options: MockReadOptions<abi, fn>): void {
+  const { request } = handle;
+  const target = options.address.toLowerCase();
+  // Compute the 4-byte selector once per mock by composing the function
+  // signature with viem's helpers. This avoids repeating expensive ABI
+  // walks on every RPC dispatch.
+  const fnAbi = options.abi.find(
+    (item): item is Extract<typeof item, { type: "function" }> =>
+      item.type === "function" && item.name === options.functionName,
+  );
```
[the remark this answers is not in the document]
0xbulma: > Disambiguate overloaded ABI functions in mockRead

Fixed in 47f2e261 — `mockRead` now computes selectors for **every** ABI entry whose name matches `options.functionName`, builds a set of those selectors, and matches the call's 4-byte selector against the full set in the dispatch. Reads against overloaded `view`/`pure` functions resolve correctly regardless of which overload `readContract` encoded. Added a colocated test (`packages/test/src/mock.test.ts`) covering two same-named overloads with different argument lists.

packages/blue-sdk-viem/src/fetch/MarketParams.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const result = await fetchMarketParams(freshId, handle.client, {
+      chainId: ChainId.EthMainnet,
+    });
+    expect(result).toBeInstanceOf(MarketParams);
+    expect(result.loanToken).toBe(params.loanToken);
+    expect(result.collateralToken).toBe(params.collateralToken);
+    expect(result.lltv).toBe(params.lltv);
+  });
+});
```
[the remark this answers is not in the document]
0xbulma: > Mocked viem client used on RPC code path in fetchMarketParams.test.ts (AGENTS.md §2.6/§5 violation)

Fixed in 47f2e261 — updated AGENTS.md §2.6 and §5 to carve out transport-level mocks for unit tests, per the convention TIB-2026-04-27 establishes:

- §2.6 now distinguishes between integration tests (must use Anvil forks) and pure unit tests (may use `createMockClient` from `@morpho-org/test/mock`, which installs a `vi.fn`-backed `custom()` transport — the same surface `viem/actions` reads from).
- §5 now states the same rule and explicitly forbids `vi.spyOn(client.readContract)` because actions don't dispatch through methods.

The TIB and the AGENTS.md update land in the same PR, so the rule and its proposed exception ship together.

packages/test/package.json
```diff
@@ -22,6 +22,7 @@
     ".": "./src/index.ts",
     "./vitest": "./src/vitest.ts",
     "./fixtures": "./src/fixtures.ts",
+    "./mock": "./src/mock.ts",
```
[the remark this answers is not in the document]
0xbulma: > Missing changeset for new ./mock sub-export in @morpho-org/test (AGENTS.md §2.10 violation)

Fixed in 47f2e261 — added `.changeset/test-mock-subexport.md` with a minor bump for `@morpho-org/test`, summarising the additive surface (`createMockClient`, `mockRead`, `expectReadCall`) and the overload-aware dispatch fix from the Codex thread.

packages/test/src/mock.ts
```diff
@@ -0,0 +1,161 @@
...
+): MockClientHandle<chain> {
+  const request = vi.fn<RpcHandler>(async ({ method, params }) => {
+    if (method === "eth_chainId") return `0x${chain.id.toString(16)}`;
+    throw new Error(
+      `[createMockClient] unhandled RPC ${method} ${JSON.stringify(params)}`,
+    );
+  });
+  const client = createClient({ chain, transport: custom({ request }) });
+  return { client, request, chain };
+}
```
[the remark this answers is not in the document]
0xbulma: > Exported functions in mock.ts missing required @param/@returns/@throws JSDoc tags (AGENTS.md §6 violation)

Fixed in 47f2e261 — added the required tags to all three exported functions in `packages/test/src/mock.ts`:

- `createMockClient`: `@param chain` (default `mainnet`), `@returns` `MockClientHandle` shape, `@throws` on unhandled RPC.
- `mockRead`: `@param handle`, `@param options` (with `address`/`abi`/`functionName`/`result` sub-params), `@throws` when `functionName` is not in the abi.
- `expectReadCall`: `@param handle`, `@param match` (with `address`/`abi`/`functionName`), `@returns` array of decoded call data.

Descriptions retained.

packages/test/src/mock.ts
```diff
@@ -0,0 +1,310 @@
...
+    throw new Error(
+      `[mockRead] function ${String(options.functionName)} not found in abi`,
+    );
+
+  const params = {
+    abi: options.abi,
+    functionName: options.functionName,
+    result: options.result,
+  } as EncodeFunctionResultParameters<abi, fn>;
+  const encoded = encodeFunctionResult(params);
```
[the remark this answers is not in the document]
0xbulma: > Encode overloaded mock results per selector — when an ABI has same-named overloads with different return types, this encodes `options.result` only once using the ambiguous `functionName`, then registers those same bytes for every overload selector.

Fixed in b0d12e19. `mockRead` now ABI-encodes the result **per overload**, using a single-item ABI containing only that overload, so each overload's selector gets bytes encoded against its declared output shape:

```ts
for (const item of fnAbiItems) {
  const selector = toFunctionSelector(toFunctionSignature(item)).toLowerCase();
  try {
    const encoded = encodeFunctionResult({
      abi: [item],
      functionName: options.functionName as string,
      result: options.result,
    });
    dispatch.set(`${target}|${selector}`, encoded);
    registered++;
  } catch {
    // Encoding failed for this overload — result doesn't match its return
    // shape. Skip; another overload may still accept it.
  }
}
if (registered === 0)
  throw new Error(`[mockRead] options.result does not match any return-type shape of overloads of ${String(options.functionName)} (${fnAbiItems.length} overload(s) tried)`);
```

Behaviour is unchanged when overloads share a return type (every selector still gets bytes); for different return types — e.g. `counter(uint256) returns (uint256)` vs `counter(address) returns (bool)` — only the matching overload's selector is now registered, and the caller gets a clear error if the result fits none. Added a test pinning the different-return-type case (encode `true` then `99n` on the two overloads, assert each `eth_call` resolves correctly).

packages/test/package.json
```diff
@@ -85,6 +89,11 @@
         "import": "./lib/esm/fixtures.js",
         "require": "./lib/cjs/fixtures.js"
       },
+      "./mock": {
+        "types": "./lib/esm/mock.d.ts",
+        "import": "./lib/esm/mock.js",
+        "require": "./lib/cjs/mock.js"
+      },
```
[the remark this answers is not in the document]
0xbulma: > Do not publish a broken CJS mock export — `require("@morpho-org/test/mock")` fails at import time because `lib/cjs/mock.js` imports vitest via `require("vitest")`, which Vitest rejects.

Fixed in b0d12e19. The `require` condition was dropped from `publishConfig.exports["./mock"]` and `src/mock.ts` was added to the CJS build's `exclude` list. The ESM entry (`./lib/esm/mock.js`) is unchanged. vitest is ESM-only so any legitimate consumer of this subpath already resolves to ESM; no source change is needed in consumer projects. Verified with a clean build: `lib/cjs/mock.*` is no longer emitted, `lib/esm/mock.*` is intact.

A patch changeset (`.changeset/test-mock-esm-only.md`) documents both this and the per-overload encoding fix above.

Rubilmax (dismissed):

packages/blue-sdk-viem/src/error.test.ts
```diff
@@ -0,0 +1,77 @@
+import { BaseError, ContractFunctionRevertedError } from "viem";
+import { describe, expect, test } from "vitest";
+import { isUnknownOfFactoryError } from "./error.js";
```
Rubilmax: **[MEDIUM]** Coverage gap: this file tests only `isUnknownOfFactoryError`, but `error.ts` also exports `getUnsupportedVaultV2Adapter` plus three error classes (`InvalidPermitDomainChainIdError`, `InvalidPermitDomainVerifyingContractError`, `UnsupportedPermitDomainExtensionsError`) — all untested. `error.ts` is named in TIB-2026-04-27 as a 0%-coverage target to fully cover.

_Suggestion: add tests for `getUnsupportedVaultV2Adapter` (non-`BaseError` -> null, matching `errorName` -> extracted adapter address, non-matching -> null) and for each error class (field preservation + message)._

packages/liquidation-sdk-viem/src/flashbots.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  // Note: two further behaviours were attempted but proved unreliable
+  // under nock + Node fetch:
+  //
+  //   - a 5xx-reply test for the `if (!response.ok) throw` branch — nock's
+  //     reply body is delivered as text/plain by default and the source's
+  //     `await response.json()` in the error block parses inconsistently
+  //     across Node fetch implementations.
+  //   - an auto-increment-id test asserting `id` strictly increases between
+  //     two calls — nock's interceptor consumption interacts unpredictably
```
Rubilmax: **[LOW]** `sendRawBundle`'s `!response.ok` error branch and `signBundle` are left uncovered. The inline comment documents this as a deliberate nock/ESM limitation deferred to fork tests — an acknowledged gap, not a defect.

_Suggestion (optional): `sendRawBundle`'s error path is pure enough to unit-test with `nock` `reply(500, body, { 'content-type': 'application/json' })`, which would close the largest behavior gap of this file._

packages/morpho-sdk/src/helpers/encodeDeallocation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    ).toThrow(NonPositiveAssetAmountError);
+  });
+
+  test("throws NonPositiveAssetAmountError on negative amount", () => {
+    expect(() =>
+      encodeForceDeallocateCall({ adapter: ADAPTER, amount: -1n }, ON_BEHALF),
+    ).toThrow(NonPositiveAssetAmountError);
+  });
+
+  test("error message references the adapter context", () => {
```
Rubilmax, line 76: **[LOW]** The test `"error message references the adapter context"` never asserts anything about the error message — it only checks `instanceof NonPositiveAssetAmountError`, which is already covered by the preceding tests, so it does not exercise its stated intent.

_Suggestion: assert on the message/fields the title claims (`expect((e as NonPositiveAssetAmountError).message).toContain(...)`), or rename/remove the test._

packages/morpho-sdk/src/helpers/metadata.ts
```diff
@@ -39,7 +46,9 @@ export function addTransactionMetadata(
   }
 
   try {
-    const origin = metadata.origin.replace(/^(?!0x)/, "");
+    // Strip a leading "0x" if present so callers can pass either
+    // "cafe" or "0xcafe" — both produce the same 4-byte appended origin.
+    const origin = metadata.origin.replace(/^0x/, "");
```
Rubilmax: **[LOW]** The new `/^0x/` strip is case-sensitive: an origin passed as `"0Xcafe"` (uppercase `X`) is not stripped, then `isHex("0x0Xcafe")` fails and the origin is silently dropped with a `console.warn` — whereas `"0xcafe"` is accepted. The changeset and JSDoc describe only the lowercase `"0x"` form, so this is an undocumented asymmetry.

_Suggestion: use `/^0x/i` if uppercase prefixes should be tolerated; otherwise acceptable but worth a doc note._

packages/morpho-sdk/src/helpers/metadata.ts
```diff
@@ -39,7 +46,9 @@ export function addTransactionMetadata(
...
 
     if (!isHex(`0x${origin}`))
```
Rubilmax, line 58: **[LOW]** Odd-length origin -> malformed calldata. `addTransactionMetadata` validates `metadata.origin` with `isHex` and `origin.length > 8` (L55) but neither rejects an odd-length fragment. `"0xabc"` now strips to `"abc"` (previously rejected, since `isHex("0x0xabc")` was false) and passes both checks, so `concatHex` appends odd-nibble calldata; viem's `hexToBytes` front-pads the stray nibble rather than throwing, corrupting the trailing analytics byte broadcast on-chain. The `0x`-strip widens this path.

_Suggestion: add an even-length guard (reject when `origin.length` is odd, or pad to a fixed 4-byte width) before `concatHex`._

packages/test/src/mock.ts
```diff
@@ -0,0 +1,334 @@
...
+ *   under test uses, or you may hide upstream `getAddress()` bugs.
+ * @param options.abi - The ABI containing the target function.
+ * @param options.functionName - The name of the `view`/`pure` function
+ *   to intercept. All overloads of this name are matched.
+ * @param options.result - The value `eth_call` should return. Encoded
+ *   via viem's `encodeFunctionResult`; the caller is responsible for
+ *   passing a shape matching the function's declared return type.
+ * @throws `Error` if `functionName` is not present in `abi` (function
+ *   typo / wrong abi). The thrown message names the missing function.
+ * @throws `Error` from `encodeFunctionResult` if `options.result` does
```
Rubilmax: **[LOW]** The `@throws` lines on `mockRead` (L184-186) and `expectReadCall` (L280) use `` @throws `Error` if ... `` instead of the canonical `@throws {ErrorClass} when ...` form from `docs/jsdoc-style.md` §5. The conditions described are accurate; only the tag syntax diverges.

_Suggestion: align to `@throws {Error} when ...` — these are newly-added JSDoc blocks on published sub-exported symbols._

packages/test/src/mock.ts
```diff
@@ -0,0 +1,334 @@
...
+ * @param match.abi - ABI containing the function.
+ * @param match.functionName - Name of the function whose calls should be
+ *   returned.
+ * @returns An array of decoded call data objects (`{ functionName, args }`)
+ *   in the order the mock observed them. Empty array if no calls matched.
+ * @throws `Error` if a candidate call's data fails to decode against the
+ *   provided ABI (signals the caller passed a mismatched abi).
+ *
+ * @example
+ * ```ts
```
Rubilmax, line 293: **[MEDIUM]** The `@example` for the exported `expectReadCall` violates `docs/jsdoc-style.md`: it embeds `0x...` placeholders (`"0x000...dEaD"`, used twice) where the guide says "never embed `0x...` placeholders", and it is not runnable — no `import` lines, no `createMockClient` setup, and it references undefined `address`/`erc20`. The `createMockClient` and `mockRead` examples in the same file are correct templates.

_Suggestion: make the example self-contained (imports, `createMockClient(mainnet)`, a real fixture address e.g. USDC `0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`)._

packages/test/src/mock.ts
```diff
@@ -0,0 +1,334 @@
...
+  return handle.request.mock.calls.flatMap(([call]) => {
+    if (call.method !== "eth_call") return [];
+    const [tx] = (call.params ?? []) as [{ to?: Address; data?: Hex }];
+    if (tx?.to?.toLowerCase() !== targetAddress) return [];
+    if (typeof tx.data !== "string") return [];
+    if (!selectors.has(tx.data.slice(0, 10).toLowerCase())) return [];
+    // Selector matches `match.functionName`, so decode is expected to
+    // succeed. If it doesn't, the caller passed a mismatched abi —
+    // bubble the decode error rather than swallowing it.
+    return [decodeFunctionData({ abi: match.abi, data: tx.data })];
```
Rubilmax, line 363: **[LOW]** `expectReadCall` is generic over `fn extends ContractFunctionName<abi>`, but `decodeFunctionData({ abi: match.abi, data })` returns a union over every function in `abi` rather than narrowing to `fn`. Callers filtering on one function get no compile-time arg typing — `calls[0].args` is a broad union.

_Suggestion: narrow the return type to the requested function, or document that the result is not narrowed to `functionName`._

packages/test/src/mock.test.ts
```diff
@@ -0,0 +1,437 @@
...
+const erc20Abi = parseAbi([
+  "function balanceOf(address owner) view returns (uint256)",
+  "function decimals() view returns (uint8)",
+  "function symbol() view returns (string)",
+]);
+
+const TOKEN: Address = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48";
+const HOLDER: Address = "0x000000000000000000000000000000000000dEaD";
+
+describe("createMockClient", () => {
```
Rubilmax, line 22: **[LOW]** `createMockClient`'s `eth_call` handler has a guard branch (`typeof tx?.to === 'string' && typeof tx.data === 'string'`) that is never exercised with a malformed `eth_call` (missing `to`/`data`), which should fall through to the `unhandled RPC` throw.

_Suggestion: add a test invoking `client.request({ method: 'eth_call', params: [{}] })` and asserting it rejects with `/unhandled RPC/`._

packages/test/src/mock.test.ts
```diff
@@ -0,0 +1,437 @@
...
+    const result = (await handle.client.request({
+      method: "eth_call",
+      params: [{ to: TOKEN, data: nonpayableCallData }, "latest"],
+    })) as Hex;
+    // 7n encoded as uint256 = 32 bytes of 0 with trailing 7.
+    expect(BigInt(result)).toBe(7n);
+  });
+});
+
+describe("expectReadCall", () => {
```
Rubilmax, line 343: **[MEDIUM]** `expectReadCall` tests miss two documented branches: (1) decode-error bubbling — no test where a mismatched ABI makes `decodeFunctionData` throw out of `expectReadCall`; (2) unrelated-call filtering — no test where an `eth_call` to the same address with a selector not in `match.abi` is correctly skipped instead of throwing on decode.

_Suggestion: add a test recording a call then calling `expectReadCall` with a deliberately wrong ABI, and a test recording an unrelated function call and asserting it is filtered out._

Rubilmax (commented):

**Reviewed commit:** `e3be441`

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 3 |
| Low | 7 |

7 of 7 review agents completed. This PR is well-constructed test infrastructure — no critical/high issues; findings are documentation polish, two narrow `metadata.ts` edge cases, and a few coverage gaps.

packages/blue-sdk-viem/src/error.test.ts
0xbulma: Fixed in fc9b73c9 — added two new describe blocks: `getUnsupportedVaultV2Adapter` (6 tests: plain Error, non-Error values, BaseError without revert, matching errorName extracts adapter, non-matching errorName returns null, missing args returns null, nested cause walk) and `InvalidPermitDomainChainIdError` / `InvalidPermitDomainVerifyingContractError` / `UnsupportedPermitDomainExtensionsError` (field preservation + message quoting + edge cases incl. defensive extensions-array copy and undefined values).

packages/liquidation-sdk-viem/src/flashbots.test.ts
0xbulma: Fixed in fc9b73c9 — added the 5xx test using your suggested explicit-content-type approach (`reply(500, { error: 'relay overloaded' }, { 'content-type': 'application/json' })`). `response.json()` parses cleanly in the error branch now, and the test asserts the rejection message contains the relay error string. The auto-increment-id branch remains deferred to fork tests (the previous comment about nock + Vitest concurrent interaction is kept).

packages/morpho-sdk/src/helpers/encodeDeallocation.test.ts
0xbulma: Fixed in fc9b73c9 — the test now asserts `(e as NonPositiveAssetAmountError).message` contains `ADAPTER`, matching the test title.

packages/morpho-sdk/src/helpers/metadata.ts
0xbulma: Fixed in fc9b73c9 — switched the regex to `/^0x/i` so uppercase `0X` is also stripped. JSDoc updated to document "0xcafe / 0Xcafe / cafe all produce the same appended 4-byte origin tag". Added a test pinning `0Xcafe` produces the same calldata as `0xcafe`/`cafe`.

packages/morpho-sdk/src/helpers/metadata.ts
0xbulma: Fixed in fc9b73c9 — added an even-length guard: `if (origin.length % 2 !== 0) throw Error("Calldata origin must have an even number of hex characters")`. Two new tests pin the rejection: `abc` (raw 3-char) and `0xabc` (3-char after strip). JSDoc updated to call out the rejection. Good catch — the round-2 0x-strip fix widened this path; the explicit guard restores byte-alignment at the source rather than relying on `isHex`'s previous gating.

packages/test/src/mock.ts
0xbulma: Fixed in fc9b73c9 — both `mockRead` and `expectReadCall` now use `@throws {Error} when …` per docs/jsdoc-style.md §5. Also updated mockRead's second `@throws` to match the actual per-overload error ("options.result does not match any return-type shape of any overload") since the round-3 fix changed the throw conditions.

packages/test/src/mock.ts
0xbulma: Fixed in fc9b73c9 — rewrote the `@example` as a self-contained runnable snippet: explicit `import` lines for the 4 symbols + viem helpers, real USDC fixture address (`0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`), the dead-address constant (`0x000…dEaD`), `createMockClient(mainnet)` + `mockRead` setup, then `readContract` + `expectReadCall` with a comment on the expected output.

packages/test/src/mock.ts
0xbulma: Documented in fc9b73c9 — added a `@returns` note: "The decoded `args` are typed as a union over every function in `match.abi` rather than narrowed to `match.functionName`; callers that need narrow argument types should cast at the call site or filter the ABI to a single-entry array before passing it in." Type-narrowing the return without breaking the public signature would mean restructuring the generic to derive a single-item-ABI projection — left for a separate change since this is documenting current behavior.

packages/test/src/mock.test.ts
0xbulma: Fixed in fc9b73c9 — added `eth_call without to/data falls through to the unhandled-RPC throw`: calls `client.request({ method: 'eth_call', params: [{}] })` and asserts rejection with `/unhandled RPC eth_call/`. Exercises the previously-uncovered guard branch in `createMockClient`'s handler.

packages/test/src/mock.test.ts
0xbulma: Partially fixed in fc9b73c9 — added the unrelated-call filtering test (records `decimals`, asks expectReadCall about `totalSupply` with a wholly-unrelated ABI, asserts `[]`). The decode-error-bubble branch turns out to be unreachable since the round-3 selector pre-filter intercepts mismatched-ABI calls before `decodeFunctionData` runs (wrong-arity `balanceOf(uint256)` has a different selector than `balanceOf(address)`, so it's filtered out and yields `[]`). Updated the JSDoc to reflect the actual no-throw behavior, and replaced the planned decode-bubble test with a wrong-arity test that pins the new `[]` semantics.

Rubilmax (approved):

PR #604, opened 2026-05-05
feat(morpho-sdk): MorphoClient.extend() with MorphoEntity classes

Foulks-Plb:
## Summary

- `MorphoClient.extend({ name: EntityClass })` lets integrators register custom entity **classes** on the client, mirroring the built-in `MorphoVaultV1` / `MorphoMarketV1` pattern. Returns a **new** client (statelessness preserved) typed as `this & ExtensionInstances<TExtension>`, so successive extends accumulate.
- Every registered class must extend the new `MorphoEntity` base, which holds the standard `protected readonly client: MorphoClientType` binding. The constructor signature `(client, ...args)` becomes the call shape `client.<name>(...args)`.
- Two-staged validation: name/identifier/shape + `extends MorphoEntity` at registration (`ExtensionNameCollisionError`, `InvalidExtensionNameError`, `InvalidExtensionShapeError`, `InvalidEntityClassError`); structural action/transaction/requirement checks lazily on each call via a Proxy (`InvalidActionShapeError`, `InvalidTransactionShapeError`, `InvalidRequirementShapeError`). Methods returning non-action values (fetchers, getters) pass through unchanged. `instanceof` against the integrator's class is preserved.
- Integrator keeps full control of `deepFreeze` and `metadata` injection. Validators do not refreeze and do not auto-inject `client.options.metadata`.

## Design

[`docs/tibs/TIB-2026-05-05-morpho-client-extend.md`](./docs/tibs/TIB-2026-05-05-morpho-client-extend.md)

## DevEx

```ts
import { MorphoClient, MorphoEntity, type MorphoClientType } from "@morpho-org/morpho-sdk";
import type { Address } from "viem";

class MyLending extends MorphoEntity {
  constructor(
    client: MorphoClientType,
    public readonly vault: Address,
    public readonly chainId: number,
  ) {
    super(client);
  }

  deposit({ amount, user }: { amount: bigint; user: Address }) {
    return {
      buildTx: () => ({
        to: this.vault,
        value: 0n,
        data: encodeMyDeposit({ vault: this.vault, amount, user }),
        action: { type: "myLendingDeposit", args: { amount, user } },
      }),
      getRequirements: async () => [
        await fetchApproval(this.client.viemClient, this.vault, amount, user),
      ],
    };
  }

  // Non-action methods (fetchers, getters) pass through unchanged.
  async getData() {
    return fetchMyVault(this.vault, this.client.viemClient);
  }
}

const client = new MorphoClient(viemClient, { supportSignature: true })
  .extend({ myLending: MyLending });

const tx = client.myLending(vault, 1).deposit({ amount: 1n, user }).buildTx();
```

## Test plan

- [x] Unit tests for `validateExtensionMap` and `wrapEntityInstance` (15 tests, colocated at `packages/morpho-sdk/src/helpers/validateExtension.test.ts`): collision, identifier shape, `extends MorphoEntity` enforcement, action wrapping, non-action pass-through, `instanceof` preservation, freeze passthrough, all error paths.
- [x] Unit tests for `MorphoClient.extend()` (20 tests, colocated at `packages/morpho-sdk/src/client/morphoClient.test.ts`): happy path, chained extends, original-not-mutated, freeze passthrough, metadata not auto-injected, entity reads `client.options`, all 4 error paths, non-writable extension property, built-in entities still functional.
- [x] `tsc --noEmit` clean.
- [x] `biome check packages/morpho-sdk/src` clean (no errors, no warnings).
- [ ] CI fork tests (require RPC env vars; not run locally).

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1778143706564599)


Foulks-Plb (commented):
_(review withdrawn)_

packages/morpho-sdk/src/types/error.ts
```diff
@@ -391,3 +391,111 @@ export class NegativeMinSharePriceError extends Error {
...
+  constructor(entityName: string, reason: string) {
+    super(`Entity "${entityName}" returned an invalid shape: ${reason}`);
+  }
+}
+
+/**
+ * Thrown when an integrator-provided action method returns an object missing `buildTx` (or whose
+ * `buildTx` / `getRequirements` is not a function).
+ */
+export class InvalidActionShapeError extends Error {
```
Foulks-Plb, line 458: **[Important]** `InvalidActionShapeError` (and its siblings `InvalidTransactionShapeError` at line 474, `InvalidRequirementShapeError` at line 490) take an options-object `{ entityName, methodName, reason }`, while the 5 other new errors and every pre-existing error in this file use positional args.

Suggestion: pick one shape — converging on positional matches the rest of `error.ts` and is the smaller change. Same finding as the previous review pass — still outstanding.

packages/morpho-sdk/src/types/error.ts
```diff
@@ -391,3 +391,111 @@ export class NegativeMinSharePriceError extends Error {
...
+      `Extension "${name}" is not a valid entity class: ${reason}. Pass a class that extends MorphoEntity.`,
+    );
+  }
+}
+
+/**
+ * Thrown when an integrator-provided entity factory returns something that is not a record of
+ * action methods. The entity name and the offending property/reason are included.
+ */
+export class InvalidEntityShapeError extends Error {
```
Foulks-Plb: **[Important]** `InvalidEntityShapeError` is exported but **never thrown** anywhere (grep confirmed). Dead public surface that locks into the SemVer minor and pollutes the stability contract.

Secondary: the `extend()` JSDoc `@throws` list at `morphoClient.ts:189` references neither it nor a wired path. Drop the export, or wire it into a real failure path and add it to the `@throws` list. Same finding as the previous review pass — still outstanding.

packages/morpho-sdk/src/types/error.ts
```diff
@@ -391,3 +391,111 @@ export class NegativeMinSharePriceError extends Error {
...
   }
 }
+
+/**
+ * Thrown by `MorphoClient.extend()` when an extension key collides with an existing client
+ * member — built-in (`viemClient`, `options`, `vaultV1`, `vaultV2`, `marketV1`, `extend`, …),
+ * inherited from `Object.prototype` (`constructor`, `toString`, …), a previously registered
+ * extension, or the `then` Promise-thenable trap. Pick a different name.
+ */
+export class ExtensionNameCollisionError extends Error {
```
Foulks-Plb, line 401: **[Medium]** Most of the 8 new error-class messages stop at description without remediation. Per root §3 format (`Borrow amount X exceeds safe maximum Y. Reduce borrow or increase collateral.`), each message should end with a `Do X.` instruction.

Applies to: `InvalidExtensionNameError` (`Rename to start with a lowercase letter and use only ASCII letters/digits.`), `InvalidExtensionShapeError` (`Pass a non-empty record of { name: EntityClass } to .extend().`), `InvalidEntityShapeError` (if kept), `InvalidActionShapeError` (`Return { buildTx: () => Transaction, getRequirements?: () => Promise<...> }.`), `InvalidTransactionShapeError` (`Return { to: Address, value: bigint, data: Hex, action: { type, args } }.`), `InvalidRequirementShapeError` (`Each item must be a Transaction or a Requirement.`).

None carry `@param` tags either — at minimum document each constructor arg per §6.

packages/morpho-sdk/src/client/morphoClient.ts
```diff
@@ -47,19 +54,40 @@ export class MorphoClient implements MorphoClientType {
    * );
    * ```
    */
+  /** @internal Entity classes registered through `.extend()`, surfaced as enumerable factory props. */
+  private readonly _extensions: ReadonlyMap<string, EntityConstructor>;
```
Foulks-Plb: **[Important]** Constructor JSDoc (lines 36-56, with `@param viemClient`, `@param _options`, `@example`) is orphaned by the `_extensions` field declaration on line 58 between it and the constructor at line 61. TypeDoc/IDEs attach the constructor's JSDoc (and `@example`) to `_extensions`, leaving the constructor undocumented.

Suggestion: move the `_extensions` field declaration above the constructor's JSDoc block, or move the JSDoc directly above `constructor(...)`. Same finding as the previous review pass — still outstanding.

packages/morpho-sdk/src/client/morphoClient.ts
```diff
@@ -47,19 +54,40 @@ export class MorphoClient implements MorphoClientType {
...
       readonly supportSignature?: boolean;
       readonly supportDeployless?: boolean;
       readonly metadata?: Metadata;
     },
+    /**
+     * @internal Used by `.extend()` to carry registered entity classes when cloning. Integrators
+     * must not pass this directly — call `.extend(map)` on an existing client instead.
+     */
+    extensions?: ReadonlyMap<string, EntityConstructor>,
   ) {
```
Foulks-Plb, line 72: **[Important — security]** The third constructor parameter `extensions: ReadonlyMap<string, EntityConstructor>` is `@internal` JSDoc only — TypeScript markers are erased at runtime. A direct `new MorphoClient(viem, opts, handCraftedMap)` bypasses `validateExtensionMap` entirely: a hand-rolled map can register a non-`MorphoEntity` constructor, a name colliding with a built-in (`Object.defineProperty(this, name)` shadows prototype methods like `marketV1` / `extend`), or even `then` (re-introducing the thenable hijack).

Suggestion: re-run `validateExtensionMap` inside the constructor loop before the `Object.defineProperty` calls, or move the merged-map handoff to a private static factory the public constructor cannot reach. Same finding as the previous review pass — still outstanding.

packages/morpho-sdk/src/client/morphoClient.ts
```diff
@@ -47,19 +54,40 @@ export class MorphoClient implements MorphoClientType {
...
     this.options = {
       ..._options,
```
Foulks-Plb, line 74: **[Important — security]** `this.options` is a plain spread, never `Object.freeze`d. Integrator entity code holding the client reference can mutate `client.options.metadata` (e.g. flip the analytics origin) and the change is observed by every subsequent built-in `buildTx` that injects metadata.

Additional concern: `_options` (line 64) is `readonly` (TS-only) — same root cause, still mutable at runtime, and forwarded by reference to `.extend()` clones via line 243.

Suggestion: `Object.freeze(this.options)` after the spread; either drop `_options` retention (recompute from `this.options`) or store a frozen copy.

packages/morpho-sdk/src/client/morphoClient.ts
```diff
@@ -47,19 +54,40 @@ export class MorphoClient implements MorphoClientType {
...
       supportSignature: _options?.supportSignature ?? false,
       supportDeployless: _options?.supportDeployless,
     };
+
+    this._extensions = extensions ?? new Map();
+    for (const [name, EntityClass] of this._extensions) {
```
Foulks-Plb: **[Important — security]** `this._extensions` is declared `private readonly ReadonlyMap` but is a plain runtime `Map`. Any integrator-supplied entity (received via `super(client)`) can call `(this.client as any)._extensions.set('marketV1', MaliciousEntity)`. The poisoned entry is then carried into every subsequent `.extend()` clone via `new Map(this._extensions)`, where it would still pass the `name in client` check on the new prototype-fresh instance and silently shadow the built-in `marketV1` / `vaultV1` / `vaultV2` factories on the cloned client.

Note: `Object.freeze(map)` is insufficient (Map ignores freeze). Use a true JS `#extensions` private field, store the entries in a closure, or wrap the Map in an immutable Proxy whose `set` / `delete` / `clear` throw. Same finding as the previous review pass — still outstanding.

packages/morpho-sdk/src/helpers/validateExtension.ts
```diff
@@ -0,0 +1,244 @@
...
+ * callbacks as constructor args. `catch` / `finally` are NOT part of the thenable protocol and
+ * pose no automatic-invocation risk.
+ *
+ * @internal
+ */
+const PROMISE_TRAP_NAMES: ReadonlySet<string> = new Set(["then"]);
+
+const VALID_NAME_REGEX = /^[a-z][a-zA-Z0-9]*$/;
+
+const isPlainObject = (value: unknown): value is Record<string, unknown> =>
```
Foulks-Plb: **[Medium]** `isPlainObject` is now duplicated verbatim here and at `packages/morpho-sdk/src/types/action.ts:292` (after this PR's extraction of `isTransactionShape`).

Per root §1 (Modularity / single source of truth), promote it to a shared internal helper (e.g. a small `helpers/typeGuards.ts`) and import from both sites.

packages/morpho-sdk/src/helpers/validateExtension.ts
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * The Proxy preserves `instanceof` against the integrator's class. Validation is non-destructive:
+ * the integrator's freeze and metadata choices on the returned `Transaction` are preserved.
+ *
+ * @param entityName - Name the entity is registered under (used in error messages).
+ * @param instance - Freshly constructed entity instance.
+ * @returns A Proxy over `instance`.
+ */
+export function wrapEntityInstance<T extends MorphoEntity>(
+  entityName: string,
```
Foulks-Plb, line 120: **[Medium]** `wrapEntityInstance` (and `validateExtensionMap` at line 52) lack the `@internal` JSDoc tag despite being intended as internal helpers. Per root §3 (`Internal symbols carry @internal JSDoc and do not participate in the stability contract`), add `@internal` so they're excluded from the published TypeDoc reference.

Secondary: `wrapEntityInstance` could use a one-line `@example` showing the Proxy invariant for future maintainers, and `@throws` lazy-error enumeration for the contract that integrators may pattern-match on.

packages/morpho-sdk/src/helpers/validateExtension.ts
```diff
@@ -0,0 +1,244 @@
...
+      return value;
+    },
+  });
+}
+
+// "Looks like an action attempt" = the integrator put a `buildTx` key on the returned object.
+// Whether the value is callable is checked by `wrapAction`, which raises `InvalidActionShapeError`
+// when `buildTx` is not a function. This separation gives the integrator a typed error for
+// `{ buildTx: 42 }` instead of a confusing `TypeError: foo.buildTx is not a function` later.
+const looksLikeActionAttempt = (
```
Foulks-Plb, line 167: **[Important — silent failure]** Async action methods bypass all shape validation. `wrapEntityInstance`'s Proxy `get` handler checks `looksLikeActionAttempt(result)` synchronously — a method declared `async deposit() { return { buildTx, getRequirements } }` returns a Promise (no `buildTx` key on the Promise), so the Proxy passes it through unwrapped, the awaited `{ buildTx, getRequirements }` is never wrapped, and `assertValidTransactionShape` / `assertValidRequirementsShape` never run.

This silently violates the documented invariant in the JSDoc on `extend()` and on `ExtensionAction`, and integrators following the natural async-fetcher-then-action style get no validation.

Suggestion: detect Promise returns inside the Proxy and either reject with a typed error explaining actions must be sync, or `await` the result and re-run `looksLikeActionAttempt` on the resolved value before wrapping. Add a test pinning whichever path is chosen. Same finding as the previous review pass — still outstanding.

packages/morpho-sdk/src/helpers/validateExtension.ts
```diff
@@ -0,0 +1,244 @@
...
+      reason: "`getRequirements` must be a function when provided",
+    });
+  }
+
+  const buildTx = action.buildTx as ExtensionAction["buildTx"];
+  const getRequirements = action.getRequirements as
+    | ExtensionAction["getRequirements"]
+    | undefined;
+
+  const wrapped: {
```
Foulks-Plb: **[Medium]** `wrapAction` constructs a fresh `{ buildTx, getRequirements? }` wrapper and discards any other fields the integrator placed on the returned action object (e.g. `description`, `version`, helper sub-methods, custom symbols). Silent data loss — undocumented.

Suggestion: preserve unknown properties via spread (`{ ...action, buildTx: ..., getRequirements: ... }`), or document the contract explicitly in `ExtensionAction` JSDoc and in `MorphoClient.extend()` JSDoc.

packages/morpho-sdk/src/types/action.ts
```diff
@@ -289,6 +289,64 @@ export interface RequirementSignature {
...
+ * @returns `true` if the value is structurally a transaction.
+ * @example
+ * ```ts
+ * if (isTransactionShape(maybeTx)) {
+ *   void maybeTx.to;
+ *   void maybeTx.action.type;
+ * }
+ * ```
+ */
+export function isTransactionShape(
```
Foulks-Plb, line 310: **[Important]** `isTransactionShape` is a NEW exported function with **no direct unit test**. Per root §5 (every exported function has a unit test), add a colocated test (`action.test.ts` next to source) covering: happy path, plus typeof-mismatch edge cases — `value` as number (not bigint), `data` as null, `to` as null/undefined, `action` missing, `action.type` non-string, `action.args` missing/non-object, plain non-object inputs (null, undefined, string, array).

packages/morpho-sdk/src/types/action.ts
```diff
@@ -289,6 +289,64 @@ export interface RequirementSignature {
...
+
+/**
+ * Strict structural type-guard for {@link Requirement}. Returns `true` when `value` matches
+ * `{ sign: function, action: { type: string, args: object } }`. Stricter than
+ * {@link isRequirementSignature}, which only checks for the presence of `sign`.
+ *
+ * @param value - Anything (typically a candidate requirement).
+ * @returns `true` if the value is structurally a requirement.
+ */
+export function isRequirementShape(value: unknown): value is Requirement {
```
Foulks-Plb, line 338: **[Important]** `isRequirementShape` is a NEW exported function with no unit test. Add colocated tests: happy path with `{ sign, action: { type, args } }`, plus `sign` not a function, `action` missing, `action.type` non-string, `action.args` missing/non-object, null/undefined/array input.

Secondary: `@example` is missing here AND on the tightened `isRequirementApproval` (line 350) / `isRequirementAuthorization` (line 369). Per §6 each exported function gets one `@example`. Mirror the `isTransactionShape` example just above.

packages/morpho-sdk/src/types/action.ts
```diff
@@ -289,6 +289,64 @@ export interface RequirementSignature {
...
+  );
+}
+
+/**
+ * Returns `true` when the given requirement is an ERC-20 approval transaction.
+ *
+ * @param requirement - Candidate requirement (or `undefined`).
+ * @returns `true` if it's a `Transaction<ERC20ApprovalAction>`.
+ */
 export function isRequirementApproval(
```
Foulks-Plb, line 360: **[Medium]** `isRequirementApproval` and `isRequirementAuthorization` were tightened to delegate to `isTransactionShape` (which now demands `value: bigint`, `data: string`, `action.args: object`). Existing fork tests in `test/actions/requirements/approval.test.ts` exercise the happy path on real fixtures, but **no regression test pins the new strictness** — e.g. a malformed `{ to, action: { type: 'erc20Approval' } }` lacking `value` / `data` / `args` should now return `false` where it previously returned `true`.

Suggestion: add a focused unit test asserting both the new rejections and the preserved acceptance of well-formed inputs.

packages/morpho-sdk/src/types/client.ts
```diff
@@ -22,4 +24,8 @@ export interface MorphoClientType {
   vaultV1: (vault: Address, chainId: number) => VaultV1Actions;
   vaultV2: (vault: Address, chainId: number) => VaultV2Actions;
   marketV1: (marketParams: MarketParams, chainId: number) => MarketV1Actions;
+
+  extend: <const TExtension extends ExtensionMap>(
```
Foulks-Plb, line 38: **[Medium]** `MorphoClientType.extend` declares its return as `MorphoClientType & ExtensionInstances<TExtension>` while the concrete `MorphoClient.extend` returns `this & ExtensionInstances<TExtension>`. Through the interface, a subclass of `MorphoClient` (or any other `MorphoClientType` impl) loses its concrete type after `.extend()`.

Secondary: this interface (and its members `viemClient`, `options`, `vaultV1` / `V2` / `marketV1`, `extend`) carries no JSDoc — at minimum document `extend` here, since it's the new surface this PR adds.

Suggestion: align the interface declaration to `this & ExtensionInstances<TExtension>`.

packages/morpho-sdk/src/types/morphoEntity.ts
```diff
@@ -0,0 +1,46 @@
...
+ *       buildTx: () => ({
+ *         to: this.vault,
+ *         value: 0n,
+ *         data: "0x",
+ *         action: { type: "myLendingDeposit", args: { amount, user } },
+ *       }),
+ *     };
+ *   }
+ * }
+ * ```
```
Foulks-Plb: **[Medium]** Filename `morphoEntity.ts` clashes with sibling `entity.ts` (which already exports `VaultParams`). Sibling files in `src/types/` use single-concept names: `action.ts`, `client.ts`, `error.ts`, `extension.ts`, `metadata.ts`. Having both `entity.ts` and `morphoEntity.ts` invites confusion.

Suggestion: fold `MorphoEntity` into `entity.ts` (one home for entity-layer types), or rename `entity.ts` so the two files share a clear scheme.

Secondary: the `protected client` field has no JSDoc — add a one-line description since it's part of the documented contract subclasses depend on. Same finding as the previous review pass — still outstanding.

packages/morpho-sdk/src/helpers/validateExtension.test.ts
```diff
@@ -0,0 +1,303 @@
...
+        return { buildTx: () => frozen };
+      }
+    }
+    const wrapped = wrapEntityInstance("e", new Frozen(stubClient));
+    const tx = wrapped.doThing().buildTx();
+    expect(Object.isFrozen(tx)).toBe(true);
+    expect(tx).toBe(frozen);
+  });
+
+  test("error: buildTx returns malformed tx (missing data)", () => {
```
Foulks-Plb, line 169: **[Medium]** `assertValidTransactionShape` has 6 distinct fail branches (`to` non-string, `value` non-bigint, `data` non-string, `action` non-object, `action.type` non-string, `action.args` non-object). Only the missing-`data` branch is asserted.

Additional gaps in this file:
- Identifier edges: only `Bad-Name` and `_hidden` covered — add `MyEntity` (uppercase), `café` (unicode), `🚀` (emoji), `""` (empty).
- No test for an integrator method returning `{ getRequirements }` only (no `buildTx`) — `looksLikeActionAttempt` does NOT detect this; pin the passthrough.
- No test for an `async` action method (silent gap pointed out at `validateExtension.ts:146`).

Similar gaps in `morphoClient.test.ts`: only 2-deep chain tested (add 3+); the non-writable test asserts only the throw, not the full descriptor — strengthen with `expect(Object.getOwnPropertyDescriptor(...)).toMatchObject({ writable: false, configurable: false, enumerable: true })`.

Foulks-Plb (commented):
## Parallel PR Review (Claude) — re-run

**Reviewed commit:** `8ce3c430`

| Severity | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 7 |
| 🔵 Medium | 11 |
| 🔵 Minor | 14 |

### What's still outstanding from the previous review

Resolved since last pass:
- Identifier-collision dynamic check via `name in client` ✅
- `then` thenable hijack guard ✅
- `any[]` factory → `unknown[]` ✅
- Symbol / null / non-fn `buildTx` cases pinned by tests ✅
- Extracted `isTransactionShape` / `isRequirementShape` ✅

Still open (top blockers):

1. **Async action methods bypass all shape validation** — `wrapEntityInstance` checks `looksLikeActionAttempt(result)` synchronously. If the integrator method is `async` (returns `Promise<{ buildTx, … }>`), the predicate is false (Promise has no `buildTx` key), the Promise passes through untouched, and `buildTx()` / `getRequirements()` are never validated. Detect thenables and either reject or `await` + re-validate.
2. **`_extensions` is `private readonly` (TS only)** — at runtime it's a plain `Map`. Integrator entity code can `(this.client as any)._extensions.set('marketV1', Malicious)` and the poisoned entry is carried into every `.extend()` clone via `new Map(this._extensions)`. Use a JS `#extensions` private field, or wrap the Map in a frozen accessor.
3. **`options` not frozen** — same TS-readonly-only situation. An integrator can mutate `client.options.metadata` and every subsequent built-in `buildTx` observes the change. `Object.freeze(this.options)` after construction.
4. **Third constructor arg `extensions` is `@internal` JSDoc only** — `new MorphoClient(viem, opts, hand-rolled-map)` bypasses `validateExtensionMap` entirely (no name validation, no `MorphoEntity` check, no thenable trap). Run validation inside the constructor on every code path.
5. **`InvalidEntityShapeError` is exported but never thrown** — dead public surface that locks into the SemVer minor.
6. **Inconsistent error-class signatures** — `InvalidActionShapeError`, `InvalidTransactionShapeError`, `InvalidRequirementShapeError` use options-objects; the 5 other new errors and every pre-existing error in `error.ts` use positional args. File convention is positional.
7. **No unit tests for the new `isTransactionShape` / `isRequirementShape` exports** — per CLAUDE.md §5 every exported function gets a unit test.

### Other significant findings

- Constructor JSDoc on `MorphoClient` (lines 36-56) is now orphaned: TypeDoc/IDEs attach it to the `_extensions` field on line 58 instead of the constructor below. Move the field above the JSDoc, or move the JSDoc directly onto the constructor.
- `wrapAction` constructs a fresh `{ buildTx, getRequirements? }` wrapper, silently dropping any other fields the integrator placed on the action object. Either spread or document.
- `MorphoClientType.extend` declares `MorphoClientType & ExtensionInstances<TExt>` while the concrete signature uses `this & ExtensionInstances<TExt>`. Through the interface, subclasses lose type after `.extend()`. Align both to `this & …`.
- The 8 new error classes (and the tightened type-guards `isRequirementApproval` / `isRequirementAuthorization` / new `isTransactionShape` / `isRequirementShape`) are missing `@example` blocks; messages like `Action "e.deposit" has an invalid shape: …` lack the actionable remediation that root §3 expects (e.g., `Return { buildTx: () => Transaction, getRequirements?: () => Promise<...> }.`).
- `validateExtensionMap` / `wrapEntityInstance` lack the `@internal` JSDoc tag despite being intended as internal — they currently leak into TypeDoc.
- `MorphoClientType.extend` itself has no JSDoc on the interface side.
- `morphoEntity.ts` clashes with sibling `entity.ts` (which already exports `VaultParams`). Same naming concern as last review — could fold `MorphoEntity` into `entity.ts`.
- Coverage gaps: only 2-deep chained extends tested; 5 of 6 transaction-shape failure branches uncovered (only missing-`data`); identifier edges (`MyEntity`, unicode, empty) untested; non-writable test asserts only the throw, not the full descriptor; `looksLikeActionAttempt({getRequirements})` passthrough untested.

### Minor (summary only — no inline)

- `isPlainObject` duplicated verbatim in `validateExtension.ts:33` and `action.ts:292`. Hoist to a shared helper.
- AGENTS.md / TIB still mention a centralised `RESERVED_MORPHO_CLIENT_NAMES` constant — it no longer exists (replaced by `name in client`).
- `_options` listed as a reserved name in `client/AGENTS.md` is unreachable — extension keys starting with `_` are rejected by the identifier regex first.
- `assertValidTransactionShape` only checks `typeof tx.to === 'string'` and `typeof tx.data === 'string'`. Tighten with `isAddress(tx.to)` / `isHex(tx.data)` from viem.
- TOCTOU window on the registered class — `MyEntity.prototype = …` after `extend()` could defeat the `instanceof MorphoEntity` check. Realistic only against malicious authors; could re-assert in `wrapEntityInstance`.
- Proxy `apply(target, args)` (deliberate trade-off) means `compose() { return this.deposit(...); }` returns the inner action raw — document.
- Symbol-keyed action getters bypass validation (pinned by test, but undocumented in JSDoc on `extend`).
- `extend()` returns `this & ExtensionInstances` but always constructs a base `MorphoClient` — silent type lie for subclasses.
- `MorphoEntity` constructor doesn't guard against `super(undefined)` — opaque `TypeError` later.
- `ExtensionNameCollisionError` message is generic ("collides with an existing client member"). Could disambiguate built-in vs Object.prototype vs duplicate vs thenable trap.
- File-wide `biome-ignore-all noExplicitAny` in `validateExtension.ts` — prefer per-line.
- Several test biome-ignore reasons say only `"deliberate misuse"` — expand each to explain what's being misused.
- `validateExtension.test.ts:59` passes `[]` for the client arg while everywhere else passes `{}` — copy-paste slip.
- `_extensions` field uses single-line `/** @internal ... */` while the file prefers multi-line JSDoc — pick one style.

_This is an automated parallel review across 7 specialized agents. Findings deduplicated within 3-line tolerance, highest severity wins. Several of the findings above match the previous review pass — they remained outstanding through the latest commits._

Foulks-Plb:
**Review follow-up — items addressed and items intentionally not implemented (with reasoning).**

Pushed `09a01d1e` addressing 16 of the 17 inline findings + minor body items. Below is the status of every finding from review `#pullrequestreview-4236996363`:

### ✅ Implemented (16 + minors)

1. **Inconsistent error signatures** → 3 options-object errors converted to positional
2. **`InvalidEntityShapeError` dead** → deleted
3. **Constructor JSDoc orphaned** → field declaration moved
4. **3rd constructor arg bypasses validation** → `validateExtensionMap` now re-runs in the constructor for non-empty supplied maps
5. **`options` mutable** → `Object.freeze(this.options)` at construction
6. **`_extensions` runtime-mutable** → converted to JS `#extensions` private field
7. **Async actions bypass validation** → now throw `InvalidActionShapeError` synchronously (matches root §1 "actions are sync")
8. **No tests for `isTransactionShape` / `isRequirementShape`** → colocated `action.test.ts` added with 12 cases incl. regression for tightened `isRequirementApproval` / `Authorization`
9. **Error messages lack remediation** → `Do X.` instruction appended to all 7 extension errors
10. **`isPlainObject` duplicated** → hoisted to `helpers/typeGuards.ts`
11. **Missing `@internal`** on validators → tag added
12. **`wrapAction` drops extra fields** → spread now preserves them
13. **Interface vs concrete `this &` mismatch** → interface aligned + JSDoc on `MorphoClientType.extend`
14. **`morphoEntity.ts` filename clash** → `MorphoEntity` folded into `entity.ts`
15. **Test branch coverage** → added: tx with `value:number`, tx with non-string `action.type`, identifier edges (uppercase/unicode/emoji/empty), 3-deep chain, full descriptor assertion, async action rejection, extra fields preserved
16. **Stale `RESERVED_MORPHO_CLIENT_NAMES` references** in AGENTS.md / TIB → updated
17. Body minors: terse `biome-ignore: deliberate misuse` reasons → expanded; `[]` vs `{}` test inconsistency → fixed; Proxy trade-offs (symbol bypass, `apply(target)`, async rejection) → documented inline in `wrapEntityInstance` JSDoc

### 💬 Intentionally not implemented (with reasoning)

1. **Tighten `tx.to` with `isAddress`, `tx.data` with `isHex`** — Out of scope. The integrator may legitimately encode custom calldata that isn't strictly viem-Hex (e.g. EIP-7702 inits, intent-style payloads). Validating `Hex` here would block legitimate use cases. The wallet/RPC layer will reject malformed bytes downstream. A test of `validTx().to === "0x09"` is also harmless — those are stub addresses for assertion only, never sent on-chain.

2. **TOCTOU on `value.prototype` post-registration** — Defensive against malicious extension authors, but the threat model assumes the integrator controls their own extension classes. Re-asserting `instance instanceof MorphoEntity` in `wrapEntityInstance` would catch it but adds runtime cost on every entity instantiation for an attack vector that requires the integrator to attack themselves. Skipped.

3. **`MorphoEntity` constructor sans `super(client)` guard** — TS catches `super(client)` omission at compile time; `this.client.viemClient` reads then surface the standard `TypeError: Cannot read properties of undefined`. A typed runtime guard would be belt-and-suspenders for a misuse pattern that's already caught statically. The opaque error is acceptable here.

4. **`ExtensionNameCollisionError` generic message** — Was deliberately simplified in commit `77d99e0c` when the validator switched to `name in client`. Distinguishing built-in / Object.prototype / duplicate / thenable requires re-introspecting the source of the collision after the fact, which complicates the validator without clear DX win — the message + the name itself usually point straight to the cause.

5. **File-wide `biome-ignore-all noExplicitAny` in `validateExtension.ts`** — Per-line annotations would multiply across the Proxy `get` trap, the closure factory, and `wrapAction` — all sites where the integrator's variadic args legitimately erase to `any`. The file-wide annotation with an explanatory comment at the top is the cleaner trade-off here (similar to viem's own extension internals).

### Other already-addressed items not in the review (carried from previous pass)

- `then` thenable hijack guard via `PROMISE_TRAP_NAMES`
- `name in client` dynamic collision check (replaced static reserved list)
- `defineEntity` removed
- `any[]` → `unknown[]` in factory closure

Tests: 69 passing (+25 vs prior pass), `tsc --noEmit` clean, `biome check` clean.

docs/tibs/TIB-2026-05-05-morpho-client-extend.md
```diff
@@ -0,0 +1,337 @@
...
+1. Subclass `MorphoClient`. Couples the integrator to our class hierarchy and breaks `instanceof`
+   checks against vendored versions.
+2. Maintain a parallel client beside ours, duplicating viem-client wiring, chain checks, and
+   options plumbing. Fragments the call site (`morpho.vaultV1(...)` vs `myProto.foo(...)`).
+
+Neither is acceptable for a stateless, composable SDK. viem solved a similar shape with
+`client.extend(actions)`, which returns a new client carrying the integrator's methods on top of
+the original surface.
+
+This TIB introduces an analogous mechanism on `MorphoClient`, but **keyed on entity classes**
```
Rubilmax, line 32: Should we not get rid of `MorphoClient`? I'm all for exposing extension helpers, but maintaining this `MorphoClient` appears to already be a burden

docs/tibs/TIB-2026-05-05-morpho-client-extend.md
```diff
@@ -0,0 +1,337 @@
...
+and offers no validation surface — typos in method names are silent until call time.
+
+### Alternative 2: Callback-based `extend((c) => ({ name: factoryFn }))` with plain object factories
+
+`extend` takes a callback returning a record of factory functions; each factory returns a plain
+object whose methods build actions.
+
+**Why rejected:** This was the first iteration of this TIB. The factory-function shape diverges
+from the existing `MorphoVaultV1` / `MorphoMarketV1` / `MorphoEntity` pattern: integrators end up
+with two mental models (classes for built-ins, callbacks for extensions). The class-based shape
```
Rubilmax, line 234: IMO this is what i would be in favor: make our move towards callbacks NOW instead of keeping this "classes for built-ins"
So we dont even need to expose viem extension helpers

docs/tibs/TIB-2026-05-05-morpho-client-extend.md
```diff
@@ -0,0 +1,337 @@
...
+they already understand, and matches root `AGENTS.md` §1 ("entities are classes that read state
+and validate").
+
+### Alternative 3: Flat method extension (viem `walletClient.extend(walletActions)` style)
+
+`extend` adds methods directly to the client root rather than namespaced entity factories.
+
+**Why rejected:** The SDK's surface is organised by **entity** (`vaultV1`, `marketV1`, etc.), and
+`client/CLAUDE.md` explicitly defines the client as a factory for entities. A flat extend would
+mix integrator action methods with our entity factories, blurring the contract and making
```
Rubilmax, line 245: Disagreed

PR #605, opened 2026-05-06
feat(evm-simulation): per-tx callResults and Transfer.txIdx (major)

jinmel:
## Summary

- Restructure `SimulationResult` so consumers can access per-transaction backend output (logs, status, returnData, gasUsed, Tenderly assetChanges) via `callResults: readonly SimulationCallResult[]`.
- Add required `txIdx: number` to every `Transfer`, indexing into `simulationTxs` so callers can attribute each parsed transfer to the emitting transaction.
- Remove the lossy top-level `SimulationResult.assetChanges` (it only carried the last bundle simulation's payload); per-tx `callResults[i].assetChanges` is the replacement.

## Why

Today, `simulate()` flattens raw logs from every transaction in a bundle into a single `RawLog[]` before parsing transfers. Consumers receive `transfers` without any way to map back to the originating tx. They also can't reach raw per-tx backend output (logs, return data, gas) without round-tripping the simulation themselves.

This PR exposes per-tx backend output as a structured public field and stamps `txIdx` onto every parsed transfer so the mapping is one lookup: `simulationTxs[transfer.txIdx]`.

## What changed

**Public API (breaking — major bump):**
- `SimulationResult.callResults: readonly SimulationCallResult[]` (new, required, 1:1 with `simulationTxs`).
- `Transfer.txIdx: number` (new, required).
- `SimulationResult.assetChanges` removed. Use `callResults[i].assetChanges`.
- `SimulationResult.simulationTxs` and `.transfers` are now `readonly` arrays.
- New public type exports: `RawLog`, `SimulationCallResult`.
- Migration notes in `.changeset/evm-simulation-callresults.md`.

**Internal:**
- `RawSimulationResult.logs` (flat) → `RawSimulationResult.callResults: RawCallResult[]` (per-tx).
- Both backends (`eth_simulateV1` via viem, Tenderly REST) populate `callResults` as they iterate per-tx.
- Tenderly Zod schema extended to extract `transaction.gas_used` and `transaction.transaction_info.call_trace.output`.
- `parseTransfers` now consumes `readonly RawCallResult[]` and stamps `txIdx` per emitting call.
- WETH9 wrap/unwrap dedup is now scoped to a single tx — eliminates a class of cross-tx data-equality false positives that the old flat dedup accepted.
- `screenAddresses` and `assertNoBundlerRetention` accept `readonly` arrays so the documented \"pass `result.simulationTxs` and `result.transfers`\" pattern typechecks against the now-readonly result.

**Spec / plan:**
- `docs/superpowers/specs/2026-05-05-evm-simulation-callresults-and-tx-mapping-design.md`
- `docs/superpowers/plans/2026-05-06-evm-simulation-callresults-and-tx-mapping.md`

## Test plan

- [ ] `pnpm --filter @morpho-org/evm-simulation test` — 203/203 passing locally.
- [ ] `pnpm --filter @morpho-org/evm-simulation exec tsc --noEmit` — clean.
- [ ] `pnpm --filter @morpho-org/evm-simulation lint` — clean.
- [ ] CI fork suite green per chain matrix.
- [ ] Manual sanity: integrator code that previously read `result.assetChanges` should switch to `result.callResults.at(-1)?.assetChanges`. Migration is a one-line change.
- [ ] New test cases verify behavior, not just shape:
  - `simulate.spec.ts` — multi-tx bundle round-trip: two callResults → two transfers with `txIdx: 0` and `txIdx: 1`.
  - `parsing/transfers.spec.ts` — txIdx attribution across multiple callResults; per-tx WETH9 dedup scope (cross-tx wrap/unwrap pairs are NOT deduped).
  - `backends/eth-simulate-v1.spec.ts` — per-tx `status`, `returnData`, `gasUsed`, `logs` assertions.
  - `backends/tenderly-rest.spec.ts` — per-tx `assetChanges` and per-step bundle logs.

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/605" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/evm-simulation/src/screen-addresses/screen-addresses.ts
```diff
@@ -27,50 +27,34 @@ const chainalysisEntitySchema = z
...
- *   chainalysisApiKey: process.env.CHAINALYSIS_API_KEY,
- * });
- * ```
+ * Throws `AddressScreeningError` on any Severe hit. Intended to be called after
+ * `simulate`, passing the returned `simulationTxs` and `transfers`.
  */
 export async function screenAddresses(params: {
-  simulationTxs: SimulationTransaction[];
-  transfers: Transfer[];
+  simulationTxs: readonly SimulationTransaction[];
```
jinmel: remove readonly

packages/evm-simulation/src/simulate/backends/tenderly-rest.ts
```diff
@@ -337,7 +335,7 @@ async function shareSimulation(params: {
     });
     if (!response.ok) {
       logger?.warn(
-        "Tenderly /share returned non-ok; tenderlyUrl will be cleared",
+        "Tenderly /share returned non-ok; tenderlyUrl will not be set",
```
jinmel: revert this changes which changed clears -> do not set

packages/evm-simulation/src/simulate/simulate.ts
```diff
@@ -84,7 +85,13 @@ export async function simulate(
     blockNumber: params.blockNumber,
     shareable,
   });
-  const transfers = parseTransfers(result.logs, config.logger);
+  if (result.calls.length !== simulationTxs.length) {
+    throw new ExternalServiceError(
+      `Backend returned ${result.calls.length} call result(s) for ${simulationTxs.length} transaction(s) — refusing to map transfers with mismatched lengths`,
+    );
```
0xbulma, line 97: **[LOW]** Per `packages/evm-simulation/AGENTS.md`, `ExternalServiceError` is the **bypassable** class — a consumer can choose to proceed past it. A backend returning the wrong number of call results is a contract violation, not a transient outage; proceeding would silently mis-attribute `txIdx`.

_Suggestion: throw a non-bypassable class — reuse `SimulationValidationError` or introduce `BackendContractError extends SimulationPackageError`. At minimum, document why `ExternalServiceError` is acceptable here._

packages/evm-simulation/src/simulate/backends/eth-simulate-v1.ts
```diff
@@ -96,18 +96,18 @@ export async function simulateV1(params: {
...
-    }
+    const rawCalls: RawCall[] = results.map((r) => ({
+      logs: (r.logs ?? []).map((log) => ({
+        address: log.address,
+        topics: [...log.topics],
+        data: log.data ?? "0x",
+      })),
+      status: r.status === "success",
+      returnData: r.data ?? "0x",
+      gasUsed: r.gasUsed,
```
0xbulma, line 107: **[LOW]** `gasUsed: r.gasUsed` is assigned without a fallback; the Tenderly path uses `BigInt(transaction.gas_used ?? 0)` (tenderly-rest.ts:397). If viem ever returns a result entry without `gasUsed`, this silently lands `undefined` despite the strict `bigint` typing of `RawCall.gasUsed`.

_Suggestion: default to `r.gasUsed ?? 0n` for parity with the Tenderly adapter, or narrow the viem return type so the fallback isn't needed._

packages/evm-simulation/src/simulate/backends/tenderly-rest.ts
```diff
@@ -46,6 +47,7 @@ const tenderlyRawResponseSchema = z
     transaction: z
       .object({
         error_message: z.string().optional(),
+        gas_used: z.number().optional(),
```
0xbulma: **[LOW]** `gas_used: z.number().optional()` combined with `?? 0` on L397 silently reports `gasUsed: 0n` if a successful Tenderly response omits `gas_used`. Successful sims always populate this field.

_Suggestion: tighten to `gas_used: z.number()` so a missing field surfaces as a Zod parse error rather than a fee-misleading `0n`._

packages/evm-simulation/src/simulate/backends/tenderly-rest.ts
```diff
@@ -65,6 +67,12 @@ const tenderlyRawResponseSchema = z
               )
               .optional(),
             asset_changes: z.unknown().optional(),
+            call_trace: z
+              .object({
+                output: hexSchema.optional(),
```
0xbulma: **[LOW]** `call_trace.output` is optional + falls back to `"0x"` on L396. A consumer reading `SimulationCall.returnData` for a real return value (e.g. a vault preview) would silently see `"0x"` if Tenderly drops `call_trace`.

_Suggestion: require both on success — `call_trace: z.object({ output: hexSchema })` — so absence is a hard parse failure._

packages/evm-simulation/src/simulate/backends/tenderly-rest.ts
```diff
@@ -224,10 +237,10 @@ async function simulateSingle(params: {
       signal,
       logger,
     });
-    if (!shared) result.tenderlyUrl = undefined;
+    if (!shared) tenderlyUrl = undefined;
   }
 
-  return result;
+  return { calls: [call], tenderlyUrl };
```
0xbulma, line 235: **[LOW]** `simulateSingle` (here) and `simulateBundle` (L294–313) duplicate the same `tenderlyUrl` resolution + `shareSimulation` block almost verbatim — only the `simulationId` source differs. The PR de-duplicated `parseTenderlyCall` but left this in place.

_Suggestion: extract a `resolveTenderlyUrl({ baseUrl, config, simulationId, shareable, signal, logger })` helper and call it from both paths._

packages/evm-simulation/src/simulate/backends/tenderly-rest.spec.ts
```diff
@@ -264,7 +269,7 @@ describe.sequential("simulateTenderlyRest — single tx", () => {
     expect(body.block_number).toBe("latest");
   });
 
-  it("returns assetChanges from transaction_info", async () => {
+  it("returns per-tx assetChanges from transaction_info", async () => {
```
0xbulma, line 284: **[LOW]** No test exercises the `gas_used ?? 0` fallback. Every `successBody()` fixture sets `gas_used: 21_000`.

_Suggestion: add a single-tx test where the response omits `gas_used` and assert `result.calls[0].gasUsed === 0n` — or, if the schema is tightened per the L50 comment, that it throws._

packages/evm-simulation/src/simulate/backends/tenderly-rest.spec.ts
```diff
@@ -308,7 +313,7 @@ describe.sequential("simulateTenderlyRest — bundle (multi-tx)", () => {
     expect(simulations[1]!.to).toBe(USDC);
   });
 
-  it("collects logs from every bundle step", async () => {
+  it("emits one call entry per bundle step with that step's logs", async () => {
```
0xbulma: **[LOW]** Bundle-mode "emits one call entry per bundle step" only asserts `logs` per step. A regression returning the same `gasUsed`/`returnData`/`assetChanges` for every entry would not be caught.

_Suggestion: extend with distinct `gas_used`/`call_trace.output`/`asset_changes` per fixture and assert each round-trips to the corresponding `result.calls[i]`._

packages/evm-simulation/src/simulate/parsing/transfers.ts
```diff
@@ -18,137 +23,146 @@ const UINT256_HEX_LENGTH = 66; // "0x" + 32 bytes
...
  * **Failure mode:** on a per-log parse failure (malformed topic length,
  * non-hex data), the log is skipped and a `warn` is emitted via the logger.
- * The caller receives a possibly-incomplete transfer list. Upstream sources
- * (e.g., Tenderly response schemas) should validate log shape before calling
- * here to avoid silent drops.
  *
  * Output is sorted canonically by token, from, to, amount for determinism.
+ * `txIdx` is attached but does not influence sort order.
  */
 export function parseTransfers(
```
0xbulma, line 42: **[LOW]** Rewritten `parseTransfers` JSDoc is missing `@param calls`, `@param logger`, and `@returns`. CLAUDE.md §6 requires these on exported functions; §9 says adopt the convention on refactor.

_Suggestion: add the missing tags. The old caveat "Upstream sources should validate log shape before calling here" was also dropped — restore if still relevant._

packages/evm-simulation/src/simulate/simulate.spec.ts
```diff
@@ -118,6 +130,64 @@ describe.sequential("simulate — success", () => {
...
+    );
+
+    expect(result.calls).toHaveLength(2);
+    const approveTransfer = result.transfers.find(
+      (t) => t.amount === APPROVE_AMOUNT,
+    );
+    const mainTransfer = result.transfers.find(
+      (t) => t.amount === TRANSFER_AMOUNT,
+    );
+    expect(approveTransfer?.txIdx).toBe(0);
```
0xbulma, line 199: **[LOW]** This multi-tx test infers the calls↔txs ordering only via `txIdx`, conflating two contracts (txIdx round-trip + 1:1 ordering of `calls`).

_Suggestion: pin the ordering directly — e.g., `expect(result.calls[0].logs[0].topics[0]).toBe(TRANSFER_TOPIC)` and assert each `calls[i].logs[0]` is the transfer emitted by tx `i`._

packages/evm-simulation/src/types.ts
```diff
@@ -134,16 +144,56 @@ export interface SimulateParams {
...
+  calls: RawCall[];
   tenderlyUrl?: string;
-  rawAssetChanges?: unknown;
 }
 
 export interface RawLog {
-  address: Address;
-  topics: Hex[];
-  data: Hex;
+  readonly address: Address;
```
0xbulma, line 162: **[HIGH]** `RawLog` is now publicly re-exported from `src/index.ts` (per the changeset) but carries zero JSDoc. CLAUDE.md §6 requires JSDoc on every exported symbol.

_Suggestion: add a short description (and ideally `@example`) explaining it's the normalized log shape — `address`, indexed `topics`, hex `data` — emitted by a single call and consumed by `parseTransfers`._

packages/evm-simulation/src/types.ts
```diff
@@ -134,16 +144,56 @@ export interface SimulateParams {
...
+  readonly topics: readonly Hex[];
+  readonly data: Hex;
+}
+
+/**
+ * Internal mirror of `SimulationCall`, mutable during construction by
+ * the simulation backends.
+ */
+export interface RawCall {
```
0xbulma, line 173: **[LOW]** `RawCall` is described as an internal mirror but is missing the explicit `@internal` JSDoc tag. CLAUDE.md §3: "Internal symbols carry `@internal` JSDoc and do not participate in the stability contract."

_Suggestion: add `@internal` to the JSDoc block so it's excluded from TypeDoc and the stability contract._

packages/evm-simulation/src/types.ts
```diff
@@ -134,16 +144,56 @@ export interface SimulateParams {
...
+
+/**
+ * Per-transaction normalized output from the simulation backend.
+ *
+ * `SimulationResult.calls[i]` corresponds 1:1 with
+ * `SimulationResult.simulationTxs[i]`. Use this to read raw logs, status,
+ * return data, gas used, and (Tenderly only) asset changes per transaction.
+ */
+export interface SimulationCall {
+  readonly logs: readonly RawLog[];
```
0xbulma, line 190: **[LOW]** `SimulationCall.logs` is the only field without per-field JSDoc; siblings (`status`, `returnData`, `gasUsed`, `assetChanges`) all carry one.

_Suggestion: add a one-liner like "Raw EVM logs emitted by this call." for consistency._

0xbulma (commented):
Active findings after self-review: **1 critical, 1 high, 1 medium, 7 lows** posted as line-anchored review comments below.

packages/evm-simulation/src/simulate/simulate.ts
```diff
@@ -84,7 +85,13 @@ export async function simulate(
     blockNumber: params.blockNumber,
     shareable,
   });
-  const transfers = parseTransfers(result.logs, config.logger);
+  if (result.calls.length !== simulationTxs.length) {
```
0xbulma, line 94: **[CRITICAL] Re: L43 (`@returns` JSDoc above this function)** — stale `@returns` still says "the opaque Tenderly `assetChanges` payload", but `assetChanges` was removed from `SimulationResult` in this PR (now per-tx on `calls[i].assetChanges`).

_Suggestion: rewrite to mention `simulationTxs`, `calls` (with per-tx `logs`/`status`/`returnData`/`gasUsed`/`assetChanges`), `transfers`, and optional `tenderlyUrl`._

0xbulma, line 94: **[MEDIUM] Re: L48 (`@throws` JSDoc above this function)** — `@throws {ExternalServiceError}` description ("when both backends are unavailable within the timeout budget") only covers the legacy trigger. The new guard right here at L88–91 throws the same class on `result.calls.length !== simulationTxs.length`. Integrators pattern-matching on the class won't know about the second reason.

_Suggestion: extend the description (or add a second `@throws` line) so both triggers are documented._

packages/evm-simulation/src/simulate/backends/tenderly-rest.ts
```diff
@@ -387,14 +390,11 @@ function parseTenderlyResponse(
...
-  const tenderlyUrl =
-    shareable && simulation.id
-      ? `https://dashboard.tenderly.co/shared/simulation/${simulation.id}`
-      : undefined;
-
   return {
     logs: rawLogs,
-    tenderlyUrl,
-    rawAssetChanges: transaction.transaction_info.asset_changes,
+    status: true,
```
0xbulma, line 409: **[LOW] Re: L384 (`if (log.raw) { ... }` earlier in this function)** — the `if (log.raw)` check silently drops Tenderly logs missing the `.raw` envelope; no `logger?.warn` is emitted, so a backend regression that strips `raw` would silently produce empty transfer lists.

_Suggestion: tighten the schema (require `raw`) or emit `logger?.warn("Tenderly log missing .raw envelope", { ... })` in the else branch._

packages/evm-simulation/src/simulate/simulate.ts
0xbulma: Withdrawing this — on reflection, the PR follows the existing `ExternalServiceError` convention for backend issues, and whether that class should be bypassable is a broader design question for the package, not specific to this PR. Marking resolved.

packages/evm-simulation/src/simulate/backends/eth-simulate-v1.ts
0xbulma: Withdrawing — viem's `simulateCalls` types `gasUsed` as required `bigint` (the missing fallback here mirrors that, while `r.data ?? "0x"` exists because `data` is optional in viem's type). The type system already guarantees this. Resolving.

packages/evm-simulation/src/simulate/parsing/transfers.ts
0xbulma: Withdrawing — `parseTransfers` is not re-exported from `src/index.ts` and isn't part of the package's public API, so CLAUDE.md §6 ("every exported symbol") doesn't strictly apply. Convention overreach for an internal function. Resolving.

packages/evm-simulation/src/simulate/backends/tenderly-rest.ts
0xbulma: Withdrawing — the `if (log.raw)` block sits between two diff hunks and wasn't modified by this PR. Pre-existing concern, not PR-introduced. Resolving.

packages/evm-simulation/src/types.ts
0xbulma: Withdrawing — stylistic consistency nit only, not a defect. Resolving.

jinmel:
Superseded by #635, which lives on the main-repo branch (`morpho-org:feat/evm-simulation-callresults-and-tx-mapping`) and points to the same HEAD. All review feedback addressed there.

PR #651, opened 2026-05-11
ci(release): publish from push-triggered release PRs

Rubilmax:
## Summary

- keep release automation push-triggered: every push runs lint, build, and tests, while `main`/`next` pushes create or refresh `changeset-release/<branch>` when pending changesets exist
- publish only after the generated release PR is merged, then push package tags and create one GitHub Release per published package from that package's generated changelog section
- configure Changesets GitHub changelog generation and add the GitHub token permissions needed to resolve PR/author metadata during `pnpm run version`
- clarify that changesets are only for semver-relevant published package source changes, add that check to the shared PR review base, and remove the `/release` command docs
- document the push-based release lifecycle in `docs/tibs/TIB-2026-05-12-release-pr-publish-on-push.md`

## Validation

- `git diff --check`
- `ruby -e 'require "yaml"; Dir[".github/workflows/*.{yml,yaml}"].sort.each { |path| YAML.load_file(path); puts "ok #{path}" }'`
- isolated worktree `pnpm run version` check to confirm multi-package changesets generate duplicated per-package changelog sections and that publish tag matching maps each tag to one package

Tests not run; changed files are release workflow/configuration and documentation.

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1778509637129949)


.github/workflows/release.yml
```diff
@@ -0,0 +1,250 @@
...
+        required: true
+        default: main
+        type: choice
+        options:
+          - main
+          - next
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'workflow_dispatch' && inputs.branch || github.ref }}-${{ github.event_name }}
+  cancel-in-progress: true
```
Foulks-Plb: `release.yml` sets `cancel-in-progress: true`, while `publish.yml` and `version-pr.yml` (correctly) set `cancel-in-progress: false`. The inner `false` doesn't protect against the parent: reusable workflows invoked via `uses: ./.github/workflows/X.yml` run as **jobs inside the same run** as `release.yml`, not as separate runs. When the parent run is cancelled by concurrency, every job in it, including jobs coming from `publish.yml`: is killed with it. 

A reusable workflow's own `concurrency` block only applies when it's triggered standalone, not via `workflow_call`.

Foulks-Plb:
The PR title and description no longer reflect the actual changes.

Foulks-Plb (dismissed):

Foulks-Plb:
CI is failing

Foulks-Plb (dismissed):

AGENTS.md
```diff
@@ -140,8 +140,8 @@ A scannable list of patterns reviewers reject. Most are review-only today (per t
 
 ## 7. Releases & versioning
 
-- **SemVer, strict.** Patch: bug fixes, internal changes, doc-only. Minor: additive surface, deprecations. Major: removed/renamed/retyped public symbols.
-- **CHANGELOG via Changesets** — every user-visible change ships with a changeset (review-only today).
+- **SemVer, strict.** Patch: bug fixes and internal source maintenance changes. Minor: additive surface, deprecations. Major: removed/renamed/retyped public symbols.
+- **CHANGELOG via Changesets** — every semver-relevant change to published package source ships with a changeset (review-only today). Do not generate a changeset for repo metadata, documentation-only, fixture-only, generated-output-only, or tests-only changes; those do not change the published package contract.
```
0xbulma: documentation only can trigger a update no ? This will then bump the developers doc website.

.github/workflows/publish.yml
```diff
@@ -120,3 +121,100 @@ jobs:
...
+
+          prerelease_args=()
+          if [ "${TARGET_BRANCH}" = "next" ]; then
+            prerelease_args=(--prerelease)
+          fi
+
+          for tag in "${tags[@]}"; do
+            body_file="${releases_dir}/${tag//\//__}.md"
+
+            node --input-type=module - "${tag}" "${body_file}" <<'NODE'
```
tarikbellamine: **P1 — Extract this ~80-line node script (lines 147–213) into a real source file.**

This is release-critical logic embedded in YAML:
- No tests, no lint, no biome pass, no type check.
- Reviewers can't easily verify changes — the diff lives inside a heredoc inside a `run: |` block.
- The matching logic has subtle behavior (longest-prefix sort, dual separators `@`/`-v`, heading regex anchoring, section-end search, regex escaping) that's exactly the kind of code that benefits from unit tests.
- Any change to `@changesets/changelog-github`'s heading format silently produces wrong release notes — that risk should be caught by a test pinned to a sample CHANGELOG fixture, not by a release running in prod.

Recommendation: move to `scripts/release/github-release-body.mjs` (or similar). Benefits:
- Lands under `/scripts/` CODEOWNERS.
- Biome lints it, tsc/vitest can cover it.
- Test cases worth pinning: scoped names, prefix collisions (`blue-sdk` vs `blue-sdk-viem`), empty/missing `packageJson.name`, missing CHANGELOG, missing version section, tag without `@` separator, version containing regex metacharacters, multiple `## ` headings in a single section.

The YAML step then becomes `run: node scripts/release/github-release-body.mjs "$tag" "$body_file"`.

.github/workflows/version-pr.yml
```diff
@@ -67,18 +69,20 @@ jobs:
...
+        if: env.TARGET_BRANCH == 'main' && hashFiles('.changeset/pre.json') != ''
         run: |
           echo "::error::.changeset/pre.json is present on main. Run pnpm changeset pre exit and commit the result before stable releases."
           exit 1
 
       - name: Version packages
         if: steps.pending-changesets.outputs.has_changesets == 'true'
         run: pnpm run version
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
```
tarikbellamine: **P1 — `GITHUB_TOKEN` is now exposed to the entire `pnpm run version` dependency graph.**

This token is reachable via `process.env.GITHUB_TOKEN` from anything `@changesets/cli` loads, including the new `@changesets/changelog-github` and its transitive set (`@changesets/get-github-info`, `node-fetch@2.7.0`, `dataloader@1.4.0`, `dotenv@8.6.0`).

**Today's blast radius is small but non-zero:**
- Workflow-level scope here is `contents: read` + `pull-requests: read` on a public repo — exfiltration mostly gets you read access to data that's already public.
- pnpm disables transitive lifecycle scripts by default, so a compromised dep can't run `postinstall` to grab the token; it'd need to be loaded by `changesets version`.
- `minimum-release-age=4320` gives a 3-day cooldown on consuming newer upstream versions.

**The risk is forward-looking:** if a future workflow change widens this token (e.g., to write CHANGELOG via the same token, or to push tags), the entire changesets dependency graph inherits that trust silently.

Recommendations:
- Add a comment on this env block explicitly stating the trust assumption ("any package reachable from `pnpm run version` can read this token; do not widen scope without re-auditing the dep graph").
- Or replace `github.token` with a fine-grained PAT scoped to `pull_requests: read` on this single repo, so even if the token leaks it can't grow.
- Long term: consider running `pnpm run version` under a sandboxed Node permission model (`--permission` / `--allow-fs-read`) if the changelog plugin tolerates it.

.github/workflows/publish.yml
```diff
@@ -120,3 +121,100 @@ jobs:
 
           git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
           git push --atomic origin "${refs[@]}"
+
+      # This only creates releases for tags made by this publish run. If npm and
+      # tag push succeed but release creation fails, reruns need manual repair.
```
tarikbellamine: **P1 — Partial-release recovery is silently manual across three failure boundaries.**

The comment correctly identifies the most obvious case but the recovery story is worse than it reads:

1. **npm succeeds → tag push fails:** safe. Re-run is idempotent — `changeset publish` skips already-published versions, `git push --atomic` retries cleanly.
2. **npm + tag push succeed → GitHub release loop fails partway:** with `set -euo pipefail` and `for tag in "${tags[@]}"`, the first failure aborts the loop. Remaining tags never get releases.
3. **Re-running step #2 doesn't fix it:** `has_tags` is gated on `comm -13 "$before_tags" "$after_tags"` from *this run only*. On rerun, no new tags → `Create GitHub releases` step is `if`-skipped entirely. The stuck state requires a maintainer to hand-create the missing releases.

Recommendation: replace the bash `for tag` loop with a fail-at-end pattern — accumulate per-tag failures with `|| failures+=("$tag")`, continue past errors, then `exit 1` if any failed (with the list logged). Or drive release creation from "tags on remote that don't have a GH release yet" instead of "tags created by this run" — that recovers automatically on rerun.

Even simpler interim mitigation: move the `Create GitHub releases` step into its own job with `if: always() && needs.publish.result == 'success'` and `continue-on-error` on the underlying script. Less elegant but immediately re-runnable.

.github/workflows/publish.yml
```diff
@@ -120,3 +121,100 @@ jobs:
...
+              const prefix = `${candidate.name}${separator}`;
+              if (tag.startsWith(prefix)) {
+                matchedPackage = candidate;
+                version = tag.slice(prefix.length);
+                break;
+              }
+            }
+
+            if (matchedPackage != null) break;
+          }
```
tarikbellamine: **P1 — Tag→package mapping heuristic has correctness gaps that can produce wrong releases.**

```js
packages.sort((left, right) => right.name.length - left.name.length);

let matchedPackage;
let version;
for (const candidate of packages) {
  for (const separator of ["@", "-v"]) {
    const prefix = `${candidate.name}${separator}`;
    if (tag.startsWith(prefix)) {
      matchedPackage = candidate;
      version = tag.slice(prefix.length);
      break;
    }
  }
  if (matchedPackage != null) break;
}
```

Issues:

1. **Unstable sort:** the comparator returns 0 for same-length names. Two packages whose names happen to be the same length get non-deterministic mapping order across runs. Add a secondary key by name string for determinism.
2. **Empty / missing `packageJson.name` matches everything:** if any package.json under `packages/*/` has an empty `name`, `prefix` becomes `"@"` (or `"-v"`) and matches every scoped tag (or every tag containing `-v`). Add `if (!packageJson.name) continue;` before pushing into `packages`.
3. **First-match-wins is silent:** there's no assertion that exactly one package matches. If the heuristic ever picks the wrong package due to a future name collision, the wrong CHANGELOG gets used as release notes and nobody notices until someone reads the release.

Deeper recommendation: don't reconstruct the mapping at all. `@changesets/get-release-plan` (already in the dep tree) produces the authoritative `tag → { name, version }` mapping that `changeset publish` itself uses. Reading the release plan eliminates the entire prefix-matching heuristic and inherits whatever future tag-format changes changesets makes.

If staying with prefix matching, at minimum: skip empty names, add a stable secondary sort key, and assert that no two packages produce a valid match for the same tag (fail loudly on ambiguity).

tarikbellamine (commented):
Security-focused review

Rubilmax:
@codex

0xbulma (dismissed):

Foulks-Plb (dismissed):

tarikbellamine (commented):
Second-pass security review covering commits since `6f83e0e` (de855b3..06c9a2c).

### Confirmed fixed from my prior review

- `GITHUB_TOKEN` for `pnpm run version` is now a fine-grained read-only App token (`version-pr.yml:63-80`) — addresses the dep-graph trust concern.
- Tag→package mapping rewritten as exact-string equality with unit tests covering prefix collisions, regex metacharacters, empty names, missing changelog/version section — addresses the heuristic-correctness concern.
- Release-notes logic extracted to `scripts/release/` and covered by vitest — addresses the YAML-embedded-script concern.
- GH-release loop made idempotent via `gh release view → edit else create` — addresses the partial-loop-failure concern (but see H1, the published-but-not-tagged case is still stuck).

### Findings without a diff-anchored line (line targets are outside this PR's hunks)

**M — `changeset-release/<branch>` branch protection.** The `--force-with-lease` design at `version-pr.yml:168-174` assumes only the workflow's App actor pushes to `changeset-release/main` and `changeset-release/next`. If branch protection isn't configured to restrict pushes to those refs, a contributor with repo write can land content there, which then feeds into the release-PR body and (via `scripts/release/github-release-body.mjs`) becomes the npm + GitHub Release notes verbatim. Configure protection to allow only the `VERSION_APP` installation actor.

**M — Stale-edit footgun on the release PR.** The force-with-lease push at `version-pr.yml:171` rebuilds `HEAD` from `github.sha` (main tip), not from the existing release branch. Manual edits a reviewer makes directly to the `chore: version packages` PR — e.g. fixing a typo in a generated `CHANGELOG.md` — are silently overwritten by the next feature push to main. By design (per TIB-2026-05-12), but worth surfacing in the PR-body heredoc at `version-pr.yml:188-193` so reviewers don't sink time into ephemeral edits.

Remaining H/M/L findings inline.

.github/workflows/publish.yml
```diff
@@ -120,3 +106,65 @@ jobs:
 
           git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
           git push --atomic origin "${refs[@]}"
+
+      - name: Upload package tag list
+        if: steps.publish-packages.outputs.has_tags == 'true'
```
tarikbellamine, line 167: **H — Partial-publish recovery is still broken when tag push fails.**

The artifact upload is gated on `has_tags == 'true'`, which `publish-packages` only sets when *this run* created new local tags. Failure mode:

1. `pnpm release` publishes to npm and creates local tags → success.
2. `git push --atomic` (line 108) fails (network blip, token hiccup).
3. `set -euo pipefail` aborts the job before this step runs.
4. `github-releases` is skipped because `needs.publish.result == 'success'` is false.

On rerun: `pnpm release` is idempotent — it sees those versions already on npm, creates no new local tags → `has_tags=false` → this step skips, `github-releases` skips. The published-but-untagged state is stuck and needs a maintainer to manually `git push refs/tags/...` and `gh release create`.

Fix options:
- Detect "tags not yet on remote" instead of "tags created this run" — `comm` between `git tag --list` and `git ls-remote --tags origin`.
- Move tag push + artifact upload into a separate job gated on `pnpm release` exit code, not on the local-tag diff.

Same failure mode I raised on the previous review pass; the `gh release view → edit` change fixed it for the GH-Releases loop, but the upstream tag-push gap remains.

.github/workflows/version-pr.yml
```diff
@@ -43,44 +44,42 @@ jobs:
...
         run: |
           echo "::error::.changeset/pre.json is present on main. Run pnpm changeset pre exit and commit the result before stable releases."
           exit 1
 
+      - name: Generate Changelog GitHub App Token
+        if: steps.pending-changesets.outputs.has_changesets == 'true'
+        id: changelog-token
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3
+        with:
+          app-id: ${{ secrets.VERSION_APP_ID }}
```
tarikbellamine, line 64: **H (config, out of band) — verify VERSION_APP install scope.**

This App ends up holding `contents:write` + `pull-requests:write` once the second token is minted at line 85-90. If the private key ever leaks via a future workflow change that exfiltrates it on `main` or `next`, blast radius is bounded by the App's installation scope. Worth confirming with whoever administers the App: install on `morpho-org/sdks` only, not org-wide. Not visible from the diff; flagging here so it's on the merge checklist.

scripts/release/detect-pending-changesets.mjs
```diff
@@ -0,0 +1,55 @@
...
+  const pendingChangesets = listPendingChangesets(options);
+  const outputFile = options.outputFile ?? process.env.GITHUB_OUTPUT;
+
+  if (outputFile != null && outputFile !== "") {
+    appendFileSync(outputFile, getGitHubOutput(pendingChangesets));
+  }
+
+  if (pendingChangesets.length > 0) {
+    process.stdout.write("Pending changesets:\n");
+    process.stdout.write(`${pendingChangesets.join("\n")}\n`);
```
tarikbellamine: **M — Workflow-command injection via `.changeset/` filenames.**

Filenames are echoed verbatim. Linux filenames can contain embedded newlines, and GitHub Actions interprets log lines starting with `::cmd::value` as workflow commands. A `.changeset/` filename containing `\n::stop-commands::<token>` or `\n::add-mask::<secret>` would, when echoed here, inject a workflow command.

Narrow vector — landing a file with a newline in its basename through git is unusual and would stand out in review — but the sanitization is cheap:

```js
const safe = (name) => name.replace(/[\r\n\t\u0000-\u001f]/g, "?");
process.stdout.write(`${pendingChangesets.map(safe).join("\n")}\n`);
```

The line written to `$GITHUB_OUTPUT` is a literal boolean, not affected. This only concerns the human-facing log.

.github/workflows/publish.yml
```diff
@@ -120,3 +106,65 @@ jobs:
...
+        with:
+          name: new-package-tags
+          path: ${{ runner.temp }}/new-package-tags.txt
+          if-no-files-found: error
+
+  github-releases:
+    if: always() && needs.publish.result == 'success' && needs.publish.outputs.has_tags == 'true'
+    name: GitHub releases
+    needs: publish
+    runs-on: ubuntu-latest
```
tarikbellamine, line 178: **L — `github-releases` job inherits unneeded `id-token: write`.**

The workflow-level `permissions:` block grants `id-token: write` for npm trusted publishing in the `publish` job. This job doesn't talk to npm — it only creates GitHub Releases. Add an explicit job-level `permissions:` block here scoping to `contents: write` only. Defense in depth: reduces blast radius if a step in this job ever processes user-influenced input.

.github/workflows/publish.yml
```diff
@@ -120,3 +106,65 @@ jobs:
 
           git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```
tarikbellamine, line 162: **L — Credentialed remote URL persists for the job lifetime.**

`git remote set-url origin "https://x-access-token:${GH_TOKEN}@..."` writes the token into `.git/config`. The filesystem is ephemeral so it doesn't leak past the job, but a future debug `set -x` or a `git remote -v` in a later step would expose it. Add a final clearing step:

```bash
git remote set-url origin "https://github.com/${GITHUB_REPOSITORY}.git"
```

Same pattern at `version-pr.yml:103` deserves the same treatment.

.github/workflows/version-pr.yml
```diff
@@ -12,7 +12,7 @@ permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-version-pr
+  group: ${{ github.workflow }}-${{ github.ref_name }}-version-pr
```
tarikbellamine: **L — Dead `concurrency` block in a reusable workflow.**

`concurrency:` on workflows invoked via `workflow_call` is ignored — the parent `push.yml` block is what actually serializes runs. `push.yml` correctly sets `cancel-in-progress: false` for `main`/`next`, so the intended serialization is achieved, but this block here is dead code. Same applies to `publish.yml:10-12`. @Foulks-Plb flagged this on `0b059e163d`; worth removing both for clarity.

.github/workflows/version-pr.yml
```diff
@@ -105,6 +104,7 @@ jobs:
 
           is_allowed_version_path() {
             [[ "$1" =~ ^packages/[^/]+/package\.json$ ]] ||
```
tarikbellamine, line 108: **L (forward-compat) — Allowlist excludes nested-package layouts.**

`^packages/[^/]+/package\.json$` and `^packages/[^/]+/CHANGELOG\.md$` only match exactly one path segment under `packages/`. Not a security issue today (no nested packages), but a sharp edge: if a future package lives at `packages/foo/bar/package.json`, `changeset version`'s legitimate write would be rejected by the second allowlist check (lines 136-140) and fail the release. Either tighten to match `packages/**/{package.json,CHANGELOG.md}`, or document the single-level assumption next to this regex.

.github/workflows/version-pr.yml
Rubilmax: Confirmed

.github/workflows/version-pr.yml
Rubilmax: Should not exist

Rubilmax:
Rulesets implemented on `changeset-release/main` and `changeset-release/next`

Foulks-Plb (approved):

tarikbellamine (approved):

PR #646, opened 2026-05-10
refactor(morpho-sdk): pass generalAdapter1 explicitly to getRequirementsAction

Foulks-Plb:
## Summary

`getRequirementsAction` no longer hard-codes the transfer destination via `getChainAddresses(chainId).bundler3.generalAdapter1`. It now takes an explicit `recipient: Address` parameter, and every in-repo caller passes the resolved address as `recipient: generalAdapter1`.

- `chainId` is removed from the helper's params (it was only used to resolve the destination address).
- Updated callers: `marketV1` (`supplyCollateral`, `supplyCollateralBorrow`, `repay`, `repayWithdrawCollateral`), `vaultV1` (`deposit`, `migrateToV2`), `vaultV2` (`deposit`).
- JSDoc rewritten to describe `recipient` generically — no more `GeneralAdapter1` lock-in in the helper's contract.
- The helper is `@internal`, so this is **not** a public-API break. Behaviour is unchanged: the same address still receives the transferred asset.

### Naming note

The SDK TS layer (`bundler-sdk-viem`) already uses `recipient` for both the `transferFrom2` and `erc20TransferFrom` action tuples, so `recipient` is consistent with the SDK's vocabulary. On-chain ABIs are split (`transferFrom2` → `recipient`, `erc20TransferFrom` → `receiver`) — out of scope for this PR.

## Test plan

- [x] `pnpm exec tsc --noEmit` in `packages/morpho-sdk` (clean locally)
- [ ] CI: fork-based integration tests on each updated builder
- [ ] Verify snapshot transactions on the affected paths are unchanged (no calldata change expected)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1778484584979589)

PR #654, opened 2026-05-11
refactor(morpho-sdk): pass generalAdapter1 explicitly to getRequirementsAction

Foulks-Plb:
## Summary

`getRequirementsAction` no longer hard-codes the transfer destination via `getChainAddresses(chainId).bundler3.generalAdapter1`. It now takes an explicit `recipient: Address` parameter, and every in-repo caller passes the resolved address as `recipient: generalAdapter1`.

- `chainId` is removed from the helper's params (it was only used to resolve the destination address).
- Updated callers: `marketV1` (`supplyCollateral`, `supplyCollateralBorrow`, `repay`, `repayWithdrawCollateral`), `vaultV1` (`deposit`, `migrateToV2`), `vaultV2` (`deposit`).
- JSDoc rewritten to describe `recipient` generically — no more `GeneralAdapter1` lock-in in the helper's contract.
- The helper is `@internal`, so this is **not** a public-API break. Behaviour is unchanged: the same address still receives the transferred asset.

### Naming note

The SDK TS layer (`bundler-sdk-viem`) already uses `recipient` for both the `transferFrom2` and `erc20TransferFrom` action tuples, so `recipient` is consistent with the SDK's vocabulary. On-chain ABIs are split (`transferFrom2` → `recipient`, `erc20TransferFrom` → `receiver`) — out of scope for this PR.

Replaces #646 (resigned commits).

## Test plan

- [x] `pnpm exec tsc --noEmit` in `packages/morpho-sdk` (clean locally)
- [ ] CI: fork-based integration tests on each updated builder
- [ ] Verify snapshot transactions on the affected paths are unchanged (no calldata change expected)

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1778484584979589)

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/654" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/get-requirements-action-general-adapter-param.md
+++ b/.changeset/get-requirements-action-general-adapter-param.md
@@ -3,9 +3,9 @@
 ---
 
 Internal: `getRequirementsAction` now takes the transfer recipient as an
-explicit `generalAdapter1` parameter instead of resolving it from `chainId`.
-The function is `@internal` and not part of the public surface; all in-repo
+explicit `recipient` parameter instead of resolving it from `chainId`. The
+function is `@internal` and not part of the public surface; all in-repo
 callers (`marketV1` supply/repay paths, `vaultV1`/`vaultV2` deposit, and
-`vaultV1` migrate-to-v2) have been updated to pass `generalAdapter1`
+`vaultV1` migrate-to-v2) have been updated to pass `recipient: generalAdapter1`
 directly. No behavior change — same destination address, just no longer
 hard-coded inside the helper.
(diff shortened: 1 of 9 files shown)


.changeset/get-requirements-action-general-adapter-param.md
```diff
@@ -0,0 +1,11 @@
+---
+"@morpho-org/morpho-sdk": patch
```
Rubilmax, line 2: Technically this would require a "major" version bump since it's a breaking change. I guess in the early stage of a package life, it's fine to bump this only as a "minor". But "patch" is probably an issue here since `getRequirementsAction` can be imported by sdk consumers?

Rubilmax (approved):

PR #655, opened 2026-05-12
refactor(morpho-sdk): extract public reallocation data

Rubilmax:
## What changed

- Added a dedicated `ReallocationData` state container for public allocator reallocation computation.
- Moved shared-liquidity option and input types into morpho-sdk, including explicit timestamp-driven computation.
- Updated MarketV1 reallocation data fetching to return `ReallocationData` instead of `SimulationState`.
- Added parity coverage for the reallocation algorithm and updated reallocation action tests to pass the fetched block timestamp.
- Optimized reallocation simulation to clone only mutated state entries per applied public reallocation.
- Added regression coverage for clone preservation and disabled allocator short-circuit behavior.
- Added a root convention to prefer `_try(accessor, UnknownError)` for optional domain lookups.

Fixes SDK-185

## Why

This decouples morpho-sdk reallocation computation from simulation-sdk block state while keeping timestamp-sensitive accrual and pending-cap behavior explicit at the call site. The partial-clone path also avoids repeatedly deep-cloning unrelated market, vault, position, and vault-market-config state during iterative public reallocation computation.

## Performance

Local comparison while reviewing the copy strategy measured:

- Partial-clone `ReallocationData`: `21.23ms`
- Full-clone `ReallocationData` reference: `97.60ms` (`4.60x` slower than partial clone)
- `SimulationState` / immer implementation: `76.28ms` (`3.59x` slower than partial clone)

The benchmark scaffolding was removed from the committed tests; the measurements are reported here for the PR discussion.

## Validation

- `pnpm exec biome check packages/morpho-sdk/src/helpers/reallocationData.ts packages/morpho-sdk/src/helpers/reallocationData.test.ts`
- `pnpm exec vitest run packages/morpho-sdk/src/helpers/reallocationData.test.ts --config vitest.config.ts --project morpho-sdk`
- `pnpm --filter @morpho-org/morpho-sdk exec tsc --noEmit`
- `git diff --check`
- Pre-commit hook: `biome check --write --no-errors-on-unmatched && lint:address`

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1778686734020289)


Rubilmax (commented):
LGTM

packages/morpho-sdk/src/helpers/computeReallocations.ts
```diff
@@ -28,14 +28,16 @@ export const computeReallocations = ({
...
   readonly marketId: MarketId;
   readonly borrowAmount: bigint;
   readonly options?: ReallocationComputeOptions;
 }): readonly VaultReallocation[] => {
   if (options?.enabled === false) return [];
 
-  const market = data.getMarket(marketId).accrueInterest(data.block.timestamp);
+  // ReallocationData does not retain the fetch block; pass that block timestamp
+  // to compute against the same accrued state, otherwise Market defaults to lastUpdate.
+  const market = data.getMarket(marketId).accrueInterest(options?.timestamp);
```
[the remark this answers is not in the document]
Rubilmax: Accruing only increases liquidity, so a bundle calculated in the past cannot fail onchain. Plus, it's not really an issue: if someone wants to calculate reallocations at some point in time without virtually accruing (using onchain state only), they are free to do so with the current state of the code

Rubilmax (commented):
LGTM

Foulks-Plb:
@codex review

Foulks-Plb:
I got this flagged by an agentic review:

> 🟢 Observation #4 — holdings fetched but never consumed
> 
> marketV1.ts:1016-1020 calls fetchHolding for each vaultAddress on the loanToken. These holdings are stored in ReallocationData.holdings, but the review claims nothing in getMarketPublicReallocations / applyPublicReallocation actually reads them. According to the review, only markets, vaults, positions, and vaultMarketConfigs are consumed by the algorithm.
> 
> So the conclusion was:
> → potentially N unnecessary RPC calls per fetch. Either remove the fetch or remove holdings from InputReallocationData. Not a bug, just extra cost.

But I believe holdings are actually used in the algorithm — could you double check?

Foulks-Plb:
Otherwise, this looks very solid.

packages/morpho-sdk/src/helpers/computeReallocations.ts
```diff
@@ -30,14 +30,16 @@ export const computeReallocations = ({
...
```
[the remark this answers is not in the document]
Rubilmax: Duplicate

Foulks-Plb (dismissed):

0xbulma:
@carapulse review

0xbulma:
@prd-carapulse[bot] review this PR deeply. Make sure it introduces no regressions in scope

Rubilmax:
> * `reallocationData.ts:449-484` can select a source market with `VaultMarketConfig.enabled === false`; `applyPublicReallocation()` also never checks source/target `enabled`. Onchain/simulation rejects this in `PublicAllocator.reallocateTo`, so SDK can return reverting reallocations.

The algorithm only checks markets that are in the withdraw queue of a vault, so the market is implicitely enabled. Double-check this @prd-carapulse[bot]

> * `delay` behavior was removed. Old path accrued target cap headroom at `block.timestamp + delay` while source withdrawals used `block.timestamp`; new single `timestamp` cannot preserve that. Public API/behavior regression for `morpho-sdk` and `liquidity-sdk-viem`.

Judged useless: shared liquidity is at worst under-estimated (conservative).

> * `getReallocationData()` requires `block.timestamp` but `ReallocationData` drops it; default `getReallocations()` falls back to `market.lastUpdate`. Existing thread raised this; I still consider it unresolved for cap/maxIn/delay consistency.

Same reason

packages/liquidity-sdk-viem/src/loader.ts
```diff
@@ -231,6 +188,7 @@ export class LiquidityLoader<chain extends Chain = Chain> {
             const { data: endState, withdrawals } =
               startState.getMarketPublicReallocations(uniqueKey, {
                 ...parameters,
+                timestamp: block.timestamp,
```
0xbulma: **[MEDIUM]** No new test asserts that `block.timestamp` is forwarded to `getMarketPublicReallocations`. The existing nock-mocked test compares the final `withdrawals` array to a fixture, which indirectly covers the new path, but a timestamp-plumbing regression would not be caught if the fixture happens to produce identical withdrawals at the default timestamp.

Suggested: add an assertion (or focused test) using a fixture whose withdrawal amount depends on accrued interest — compare results at two timestamps and assert they differ — to lock the timestamp plumbing.

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -911,13 +913,25 @@ export class MorphoMarketV1 implements MarketV1Actions {
...
   }
 
+  /**
+   * Fetches all on-chain inputs needed to compute public allocator reallocations.
+   *
+   * @param params - Reallocation data fetch parameters.
+   * @param params.vaultAddresses - Vaults to inspect for source-market liquidity.
+   * @param params.block - Block number and timestamp used for consistent RPC reads.
+   * @returns Reallocation data ready for {@link getReallocations}.
+   * @throws When the client chain does not match this market.
```
0xbulma: **[MEDIUM]** `@throws When the client chain does not match this market.` — AGENTS §3 and `docs/jsdoc-style.md` require the error class identity, not prose. The interface-level JSDoc at lines 280–299 (also new in this PR) omits `@throws` entirely.

Suggested: `@throws {ChainIdMismatchError} when the client chain does not match this market.` on both impl and interface declaration for consistency.

packages/morpho-sdk/src/helpers/constant.ts
```diff
@@ -8,3 +9,37 @@ export const DEFAULT_LLTV_BUFFER = MathLib.WAD / 200n;
...
+
+/**
+ * The default target utilization above which shared liquidity reallocations are triggered.
+ */
+export const DEFAULT_SUPPLY_TARGET_UTILIZATION = 90_5000000000000000n;
+
+/**
+ * Tokens that require setting allowance from zero before changing approval.
+ */
+export const APPROVE_ONLY_ONCE_TOKENS: Partial<Record<number, Address[]>> = {
```
0xbulma, line 26: **[MEDIUM, downgraded → LOW given context]** `APPROVE_ONLY_ONCE_TOKENS` and `MAX_TOKEN_APPROVALS` are duplicated byte-for-byte with `simulation-sdk/src/constants.ts`.

> Maintainer note (per PR thread): simulation-sdk will be deprecated.

Given simulation-sdk is going away, the duplication will resolve on its own when that package is removed. The earlier suggestion to have simulation-sdk re-export from morpho-sdk is moot. Until simulation-sdk is removed, the two copies can drift if a new token is added on one side only — worth a short note in the simulation-sdk source pointing to morpho-sdk as the canonical home, or just relying on the deprecation timeline being short.

packages/morpho-sdk/src/helpers/reallocationData.test.ts
```diff
@@ -0,0 +1,1205 @@
...
+    readonly withdrawal: PublicReallocation;
+    readonly timestamp: bigint;
+  }): ReallocationData;
+};
+
+const applyPublicReallocation = (
+  data: ReallocationData,
+  withdrawal: PublicReallocation,
+) =>
+  (data as unknown as ReallocationDataInternals).applyPublicReallocation({
```
0xbulma: **[MEDIUM]** `applyPublicReallocation` is reached via `(data as unknown as ReallocationDataInternals)`. AGENTS §2 rule 1 lists `as unknown as` as a forbidden pattern.

Suggested: change visibility of `applyPublicReallocation` (and `getLargestVaultWithdrawal`) to `public` with `@internal` JSDoc (AGENTS §3's documented mechanism), or mark `protected` and use a test-only subclass. Avoids the type-system escape hatch.

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,576 @@
...
+ * @internal
+ */
+const clonePosition = (position: Position) => new Position(position);
+
+/**
+ * Clones vault state, including queue arrays and public allocator accounting.
+ *
+ * @internal
+ */
+const cloneVault = (vault: Vault) =>
```
0xbulma, line 72: **[LOW]** Clone helpers hand-list every field (`cloneVault` lists 22, `cloneVaultMarketConfig` similarly). `Vault`'s constructor accepts a full `IVault`, so `new Vault({ ...vault, supplyQueue: [...vault.supplyQueue], withdrawQueue: [...vault.withdrawQueue], publicAllocatorConfig: vault.publicAllocatorConfig ? { ...vault.publicAllocatorConfig } : undefined })` is equivalent and shorter. The current code also shallow-aliases `pendingGuardian` (line 89) while deep-spreading `pendingTimelock` — inconsistent.

More importantly: any new field added to `IVault` / `IMarket` / `IPosition` / `IVaultMarketConfig` upstream is silently dropped at clone time, producing degraded snapshots. Fixture-based tests won't catch it because the fixtures only set fields the test author knew about.

Suggested: switch to spread-based clones and add a round-trip test that populates every documented field of each `I*` interface and asserts `cloneX(entity)` preserves them via `toEqual`.

packages/morpho-sdk/src/helpers/reallocationData.ts
```diff
@@ -0,0 +1,576 @@
...
+  /**
+   * Calculates public reallocations that can supply liquidity to `marketId`.
+   *
+   * @remarks
+   * The algorithm repeatedly chooses the single largest currently available
+   * source-market withdrawal across reallocatable vaults, applies that
+   * withdrawal to a cloned state, and repeats until no valid withdrawal remains.
+   *
+   * @param marketId - Target market to supply with shared liquidity.
+   * @param timestampOrOptions - Optional accrual timestamp or allocator options.
```
0xbulma: **[MEDIUM]** Overload signature has two latent issues:

1. **Precedence undocumented.** When both `timestampOrOptions` (as bigint) and `options.timestamp` are supplied, the standalone bigint wins (lines 329–332). The two `@param` lines describe what each slot holds but never state precedence.

2. **Silent discard.** When the caller passes an options object as the 2nd argument, `maybeOptions` (3rd arg) is silently ignored — invisible at call sites.

Suggested: collapse to a single `(marketId, options)` signature now that `computeReallocations` could move `timestamp` inside `options`. Aligns with root §1 "Small primitives that combine. No kitchen-sink helpers." If the overload is kept, document precedence in `@remarks` and that the 3rd arg is ignored when the 2nd is an object.

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,576 @@
...
+   * withdrawal to a cloned state, and repeats until no valid withdrawal remains.
+   *
+   * @param marketId - Target market to supply with shared liquidity.
+   * @param timestampOrOptions - Optional accrual timestamp or allocator options.
+   * @param maybeOptions - Allocator options when the timestamp is passed separately.
+   * @returns Computed source-market withdrawals and the post-reallocation state.
+   * @throws {@link UnknownReallocationMarketError} when the target market is absent.
+   */
+  // biome-ignore lint/complexity/useMaxParams: preserves the timestamp/options overload used by callers.
+  public getMarketPublicReallocations(
```
0xbulma, line 365: **[MEDIUM]** Public method `getMarketPublicReallocations` is missing the required `@example` block. AGENTS §6 mandates one `@example` per exported method with runnable code (imports, client setup, the call, expected return shape). As the headline method of the new public class, the example also serves MISSION.md's AI-legibility goal.

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,576 @@
...
+      typeof timestampOrOptions === "object"
+        ? options.timestamp
+        : (timestampOrOptions ?? options.timestamp);
+    const {
+      enabled = true,
+      reallocatableVaults,
+      defaultMaxWithdrawalUtilization = DEFAULT_WITHDRAWAL_TARGET_UTILIZATION,
+      maxWithdrawalUtilization = {},
+    } = options;
+    const accrualTimestamp = BigInt(
```
0xbulma, line 382: **[MEDIUM]** `accrualTimestamp = BigInt(timestampInput ?? this.getMarket(marketId).lastUpdate)` is computed before the `if (!enabled) return ...` short-circuit at line 343. When a caller passes a missing `marketId` together with `{ enabled: false }` and omits `timestamp`, this throws `UnknownReallocationMarketError` instead of returning the documented empty result.

Suggested: move the `if (!enabled) return { withdrawals: [], data: this };` check above the `accrualTimestamp` computation.

packages/morpho-sdk/src/helpers/reallocationData.ts
```diff
@@ -0,0 +1,576 @@
...
+      enabled = true,
+      reallocatableVaults,
+      defaultMaxWithdrawalUtilization = DEFAULT_WITHDRAWAL_TARGET_UTILIZATION,
+      maxWithdrawalUtilization = {},
+    } = options;
+    const accrualTimestamp = BigInt(
+      timestampInput ?? this.getMarket(marketId).lastUpdate,
+    );
+
+    if (!enabled) return { withdrawals: [], data: this };
```
0xbulma, line 380: **[MEDIUM]** Disabled branch returns `{ withdrawals: [], data: this }` while the enabled branch returns a fresh clone (line 364). Callers writing `const { data } = obj.getMarketPublicReallocations(id, { enabled: false }); /* mutate data */` would end up mutating the original `ReallocationData`. The colocated test (`reallocationData.test.ts:875`) only asserts `toEqual`, missing the identity leak.

Suggested: return `this.clone()` for symmetry, and add a `toBe` (or mutate-and-verify-original-unchanged) assertion to lock the new contract.

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,576 @@
...
+    ).filter((vault) => {
+      const vaultMarketConfig = this.vaultMarketConfigs[vault]?.[marketId];
+
+      return (
+        vaultMarketConfig?.enabled === true &&
+        this.vaults[vault]?.publicAllocatorConfig != null
+      );
+    });
+
+    let data = this.clone();
```
0xbulma, line 405: **[MEDIUM]** Loop-clone performance: every iteration calls `data = data.applyPublicReallocation(...)`, and `applyPublicReallocation` (line 520) calls `this.clone()` — which deep-clones every entry of `markets`, `vaults`, `positions`, and `vaultMarketConfigs` via the constructor (lines 160–211). For a target with N source markets across V vaults, total work is O((N+1) × (|markets| + |vaults| + |positions| + |vaultMarketConfigs|)) per call, and `computeReallocations` invokes it up to twice (friendly + aggressive).

> Maintainer note (per PR thread): simulation-sdk will be deprecated, so comparing to its immer-based copy-on-write is informational only — not a target to restore.

Suggested (intrinsic to this code, independent of simulation-sdk): clone only the entities `applyPublicReallocation` actually mutates (the two markets, the two positions, the two vault-market-configs, and the vault's `publicAllocatorConfig` + `totalAssets`) and alias the rest. Avoids full deep clones per iteration.

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,576 @@
...
+    readonly withdrawal: { readonly id: MarketId; readonly assets: bigint };
+    readonly timestamp: bigint;
+  }) {
+    const data = this.clone();
+
+    const vaultPublicAllocatorConfig =
+      data.getVault(vault).publicAllocatorConfig;
+    if (vaultPublicAllocatorConfig == null)
+      throw new MissingPublicAllocatorConfigError(vault);
+    vaultPublicAllocatorConfig.accruedFee += vaultPublicAllocatorConfig.fee;
```
0xbulma, line 594: **[LOW]** `applyPublicReallocation` accrues fee per-withdrawal: each iteration unconditionally does `vaultPublicAllocatorConfig.accruedFee += fee`. When a single vault contributes N withdrawals to one onchain `reallocateTo` call, the simulated `accruedFee` overcounts by N (onchain accrues exactly once per `reallocateTo`).

The fee placed into the actual bundle is correct — `computeReallocations.ts:218` reads `data.getVault(vault).publicAllocatorConfig.fee` separately — so this only affects the simulated `accruedFee` field on the returned `data`. The colocated test asserts that value (`reallocationData.test.ts:914`) so the over-count is observable to consumers.

> Maintainer note (per PR thread): simulation-sdk will be deprecated, so the prior `@warning` on `SimulationState.getMarketPublicReallocations` will disappear with it. That makes restoring the warning on this method more important, not less, since callers reading `accruedFee` won't have any other documented heads-up.

Suggested: either fix by tracking which vaults already accrued in this iteration (add the fee once per `(vault, supplyMarketId)` group), or add an `@warning` to `getMarketPublicReallocations` JSDoc so the over-estimate is documented.

packages/morpho-sdk/src/helpers/computeReallocations.ts
```diff
@@ -79,11 +151,15 @@ export const computeReallocations = ({
...
-        defaultMaxWithdrawalUtilization: MathLib.WAD,
-        maxWithdrawalUtilization: {},
-      }).withdrawals,
+      ...friendlyReallocationData.getMarketPublicReallocations(
+        market.id,
+        options?.timestamp,
+        {
+          ...options,
+          defaultMaxWithdrawalUtilization: MathLib.WAD,
+          maxWithdrawalUtilization: {},
```
0xbulma: **[LOW]** Phase 2 hard-codes `maxWithdrawalUtilization: {}` alongside `defaultMaxWithdrawalUtilization: MathLib.WAD`, discarding any caller-provided per-market overrides. This matches legacy behavior, but the per-market override is new public API via `ReallocationComputeOptions` in this PR and the contract isn't documented.

Suggested: one-sentence `@remarks` ("per-market `maxWithdrawalUtilization` overrides apply only to phase 1; phase 2 forces 100% utilization on every source market") so integrators aren't surprised when a per-market cap is silently ignored in the fallback path.

packages/morpho-sdk/src/helpers/computeReallocations.test.ts
```diff
@@ -121,7 +120,7 @@ function makeMockState({
         ? { admin: vault, fee: vaultFees[vault]!, accruedFee: 0n }
         : undefined,
     }),
-  } as unknown as SimulationState;
+  } as unknown as ReallocationData;
```
0xbulma, line 132: **[LOW]** `makeMockState` returns `{} as unknown as ReallocationData`. Forbidden by §2 rule 1. Silent test-rot risk: today the mock implements only `getMarket`/`getMarketPublicReallocations`/`getVault`; if `computeReallocations` starts using a new method, the mock returns `undefined` at runtime and tests may pass or throw with a non-typed error rather than catch the regression.

Suggested: type the parameter as `Pick<ReallocationData, "getMarket" | "getMarketPublicReallocations" | "getVault">` and accept that surface in `computeReallocations`, or add a contract test that constructs a real `ReallocationData` covering the methods exercised.

packages/liquidity-sdk-viem/codegen.ts
```diff
@@ -2,6 +2,8 @@ import type { CodegenConfig } from "@graphql-codegen/cli";
 
 import { BLUE_API_GRAPHQL_URL } from "@morpho-org/morpho-ts";
 
+const hexStringScalar = "`0x$" + "{string}`";
```
0xbulma, line 6: **[LOW]** `const hexStringScalar = "`0x$" + "{string}`";` is an undocumented workaround. The string concatenation produces the literal `` `0x${string}` `` while preventing the codegen tool from interpreting `${string}` as a substitution. Same unexplained code appears in `packages/liquidation-sdk-viem/codegen.ts`.

Suggested: add a one-line comment in both files (e.g. `// Concatenated to defeat the codegen parser interpreting ${string} as a substitution.`) so future maintainers don't "simplify" it back to the broken form.

packages/morpho-sdk/src/helpers/computeReallocations.ts
```diff
@@ -133,22 +206,20 @@ export const computeReallocations = ({
...
-      vault: vault as Address,
+  return reallocations
+    .filter(({ withdrawals: vaultWithdrawals }) => vaultWithdrawals.length > 0)
+    .map(({ vault, withdrawals: vaultWithdrawals }) => ({
+      vault,
       fee: (() => {
-        const config = data.getVault(vault as Address).publicAllocatorConfig;
+        const config = data.getVault(vault).publicAllocatorConfig;
         if (config == null) {
           throw new MissingPublicAllocatorConfigError(vault);
```
0xbulma, line 254: **[MEDIUM]** This `throw` of `MissingPublicAllocatorConfigError` is not reflected in the function's JSDoc `@throws` (line 95), which lists only `InsufficientSharedLiquidityError`.

Suggested: add `@throws {MissingPublicAllocatorConfigError} when a selected vault is missing its public allocator config.` to the JSDoc block.

packages/morpho-sdk/src/helpers/constant.ts
0xbulma: Simulation will be deprecated

packages/morpho-sdk/src/entities/reallocationData.test.ts
```diff
@@ -0,0 +1,1205 @@
...
+        marketIds: [targetParams.id, sourceParams.id],
+        vaults: [VAULT],
+      }),
+    ).toEqual(
+      snapshotPreservedState(legacyResult.data, {
+        marketIds: [targetParams.id, sourceParams.id],
+        vaults: [VAULT],
+      }),
+    );
+  });
```
0xbulma, line 793: **[MEDIUM]** The 3-arg overload `getMarketPublicReallocations(marketId, timestampOrOptions, maybeOptions)` has untested precedence when both `timestampOrOptions` is a `bigint` AND `maybeOptions.timestamp` is provided. Source lines 329–332 resolve `timestampInput = timestampOrOptions ?? options.timestamp`, so the positional bigint wins — but no test passes both with distinct values to lock that in. The existing overload test at 759–784 only sets the positional bigint (`maybeOptions` has no `timestamp`); the test at 876–881 sets only `maybeOptions.timestamp` (positional is `undefined`).

Suggested: add a case that passes a positional `TIMESTAMP` AND a different `maybeOptions.timestamp`, then asserts the result matches the positional-only baseline so a future refactor that flips precedence fails loudly.

Note: this finding becomes moot if you take the suggestion in https://github.com/morpho-org/sdks/pull/655#discussion_r3264698205 (collapse the overload into a single `(marketId, options)` signature).

Foulks-Plb (dismissed):

Rubilmax (commented):
_(duplicate review created by a transient 502 retry; inline comments removed — see the next review for the real review of 21 findings.)_

packages/liquidity-sdk-viem/src/loader.ts
```diff
@@ -231,6 +188,7 @@ export class LiquidityLoader<chain extends Chain = Chain> {
...
                 maxWithdrawalUtilization,
                 enabled: true,
               });
```
Rubilmax, line 193: **[HIGH]** (anchored here; actual code at lines 202-206 of the new file)

`catch (error) { return Error(\`An error occurred while simulating reallocations: ${error}\`); }` unwraps every typed exception from `ReallocationData.getMarketPublicReallocations` (e.g. `UnknownReallocationMarketError`, `MissingPublicAllocatorConfigError`, `DisabledReallocationMarketError`, `UnknownReallocationVaultMarketConfigError`, `InsufficientSharedLiquidityError`) into an untyped stringified `Error`. `DataLoader` rejects with that flat Error, so integrators cannot `instanceof`-match the typed classes that `morpho-sdk` exports as public API (AGENTS.md §3 typed-errors as public API; "preserve `cause` when wrapping"). The catch block is pre-existing, but the new typed-error surface from `ReallocationData` makes the swallow materially worse.

_Suggestion: Re-throw `error` directly (DataLoader rejects with it untouched), or wrap with `new Error("simulating reallocations failed", { cause: error })` to preserve class identity; update the colocated test to assert by class rather than message text._

packages/liquidity-sdk-viem/test/loader.unit.test.ts
```diff
@@ -0,0 +1,284 @@
+import { ChainId, type MarketId } from "@morpho-org/blue-sdk";
```
Rubilmax: **[MEDIUM]** New 284-line unit test placed under `test/` even though `liquidity-sdk-viem`'s sibling `src/loader.test.ts` is already colocated and the existing project glob picks up `src/**/*.test.ts`. The test also mocks named imports of `@morpho-org/blue-sdk-viem`, `@morpho-org/morpho-sdk`, and `viem/actions` via `vi.doMock`, rather than the AGENTS.md §2 #6 permitted exception (`createMockClient` transport mock from `@morpho-org/test/mock`).

_Suggestion: Move next to source as `packages/liquidity-sdk-viem/src/loader.unit.test.ts` (or fold cases into `src/loader.test.ts`), and migrate to transport-level `createMockClient` so the test path resolves through `client.transport` like production code (§5)._

Rubilmax: **[LOW]** Uses the `.unit.test.ts` suffix while every other test in this package uses `.test.ts`. No documented convention for the variant.

_Suggestion: Drop the `.unit` segment for consistency, or, if the unit/fork distinction is intentional, document it in `packages/liquidity-sdk-viem/AGENTS.md` and apply consistently._

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -1017,31 +1025,30 @@ export class MorphoMarketV1 implements MarketV1Actions {
...
-      block,
       markets: marketsRecord,
       vaults: vaultsRecord,
       vaultMarketConfigs: vaultMarketConfigsRecord,
       positions: positionsRecord,
-      holdings: holdingsRecord,
     });
   }
 
+  /**
```
Rubilmax, line 1041: **[MEDIUM]** `MorphoMarketV1.getReallocations` (both the `MarketV1Actions` interface at lines 309–328 and the class method at 1046–1061) has no `@throws` annotation. It delegates to `computeReallocations`, which can throw `InsufficientSharedLiquidityError`, `MissingPublicAllocatorConfigError`, and (via `data.getMarket`) `UnknownReallocationMarketError`. `docs/jsdoc-style.md` §5 mandates one `@throws {ErrorClass} when …` per exported error class an integrator may pattern-match on.

_Suggestion: Add `@throws` entries on both the interface declaration and the class method covering all three classes (plus any from `validateReallocations` if borrow paths exercise them indirectly)._

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -289,20 +287,24 @@ export interface MarketV1Actions {
...
    * @param params.block - The block to fetch data at (number and timestamp).
-   * @returns A SimulationState populated with all required data.
+   * @returns A ReallocationData instance populated with all required data.
+   * @throws {ChainIdMismatchError} when the client chain does not match this market.
    */
   getReallocationData: (params: {
     vaultAddresses: readonly Address[];
-    block: MinimalBlock;
-  }) => Promise<SimulationState>;
+    block: {
```
Rubilmax, line 303: **[LOW]** The block parameter shape `{ readonly number: bigint; readonly timestamp: bigint }` is inlined twice in this file (once on the `MarketV1Actions.getReallocationData` interface declaration at lines 301–306, once on the `MorphoMarketV1.getReallocationData` class method at lines 928–934). AGENTS.md §3 says "Reuse SDK types … instead of re-declaring at call sites."

_Suggestion: Define a `ReallocationBlock` type next to `PublicAllocatorOptions` in `types/sharedLiquidity.ts` and have both signatures reference it._

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
```diff
@@ -1017,31 +1025,30 @@ export class MorphoMarketV1 implements MarketV1Actions {
...
   getReallocations({
     reallocationData,
     borrowAmount,
     options,
   }: {
-    reallocationData: SimulationState;
+    reallocationData: ReallocationData;
     borrowAmount: bigint;
     options?: ReallocationComputeOptions;
   }): readonly VaultReallocation[] {
```
Rubilmax, line 1062: **[LOW]** `MorphoMarketV1.getReallocations` does not assert `reallocationData.chainId === this.chainId`. A caller passing `ReallocationData` built for chain X into a market on chain Y silently produces nonsensical reallocations; the mismatch only surfaces downstream in `borrow()`/`supplyCollateralBorrow()`'s `validateChainId`.

_Suggestion: At the top of `getReallocations`, call `validateChainId(reallocationData.chainId, this.chainId)` (or a dedicated helper) so the cross-chain mismatch is caught at the entity boundary._

packages/morpho-sdk/src/helpers/computeReallocations.ts
```diff
@@ -1,43 +1,117 @@
...
+ * @remarks Per-market `maxWithdrawalUtilization` overrides apply only to phase 1; phase 2 forces 100% utilization on every source market.
+ * @param params.reallocationData - The local state containing market, vault, and position data.
  * @param params.marketId - The target market to reallocate liquidity into.
  * @param params.borrowAmount - The intended borrow amount (used to compute post-borrow utilization).
  * @param params.options - Optional reallocation computation options.
  * @returns Array of vault reallocations, sorted with withdrawals in ascending market id order.
- * @throws {InsufficientSharedLiquidityError} If shared liquidity cannot cover the borrow shortfall on the target market — preventing fee-bearing reallocations from being attached to a borrow that would still revert onchain.
+ * @throws {InsufficientSharedLiquidityError} when shared liquidity cannot cover the borrow shortfall on the target market — preventing fee-bearing reallocations from being attached to a borrow that would still revert onchain.
+ * @throws {MissingPublicAllocatorConfigError} when a selected vault is missing its public allocator config.
  */
```
Rubilmax, line 138: **[MEDIUM]** `computeReallocations` is re-exported from `helpers/index.ts` and the package root, but its JSDoc has no `@example` block. `docs/jsdoc-style.md` §6 mandates exactly one runnable `@example` per exported function.

_Suggestion: Add a single `@example` showing the typical chain: `getReallocationData → computeReallocations → borrow`, matching the canonical exemplar in `actions/marketV1/borrow.ts`._

packages/morpho-sdk/src/helpers/computeReallocations.ts
```diff
@@ -1,43 +1,117 @@
...
   readonly marketId: MarketId;
   readonly borrowAmount: bigint;
   readonly options?: ReallocationComputeOptions;
 }): readonly VaultReallocation[] => {
   if (options?.enabled === false) return [];
 
-  const market = data.getMarket(marketId).accrueInterest(data.block.timestamp);
+  // ReallocationData does not retain the fetch block; pass that block timestamp
+  // to compute against the same accrued state, otherwise Market defaults to lastUpdate.
+  const market = data.getMarket(marketId).accrueInterest(options?.timestamp);
```
Rubilmax, line 154: **[LOW]** `computeReallocations` reads `options?.timestamp` and passes it to `market.accrueInterest(options?.timestamp)`. When the caller forgets to forward `block.timestamp` from `getReallocationData`, `accrueInterest(undefined)` silently falls back to the market's own `lastUpdate`, which can diverge from the block the source rows were fetched at, producing internally-inconsistent plans whose on-chain utilization no longer matches at submit time.

_Suggestion: Require `options.timestamp` (or default it from a timestamp stored on `ReallocationData`); fail loudly when missing rather than silently using a stale `lastUpdate`._

packages/morpho-sdk/src/helpers/computeReallocations.ts
```diff
@@ -133,19 +204,22 @@ export const computeReallocations = ({
...
         }
         return config.fee;
       })(),
       withdrawals: vaultWithdrawals
-        // Reallocation withdrawals must be sorted by market id in ascending order.
+        // Reallocation withdrawals must be sorted by market id in ascending
+        // order. Use a byte-wise comparison, not `localeCompare`: the
+        // PublicAllocator contract orders by the market id's raw bytes, and
+        // `localeCompare` is locale-dependent and not guaranteed to match.
         .sort(({ id: idA }, { id: idB }) =>
```
Rubilmax: **[LOW]** The reallocation withdrawal sort uses `idA > idB ? 1 : idA < idB ? -1 : 0`, which is byte-wise correct only when all market IDs are consistently lowercase hex. `MarketUtils.getMarketId` emits lowercase, but `isMarketId` accepts mixed-case `0-9A-Fa-f`. A caller-supplied mixed-case `MarketId` would mis-sort and trigger PublicAllocator's `InconsistentWithdrawals` revert on-chain. Same risk in `validateReallocations`.

_Suggestion: Normalize via `marketParams.id.toLowerCase()` as the sort key (and in `validateReallocations`'s `prevId` comparison), or assert lowercase in the `MarketId` brand at construction time._

packages/morpho-sdk/src/helpers/constant.ts
```diff
@@ -8,3 +9,37 @@ export const DEFAULT_LLTV_BUFFER = MathLib.WAD / 200n;
...
+
+/**
+ * The default maximum utilization allowed to reach when withdrawing shared liquidity.
+ */
+export const DEFAULT_WITHDRAWAL_TARGET_UTILIZATION = 92_0000000000000000n;
+
+/**
+ * The default target utilization above which shared liquidity reallocations are triggered.
+ */
+export const DEFAULT_SUPPLY_TARGET_UTILIZATION = 90_5000000000000000n;
```
Rubilmax, line 21: **[HIGH]** `DEFAULT_SUPPLY_TARGET_UTILIZATION = 90_5000000000000000n` is newly exported from `morpho-sdk` and re-exported from `morpho-sdk/src/helpers/index.ts`, but the same constant with the same value is still publicly exported from `packages/bundler-sdk-viem/src/operations.ts` (around line 44). Two source-of-truth modules for the same protocol constant — AGENTS.md §1 Modularity violation ("Duplication is a refactor, not a feature").

_Suggestion: Pick one canonical home (most likely `morpho-sdk` since the constant now drives a public reallocation API there), re-export from `bundler-sdk-viem` (`export { DEFAULT_SUPPLY_TARGET_UTILIZATION } from "@morpho-org/morpho-sdk"`), and delete the local definition in `bundler-sdk-viem/src/operations.ts`._

packages/morpho-sdk/src/helpers/constant.ts
Rubilmax, line 26: **[HIGH]** `APPROVE_ONLY_ONCE_TOKENS` and `MAX_TOKEN_APPROVALS` are now defined locally (lines 26 and 36) and re-exported from `morpho-sdk/src/helpers/index.ts`, but the **identical** constants still live and ship publicly from `packages/simulation-sdk/src/constants.ts` (re-exported via `simulation-sdk/src/index.ts`) and are consumed by `packages/bundler-sdk-viem/src/operations.ts` and `packages/bundler-sdk-viem/src/actions.ts`. Two source-of-truth modules will drift on the next chain/token addition: bundler-sdk-viem and morpho-sdk can disagree on USDT-like reset-to-zero behavior (residual allowance / on-chain revert) or on `maxUint96`-capped tokens (UNI/COMP/FLUID/ONDO → on-chain `approve` revert).

_Suggestion: Pick one canonical home (e.g. `morpho-sdk/helpers/constant.ts`) and re-export from the other consuming packages, or extract a tiny shared `@morpho-org/approval-registry` consumed by both `morpho-sdk` and `bundler-sdk-viem`. As a backstop, add a CI parity check that diffs the two literal lists._

packages/morpho-sdk/src/helpers/index.ts
```diff
@@ -1,5 +1,18 @@
...
+  APPROVE_ONLY_ONCE_TOKENS,
+  DEFAULT_LLTV_BUFFER,
+  DEFAULT_SUPPLY_TARGET_UTILIZATION,
+  DEFAULT_WITHDRAWAL_TARGET_UTILIZATION,
+  MAX_ABSOLUTE_SHARE_PRICE,
+  MAX_SLIPPAGE_TOLERANCE,
+  MAX_TOKEN_APPROVALS,
+} from "./constant.js";
 export { addTransactionMetadata } from "./metadata.js";
+export {
```
Rubilmax: **[LOW]** `ReallocationData` is now part of the published API of `@morpho-org/morpho-sdk`, but its public fields (`markets`, `vaults`, `positions`, `vaultMarketConfigs`) are not documented in JSDoc as inputs an integrator may inspect vs. opaque state.

_Suggestion: Either mark the fields `@internal` and route access through the existing getters, or document the read contract on the class JSDoc and switch the declared types to `Readonly<Record<..., Readonly<...>>>` so the public contract matches what integrators can safely do._

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,621 @@
...
+/**
+ * Narrow state container for computing public allocator reallocations.
+ *
+ * @remarks
+ * The class owns only the market, vault, position, vault-market-config,
+ * and chain data needed by the shared-liquidity algorithm. Constructor inputs
+ * are cloned, and simulation steps return cloned `ReallocationData` instances
+ * so fetched caller inputs are not mutated.
+ */
+export class ReallocationData implements InputReallocationData {
```
Rubilmax, line 116: **[MEDIUM]** `ReallocationData` is placed under `src/helpers/`, but the helpers boundary is "encode-only helpers (ABI-only); validators and constants are pure" (root §1) and "they return new objects and never mutate inputs" (helpers AGENTS.md). This class is a stateful container with reads off `this` (`getMarket`/`getVault`/…), a public-allocator simulation loop, and a protected `applyPublicReallocation` that mutates fresh clones in place (line 562 `accruedFee += fee`, lines 592–595 mutate `maxIn`/`maxOut`). It is closer to a domain entity than a helper.

_Suggestion: Move `reallocationData.ts` out of `helpers/` to `src/entities/` (peer of `marketV1`) or a new `src/state/` namespace, and update the per-folder `AGENTS.md` files so the helpers contract continues to mean "pure utilities."_

Rubilmax (commented):
## Local-only Code Review

**Range:** `0ff61914..1b736238` (24 files)

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 3 |
| Medium | 8 |
| Low | 10 |

Generated via `/pr-review-local`. 21 findings posted as inline comments below.

_Scope filter dropped 7 out-of-scope findings targeting AGENTS.md files outside the diff (`src/types/AGENTS.md`, `src/helpers/AGENTS.md`, `src/entities/marketV1/AGENTS.md`, `.agents/personas/code-quality.md`) and 2 missing-test paths (`constant.test.ts`, `encodeErc20Approval.test.ts`)._

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,621 @@
...
+  /** Chain id associated with the fetched reallocation data. */
+  public readonly chainId: number;
```
Rubilmax, line 118: **[MEDIUM]** `markets`, `vaults`, `positions`, `vaultMarketConfigs` are declared `public readonly`, but the bindings only protect the reference — the inner Record values, nested per-vault records, and the entity instances (`Market`, `Vault`, `VaultMarketConfig`, `Position`, including `Vault.publicAllocatorConfig` and `VaultMarketPublicAllocatorConfig`) are all mutable. The constructor accepts `Readonly<Record<...>>` (lines 37–50) but drops the inner `Readonly` on storage, so a consumer can write `data.markets[id] = bogus` after construction. `applyPublicReallocation` itself relies on this mutability (line 562 + 592–595). AGENTS.md §3 says "`readonly` on every public field. Frozen outputs."

_Suggestion: Either `deepFreeze` the result of every constructor and `clone()` call (and route every internal mutating step through a private writable copy before freezing on return), or declare the public fields as `Readonly<Record<MarketId, Readonly<Market> | undefined>>` etc. and switch internal writes to a narrow type assertion or a private mutable mirror._

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,621 @@
...
+    options: PublicAllocatorOptions = {},
+  ): {
+    readonly withdrawals: readonly PublicReallocation[];
+    data: ReallocationData;
+  } {
+    const {
+      enabled = true,
+      timestamp,
+      reallocatableVaults,
+      defaultMaxWithdrawalUtilization = DEFAULT_WITHDRAWAL_TARGET_UTILIZATION,
```
Rubilmax, line 376: **[MEDIUM]** The original `SimulationState.getMarketPublicReallocations` took a `delay` parameter (default 1h via `Time.s.from.h(1n)`) that called `accrueInterest(timestamp + delay)` on the **target-market vault position** before computing `suppliable`, conservatively reducing it to absorb interest accrual between plan computation and tx inclusion. The new `getMarketPublicReallocations` drops this — `accrualTimestamp = timestamp ?? lastUpdate` only. For markets near cap, `suppliable` is now over-estimated by ~1h of accrual; the resulting `reallocateTo` can revert on-chain because the projected target supply exceeds the cap by the time the tx lands. The matching changeset declares this as part of the major bump but the JSDoc on `getMarketPublicReallocations` does not call out the lost margin.

_Suggestion: Either restore an opt-in `delay` (or `safetyMargin`) on `PublicAllocatorOptions` applied to target-vault accrual, or document the loss-of-margin explicitly in `getMarketPublicReallocations`'s JSDoc and the changeset, so integrators know to add their own margin._

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,621 @@
...
+    data.positions[vault] ??= {};
+    data.positions[vault]![withdrawal.id] = clonePosition(sourcePosition);
+
+    const { position: supplyPosition } = data
+      .getAccrualPosition(vault, supplyMarketId)
+      .supply(withdrawal.assets, 0n, timestamp);
+    data.markets[supplyMarketId] = cloneMarket(supplyPosition.market);
+    data.positions[vault]![supplyMarketId] = clonePosition(supplyPosition);
+
+    void _try(() => {
```
Rubilmax, line 642: **[MEDIUM]** `void _try(() => { vaultData.totalAssets = data.getAccrualPosition(vault, marketId).supplyAssets … }, UnknownDataError)` recomputes `totalAssets` over the withdraw queue but discards both the return and any `UnknownDataError`. If a market in `withdrawQueue` lacks an accrual position (partial state, missing fetch), `totalAssets` remains at the pre-reallocation value while market/position state above has already been updated, leaving the returned `data` snapshot internally inconsistent. The `void` discards even the signal that recompute failed.

_Suggestion: Either guarantee the precondition (every market in `withdrawQueue` has a position) and drop the `_try`, or surface the inconsistency via a typed error rather than silently leaving stale `totalAssets`._

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,621 @@
...
+
+    return data;
+  }
+
+  /**
+   * Gets a market by id.
+   *
+   * @param marketId - Market id to read.
+   * @returns The cloned market data.
+   * @throws {@link UnknownReallocationMarketError} when the market is absent.
```
Rubilmax, line 234: **[LOW]** `@throws` annotations in this file use `@throws {@link UnknownReallocationMarketError}`. `docs/jsdoc-style.md` §5 specifies the canonical format as `@throws {ErrorClass} when <condition>` (no `@link`), matching the exemplar in `actions/marketV1/borrow.ts`. Same issue on lines 244, 260, 277, 278, 293, 322, 535, 536.

_Suggestion: Replace each `{@link ClassName}` with `{ClassName}` and ensure each `@throws` includes a `when …` clause that reads like an instruction._

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,621 @@
...
+  }):
+    | {
+        readonly vault: Address;
+        readonly largestWithdrawal?: {
+          readonly id: MarketId;
+          readonly assets: bigint;
+        };
+      }
+    | undefined {
+    return _try(() => {
```
Rubilmax, line 498: **[LOW]** Outer `_try(() => { ... }, UnknownDataError)` in `getLargestVaultWithdrawal` (and the inner one at line 491) swallow `UnknownDataError` from `getVaultMarketConfig` / `getAccrualPosition` / `getVault`, returning `undefined`. The callsite filters undefined out (line 410), so a vault with missing source-market data is silently dropped from the candidate pool with no diagnostic, which can yield an under-planned reallocation followed by an `InsufficientSharedLiquidityError` from `computeReallocations` (with no signal of which vault was silently skipped).

_Suggestion: Optionally surface skipped vaults — e.g. accumulate a `skipped: Address[]` alongside `withdrawals`, or include a debug-only callback option — so callers with incomplete state can distinguish "no liquidity" from "stale input"._

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,621 @@
...
+    data.positions[vault] = { ...(this.positions[vault] ?? {}) };
+    data.vaultMarketConfigs[vault] = {
+      ...(this.vaultMarketConfigs[vault] ?? {}),
+    };
+
+    const vaultPublicAllocatorConfig =
+      data.getVault(vault).publicAllocatorConfig;
+    if (vaultPublicAllocatorConfig == null)
+      throw new MissingPublicAllocatorConfigError(vault);
+    vaultPublicAllocatorConfig.accruedFee += vaultPublicAllocatorConfig.fee;
```
Rubilmax, line 594: **[LOW]** The JSDoc at lines 312–318 warns that `vault.publicAllocatorConfig.accruedFee` may be over-estimated (simulation accrues fee once per computed withdrawal, on-chain `reallocateTo` charges once per vault group), but `liquidity-sdk-viem/src/loader.ts:198` exposes the resulting `endState` (with the over-accrued figure) to all consumers with no caveat in the loader API.

_Suggestion: Mirror the caveat on `LiquidityLoader.fetch`'s return-shape JSDoc, OR fix `applyPublicReallocation` to increment `accruedFee` only on the FIRST withdrawal per `(vault, supplyMarketId)` so the simulated and on-chain values match._

packages/morpho-sdk/test/actions/marketV1/reallocations.test.ts
```diff
@@ -655,6 +655,7 @@ describe("getReallocationData and getReallocations", () => {
         const reallocations = market.getReallocations({
```
Rubilmax, line 670: **[LOW]** (anchored here; the relevant `describe("getReallocationData and getReallocations", …)` block starts at line 618)

The fork-based `getReallocationData`/`getReallocations` suite covers happy paths and the `reallocations === []` case, but does not assert the documented `ChainIdMismatchError` throw from `getReallocationData`'s JSDoc (`@throws {ChainIdMismatchError} when the client chain does not match this market`).

_Suggestion: Add a test that constructs a `MorphoClient` whose `client.chain.id` differs from the market's chainId and asserts the rejection is `instanceof ChainIdMismatchError`._

packages/morpho-sdk/test/reallocationData/publicAllocator.test.ts
```diff
@@ -0,0 +1,534 @@
+import {
```
Rubilmax, line 1: **[MEDIUM]** New 534-line unit test placed under `test/` for the new colocated module `src/helpers/reallocationData.ts`. `morpho-sdk` is colocation-wired and the PR already adds the colocated `src/helpers/reallocationData.test.ts`. Per root §5 and `packages/morpho-sdk/AGENTS.md` ("tests for this package are colocated"), new tests for new modules must be colocated.

_Suggestion: Move to `packages/morpho-sdk/src/helpers/reallocationData.publicAllocator.test.ts` (or fold into the existing colocated test)._

packages/liquidity-sdk-viem/src/loader.ts
Rubilmax: Was already the case, out-of-scope

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
Rubilmax: Inlining twice is fine. Thrice is maybe too many

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
Rubilmax: @Foulks-Plb wdyt?

packages/morpho-sdk/src/helpers/constant.ts
Rubilmax: Overkill. Part of the process

packages/morpho-sdk/src/helpers/constant.ts
Rubilmax: Overkill. Part of the process

packages/morpho-sdk/src/helpers/computeReallocations.ts
Rubilmax: Documented

packages/morpho-sdk/src/entities/reallocationData.ts
Rubilmax: feature request: ignored

packages/morpho-sdk/src/entities/reallocationData.ts
Rubilmax: wasn't the case before, feature request: ignored

packages/morpho-sdk/src/entities/reallocationData.ts
Rubilmax: @Foulks-Plb wdyt?

Foulks-Plb: mmmm, imo deepFreeze + clone ? 
I think this is worth fixing. In this flow, a consumer could mutate the requested ETH amounts after construction, which makes the whole pipeline fairly fragile.

packages/morpho-sdk/src/entities/marketV1/marketV1.ts
Foulks-Plb: great, it's a nice to have.

0xbulma (dismissed):

0xbulma:
**[MEDIUM] — rebase needed**

Branch is 1 commit behind `origin/main`. The release commit `f8dc64e6` (`chore: version packages (main) (#713)`) bumped:

- `@morpho-org/liquidity-sdk-viem` → `2.0.1`
- `@morpho-org/liquidation-sdk-viem` → `4.0.0`

This branch's working tree still has `2.0.0` and `3.0.0` respectively, so the PR's diff against current `main` reads as a version downgrade on both `package.json` files. Once merged without a rebase, Changesets will compute the next bump against the stale base (e.g. `2.0.0 → 3.0.0` major for liquidity-sdk-viem), skipping the `2.0.1` patch line that's already on the registry.

Not introduced by the PR's commits — it's an artifact of the stale base — but a real merge concern.

**Suggested:** rebase `extract-public-reallocation` on `origin/main` to pick up commit `f8dc64e6`. The two changeset markdowns stay as-is; Changesets will then compute `2.0.1 → 3.0.0` (major) and `4.0.0 → 5.0.0` cleanly.

_Filed as a general comment since the version field wasn't modified by the PR commits — only by the missing release commit on `main`._

Foulks-Plb (dismissed):

packages/morpho-sdk/src/entities/reallocationData.ts
Rubilmax: If RPC lies or data is mutated, onchain will revert thanks to the first line check in `PublicAllocator.reallocateTo` which checks `msg.value == fee[vault]` so no vulnerability on this

Deep-freezing an entity would be the first time we do such thing; instead we prefer deep-freezing a tx but never inputs to the tx

Foulks-Plb (dismissed):

packages/morpho-sdk/src/helpers/computeReallocations.ts
```diff
@@ -1,43 +1,165 @@
...
+    }
+
+    remainingAssets -= cappedAssets;
+    if (remainingAssets === 0n) break;
+  }
+
+  return cappedWithdrawals;
+};
+
+const compareMarketIds = (idA: MarketId, idB: MarketId) => {
```
Foulks-Plb: You already have this helper in validate.ts

packages/morpho-sdk/src/entities/reallocationData.ts
```diff
@@ -0,0 +1,632 @@
...
```
Foulks-Plb, line 594: **[LOW] `accruedFee` over-counting leaks into the public API surface**

`applyPublicReallocation` accrues the public allocator fee once per simulated withdrawal:

```ts
vaultPublicAllocatorConfig.accruedFee += vaultPublicAllocatorConfig.fee;
```

…while the on-chain `PublicAllocator.reallocateTo` charges the fee once per vault reallocation group. When a single vault contributes N withdrawals, the simulated `accruedFee` overshoots by `(N-1) * fee`.

This is documented in the JSDoc on `getMarketPublicReallocations` (line 325) and in the changeset — good defensive disclosure. But the over-counted value is still exposed verbatim through `endState.getVault(vault).publicAllocatorConfig.accruedFee`, which is a publicly readable field. Any integrator who inspects `endState` to display "fees paid for this reallocation plan" or to feed analytics will silently get the wrong number, with no type-level signal that the field is unreliable post-simulation.

Two options worth considering:

1. **Normalize on output**: in `getMarketPublicReallocations`, after the loop completes, deduplicate per-vault fee accrual on the returned `data` so the exposed value matches the on-chain expectation.
2. **Don't expose it**: clear `publicAllocatorConfig.accruedFee` on the returned `endState`, or set it to `undefined`, so callers can't accidentally rely on a value the simulation doesn't model faithfully.

Option 1 keeps the field semantically meaningful; option 2 makes the contract honest about what `ReallocationData` does and doesn't track.

Foulks-Plb: It doesn’t seem serious, just a bit confusing from a devex perspective, no?

Foulks-Plb (approved):

PR #656, opened 2026-05-12
chore(morpho-sdk): deprecate MorphoClient in favor of morphoViemExtension

Foulks-Plb:
## Summary

- Marks `MorphoClient` (class + constructor) as `@deprecated` and points integrators at `morphoViemExtension`, so the SDK rides on top of a viem client (public or wallet) the integrator already owns.
- Adds a before/after JSDoc example using `createPublicClient(...).extend(morphoViemExtension())` and `client.morpho.vaultV1(...)`, so a single client serves both native viem reads and Morpho factories.
- Adds a `morpho-sdk` minor changeset announcing the deprecation and the removal in the next major.

## Test plan

- [ ] CI typecheck / lint
- [ ] Confirm changeset is picked up by the release workflow

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1778592268667869)

### DIFF
--- a/.changeset/deprecate-morpho-client.md
+++ b/.changeset/deprecate-morpho-client.md
@@ -1,0 +1,5 @@
+---
+"@morpho-org/morpho-sdk": minor
+---
+
+Deprecate `MorphoClient` in favor of `morphoViemExtension`. Extend a viem public (or wallet) client with `morphoViemExtension(...)` and use `client.morpho.vaultV1 / vaultV2 / marketV1` instead of constructing `MorphoClient` directly. `MorphoClient` will be removed in the next major release.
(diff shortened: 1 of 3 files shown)


Rubilmax (commented):
This is actually a breaking change
I would prefer to ship it as a major and thus wait a bit longer for our first implementation to be fully stable

Foulks-Plb:
> This is actually a breaking change I would prefer to ship it as a major and thus wait a bit longer for our first implementation to be fully stable

It’s not a breaking change today, but it does announce a future breaking change.
Why wait? The whole point of the deprecated label is precisely to discourage its use in future integrations.

Rubilmax (approved):

PR #668, opened 2026-05-13
docs: Add SDK consolidation and deprecation TIBs

Rubilmax:
## Summary

Adds two proposed TIBs under `docs/tibs/`:

- `TIB-0002`: consolidates SDK package consumption around `@morpho-org/morpho-sdk`, including direct dependency ownership for the canonical consumer entrypoint, minimized public subpaths, tree-shakeability requirements, re-export strategy, utility export hierarchy, deprecation scope, and release coordination rules.
- `TIB-0003`: defines the default SDK package deprecation lifecycle: extract features, communicate with integrators, deprecate the npm package, then delete source code from the monorepo after three months.

## Notes

`TIB-0002` requires README/public docs updates warning that installing `morpho-sdk` alongside incompatible direct versions of `blue-sdk`, `blue-sdk-viem`, or `morpho-ts` is unsupported because duplicate runtime copies can break `instanceof` and other identity-sensitive checks.

It also adds `@morpho-org/morpho-sdk/utils` as the single public utility import path for every public `morpho-ts` utility, plus selected utility/helper exports from `morpho-sdk`, `blue-sdk`, and `blue-sdk-viem`.

## Validation

- `git diff --check`
- Tests not run; staged files are documentation/non-code only.

### DIFF
--- a/docs/tibs/TIB-0002-consolidate-sdk-packages.md
+++ b/docs/tibs/TIB-0002-consolidate-sdk-packages.md
@@ -95,7 +95,7 @@ The dependency graph shows that several target packages are still consumed indep
 | `@morpho-org/blue-sdk-viem` | Yes. Runtime use in `liquidity-sdk-viem`, `liquidation-sdk-viem`, `migration-sdk-viem`, and `morpho-sdk`. | Keep as a maintained viem integration package. `morpho-sdk` depends on it directly and re-exports every fetcher; other packages keep their existing peer dependency model. |
 | `@morpho-org/bundler-sdk-viem` | Yes. Runtime use in `migration-sdk-viem` and `morpho-sdk`. | Move selected action/bundle APIs into `morpho-sdk`; migrate or deprecate `migration-sdk-viem` before removal. |
 | `@morpho-org/migration-sdk-viem` | No maintained monorepo runtime consumers outside its own package, tests, and documentation. | Deprecate as a public package. Retain no migration APIs under `morpho-sdk`; document that migration SDK workflows are no longer a supported public SDK surface. |
-| `@morpho-org/simulation-sdk` | Yes. Runtime use in `bundler-sdk-viem`, `liquidity-sdk-viem`, `migration-sdk-viem`, and `morpho-sdk`. | Deprecate broad simulation engine, but first move required constants/types/helpers into owning packages. |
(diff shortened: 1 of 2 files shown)


Foulks-Plb:
This PR introduces a numbered scheme (TIB-0002-..., TIB-0003-...).
I’m not opposed to it, but we need to follow/update what the project requires.

docs/tibs/TIB-0002-consolidate-sdk-packages.md
```diff
@@ -0,0 +1,591 @@
...
+  the thin utility exports from `blue-sdk` and `blue-sdk-viem`. This is the canonical utility import
+  path.
+- `@morpho-org/morpho-sdk/utils/time`: mirror `packages/morpho-ts/src/time/index.ts`.
+- `@morpho-org/morpho-sdk/utils/format`: mirror `packages/morpho-ts/src/format/index.ts`.
+- `@morpho-org/morpho-sdk/augment`: expose the optional side-effecting module augmentation
+  entrypoint. This is the only side-effecting public subpath.
+
+Example imports after consolidation:
+
+```ts
```
Foulks-Plb, line 154: **Redundant export paths for `morpho-ts/time` and `morpho-ts/format` symbols.**

`@morpho-org/morpho-sdk/utils` is defined here as re-exporting *every* public export from `morpho-ts`, explicitly including `time` and `format`. The next two bullets then also expose `@morpho-org/morpho-sdk/utils/time` and `@morpho-org/morpho-sdk/utils/format` as dedicated subpaths.

The example imports a few lines above confirm the overlap — `format` is imported from `/utils` while `formatLongString` is imported from `/utils/format`:

```ts
import { Time, deepFreeze, format, safeParseUnits } from "@morpho-org/morpho-sdk/utils";
import { formatLongString } from "@morpho-org/morpho-sdk/utils/format";
```

So every symbol in `morpho-ts/time` and `morpho-ts/format` has two valid public import paths. This contradicts the same section's stated goal ("Minimize package subpaths") and the "single source of truth" principle in `CLAUDE.md` §1. It also asymmetrically privileges `time` and `format` over equally cohesive `morpho-ts` domains like `urls` and `types`, which get no dedicated subpath — the "cohesive domain" criterion isn't applied consistently.

Two clean options:

- **Drop `/utils/time` and `/utils/format`.** Everything goes through `/utils`. Tree-shaking handles unused symbols as long as `sideEffects` stays correctly scoped.
- **Keep the grouped subpaths, but remove their contents from `/utils`.** `Time` is only reachable via `/utils/time`, `formatLongString` only via `/utils/format`. One symbol, one path.

Either is defensible — exposing both at once is the option to avoid.

Foulks-Plb (approved):

PR #669, opened 2026-05-13
refactor(morpho-sdk): Extract morpho-sdk bundler action encoding

Rubilmax:
## Summary

- Extract the Bundler3 action encoding surface required by `morpho-sdk` into `packages/morpho-sdk`.
- Remove `@morpho-org/bundler-sdk-viem` from `morpho-sdk` dependencies and lockfile importer data.
- Keep `migration-sdk-viem` unchanged, including its existing `bundler-sdk-viem` wiring, because it is slated for deprecation.
- Move the new `BundlerErrors` namespace into `packages/morpho-sdk/src/types/error.ts` and re-export it from the local bundler action surface.

Fixes SDK-186

## Validation

- `pnpm --filter @morpho-org/morpho-sdk exec tsc --noEmit -p tsconfig.build.esm.json`
- `pnpm --filter @morpho-org/morpho-sdk exec tsc --noEmit -p tsconfig.build.cjs.json`
- `pnpm exec vitest --root . --project morpho-sdk packages/morpho-sdk/src/actions/bundler.test.ts --run`
- `pnpm lint`
- `git diff --check`


Rubilmax (commented):
LGTM

Rubilmax:
@codex

packages/morpho-sdk/BUNDLER3.md
```diff
@@ -6,7 +6,7 @@ This document describes how **Bundler3** and **GeneralAdapter1** are used in the
 
 Bundler3 is a _multicall_ contract specific to the Morpho ecosystem. It takes as input a **list of typed actions** (ERC20 transfers, permit, permit2, native wrapping, ERC-4626 deposits, Morpho Blue calls, reallocations…) and executes them **atomically in a single transaction**.
 
-Instead of exposing the user directly to target contracts (ERC-4626 vault, Morpho Blue, PublicAllocator, WETH…), the SDK encodes a bundle via [`BundlerAction.encodeBundle(chainId, actions)`](src/actions/vaultV1/deposit.ts#L146) from `@morpho-org/bundler-sdk-viem`. The `to` of the resulting transaction is **always** the Bundler3 address for the target chain.
+Instead of exposing the user directly to target contracts (ERC-4626 vault, Morpho Blue, PublicAllocator, WETH…), the SDK encodes a bundle via its local [`BundlerAction.encodeBundle(chainId, actions)`](src/actions/bundler.ts). The `to` of the resulting transaction is **always** the Bundler3 address for the target chain.
```
Rubilmax: **[HIGH]** Broken markdown link — sentence points to `src/actions/bundler.ts` but the new file is at `src/bundler/actions.ts`.

_Suggestion:_ update the inline link to `src/bundler/actions.ts`. Sweep `ARCHITECTURE.md` for any other stale references to the old path.

packages/morpho-sdk/package.json
```diff
@@ -49,6 +49,16 @@
         "types": "./lib/esm/index.d.ts",
         "import": "./lib/esm/index.js",
         "require": "./lib/cjs/index.js"
+      },
+      "./abis": {
```
Rubilmax, line 70: **[MEDIUM]** The new `./abis` subpath publishes 9 ABIs morpho-sdk doesn't consume (`universalRewardsDistributorAbi`, `ethereumGeneralAdapter1Abi`, `paraswapAdapterAbi`, `erc20WrapperAdapterAbi`, the four migration adapter ABIs). Only `bundler3Abi`, `coreAdapterAbi`, `generalAdapter1Abi` are used. Per §4/§1, morpho-sdk's reason to exist is transaction builders, not an ABI registry.

_Suggestion:_ trim `src/abis.ts` and `./abis` to the three ABIs the package actually uses, OR split the registry into a leaf package both `morpho-sdk` and `bundler-sdk-viem` can depend on.

packages/morpho-sdk/src/abis.ts
```diff
@@ -0,0 +1,2070 @@
+/** ABI for the Universal Rewards Distributor contract. */
```
Rubilmax: **[CRITICAL]** The entire file is a near-verbatim duplicate of `packages/bundler-sdk-viem/src/abis.ts`. The same 12 ABI literals now live in two places. Violates §1 (single source of truth: "one place per ABI ... Duplication is a refactor, not a feature") and §7 (pinned ABIs, no address drift). The next contract upgrade can land in one copy and silently miss the other.

_Suggestion:_ pick one canonical home — given the branch is `drop-bundler-sdk-viem`, the long-term direction is `morpho-sdk/abis`. Delete `bundler-sdk-viem/src/abis.ts` in the same PR and re-export the ABIs from `@morpho-org/morpho-sdk/abis` so consumers (and `bundler-sdk-viem` itself, until it's retired) read from one source.

packages/morpho-sdk/src/abis.ts
```diff
@@ -0,0 +1,2070 @@
...
+export const universalRewardsDistributorAbi = [
```
Rubilmax: **[LOW]** 9 of the 12 ABIs aren't referenced anywhere in `packages/morpho-sdk/src/`. Either keep with a stated reason (downstream consumer support; document in JSDoc/AGENTS.md), or remove — they widen public surface morpho-sdk must support across semver without serving morpho-sdk's stated mission.

_Suggestion:_ same fix as the package.json `./abis` finding — addressing either drops both.

packages/morpho-sdk/src/bundler/actions.test.ts
```diff
@@ -0,0 +1,759 @@
+import {
```
Rubilmax, line 1: **[LOW]** §5 recommends property-based tests on calldata encoders (`fast-check`). The new `BundlerAction.encode` has an enumerable discriminated-union input space well-suited to property testing (no-throw invariants, round-trip stability for self-describing actions). Not a blocker — a natural place to seed the convention.

_Suggestion:_ add a `fast-check` property exercising arbitrary `Action` inputs to `BundlerAction.encode`.

packages/morpho-sdk/src/bundler/actions.test.ts
```diff
@@ -0,0 +1,759 @@
...
+import { bundler3Abi, coreAdapterAbi, generalAdapter1Abi } from "../abis.js";
+import {
+  type Action,
+  BundlerAction,
+  type BundlerCall,
+  BundlerErrors,
+  type Permit2PermitSingle,
+} from "./index.js";
+
+const legacyBundlerActionModulePath =
```
Rubilmax: **[HIGH]** The "bundler-sdk-viem parity" suite (lines 193-272) imports the legacy `BundlerAction` via a relative filesystem path (`../../../bundler-sdk-viem/src/BundlerAction.ts`) using a dynamic string-variable `await import(...)` and an `as { readonly BundlerAction: typeof BundlerAction }` cast. `@morpho-org/bundler-sdk-viem` is no longer in `dependencies` or `devDependencies` of morpho-sdk. Violates §2.5 (deep imports across packages bypass `src/index.ts`). The dynamic specifier defeats TS+Biome static analysis; the cast type-checks against the NEW `BundlerAction` shape, so if the legacy surface drifts the test stays green. When `bundler-sdk-viem` is removed (the branch goal), this test breaks only at runtime.

_Suggestion:_ declare `"@morpho-org/bundler-sdk-viem": "workspace:^"` in `devDependencies` and import with a literal package specifier so TypeScript resolves the real symbol, OR remove the parity suite once parity has been verified and rely on inline snapshots.

packages/morpho-sdk/src/bundler/actions.test.ts
```diff
@@ -0,0 +1,759 @@
...
+            1n,
+            [{ marketParams: market, amount: 2n }],
+            market,
+          ),
+        ),
+      );
+    });
+  });
+
+  test("encodeBundle", () => {
```
Rubilmax, line 637: **[MEDIUM]** The standalone `encodeBundle` test pins value accounting for `nativeTransfer` actions but never asserts the matching contract for `reallocateTo`: the fee must NOT be summed into `encodeBundle`'s `tx.value` because the actions layer overrides `value` separately (`marketV1/borrow.ts:111`, `supplyCollateralBorrow.ts:212`). Without this regression test, a future contributor "fixing" `encodeBundle` to include `reallocateTo.fee` would silently double-count at the action layer.

_Suggestion:_ add either `expect(BundlerAction.encodeBundle(chainId, [{ type: "reallocateTo", args: [vault, 5n, [...], market] }]).value).toBe(0n)` to pin the current encoder contract, or a `marketV1Borrow({ reallocations })` snapshot pinning `tx.value === reallocationFee` end-to-end.

packages/morpho-sdk/src/bundler/actions.test.ts
```diff
@@ -0,0 +1,759 @@
...
+          "data": "0x3790767d000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000030000000000000000000000000000000000000000000000000000000000000005",
+          "skipRevert": false,
+          "to": "0x0000000000000000000000000000000000000004",
+          "value": 0n,
+        },
+      ]
+    `);
+  });
+
+  test("encode", () => {
```
Rubilmax: **[LOW]** The `encode` smoke test only asserts `expect(() => BundlerAction.encode(chainId, action)).not.toThrow()` per action type. Once the parity suite is removed (per the related finding) the `encode` switch loses positive-assertion coverage for dispatch wiring — a regression that returns `[]` from one branch would still pass.

_Suggestion:_ replace the `not.toThrow` loop with per-branch inline-snapshot assertions on the encoded `BundlerCall[]`, or rely on the dedicated `test("nativeTransfer")`/`test("erc20Transfer")`/etc. blocks and remove the redundant `encode` smoke test.

Rubilmax (commented):
Automated `/pr-review-local` findings, posted as inline comments.

| Severity | Count |
|----------|-------|
| Critical | 1 |
| High | 4 |
| Medium | 11 |
| Low | 4 |

9 personas fired (baseline + `ci-release-security`). 0 agent failures. Deduplicated against branch base `extract-public-reallocation` (merge-base `1b736238`). Scope filter dropped 2 findings on files outside the diff.

Guard against `/pr-review-local`'s known limits: messages may drift between reruns (LLM nondeterminism); the line of a finding is always within the changed-line window of the file, but the *exact* line can lag the symbol by a few lines on dense files.

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1063 @@
...
+} from "viem";
+import { bundler3Abi, coreAdapterAbi, generalAdapter1Abi } from "../abis.js";
+import { BundlerErrors } from "../types/error.js";
+import type {
+  Action,
+  InputReallocation,
+  Permit2PermitSingle,
+} from "./types.js";
+
+export { BundlerErrors } from "../types/error.js";
```
Rubilmax: **[MEDIUM]** `BundlerErrors` is exposed on two public morpho-sdk surfaces. `bundler/actions.ts` re-exports it from `../types/error.js`, AND it's already reachable via the root entrypoint through `./types/index.js`. Per §4, each public symbol has one canonical path.

_Suggestion:_ drop the `export { BundlerErrors }` line at `bundler/actions.ts:27`, OR move the `BundlerErrors` namespace into `bundler/errors.ts` (bundler-scoped) and let the root barrel reach it transitively via `export * from "./bundler/index.js"`. Pick one.

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1063 @@
...
+
+/**
+ * Encodes the Bundler3 action subset used by `morpho-sdk` transaction builders.
+ *
+ * @remarks
+ * The namespace covers only the Bundler3 actions required by `morpho-sdk`.
+ * It does not expose operation population or simulation helpers from
+ * `@morpho-org/bundler-sdk-viem`.
+ */
+export namespace BundlerAction {
```
Rubilmax, line 113: **[MEDIUM]** All 16 `@example` blocks in the new `BundlerAction` namespace use bare placeholder addresses like `"0x0000000000000000000000000000000000000001"`. `docs/jsdoc-style.md` explicitly prohibits this: "Use named protocol constants and fixture addresses; never embed `0x…` placeholders." The reference exemplar `actions/marketV1/borrow.ts` uses named symbols (`marketParams`, `borrower`).

_Suggestion:_ replace zero-suffix hex literals with named consts (`loanToken`, `collateralToken`, `recipient`, `vault`, `marketParams`) declared at the top of each example, matching the existing pattern.

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1063 @@
...
+   *     args: [
+   *       "0x0000000000000000000000000000000000000001",
+   *       "0x4A6c312ec70E8747a587EE860a0353cd42Be0aE0",
+   *       1_000000000000000000n,
+   *     ],
+   *   },
+   * ]);
+   * ```
+   */
+  export function encodeBundle(chainId: number, actions: Action[]) {
```
Rubilmax, line 154: **[HIGH]** `BundlerAction.encodeBundle` sums only `nativeTransfer` action amounts into `value` (lines 114-128); it ignores `reallocateTo` `fee` (set as `BundlerCall.value` at line 1011). Internal callers in the action layer override `tx.value` after the call, but `encodeBundle` is now a PUBLIC export via the `./bundler` subpath and the root entrypoint. An integrator calling `encodeBundle([reallocateTo, morphoBorrow])` directly gets `value: 0n`; the bundler3 `multicall` is `payable` and the inner call demands `fee` — the transaction reverts (or, depending on bundler3 reentry semantics, draws from previous reentry state). The JSDoc claims "native value inferred from native-transfer actions" without warning about the gap.

_Suggestion:_ sum every encoded `BundlerCall.value` inside `encodeBundle` (after the `flatMap`) so the returned `value` always matches what bundler3 will demand, removing the footgun. If that would double-count with the action layer's manual override, drop the manual override and rely on `encodeBundle` exclusively. Update `BUNDLER3.md` line 69 ("native fees are aggregated into `tx.value`") to match.

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1063 @@
...
+
+    let value = 0n;
+
+    for (const { type, args } of actions) {
+      if (type !== "nativeTransfer") continue;
+
+      const [owner, recipient, amount] = args;
+
+      if (
+        owner !== bundler3 &&
```
Rubilmax: **[HIGH]** Address comparisons in `encodeBundle` (122-128) and `nativeTransfer` (302, 304) use strict `===` against checksum-cased registry addresses from `getChainAddresses`. viem's `Address` type does not enforce checksum at runtime. A caller passing lowercased `bundler3`/`generalAdapter1` will silently bypass the bundler-detection logic — `value` will miscount and `nativeTransfer` will emit a plain ETH transfer to what looks-like-but-isn't the registered bundler. The legacy `bundler-sdk-viem` has the same flaw (parity passes), but the extraction puts the code onto morpho-sdk's stable public surface.

_Suggestion:_ replace `===`/`!==` with `isAddressEqual` from viem on every address comparison in this file (and apply the same fix to bundler-sdk-viem to keep behaviour aligned until it is retired).

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1063 @@
...
+  }
+
+  /**
+   * Encodes a single supported Bundler3 action into one or more low-level
+   * Bundler3 calls.
+   *
+   * @param chainId - Chain where the action will execute.
+   * @param action - Bundler3 action to encode.
+   * @returns Encoded Bundler3 calls.
+   * @throws {@link BundlerErrors.MissingSignature} when a signature action is unsigned.
```
Rubilmax: **[LOW]** `@throws {@link BundlerErrors.MissingSignature} when ...` (lines 152, 153, 490, 966) uses the markdown-link form `{@link ...}`, diverging from the canonical `docs/jsdoc-style.md` format (`@throws {ErrorClass} when <condition>.`) and from the reference exemplar `actions/marketV1/borrow.ts`. Either form is functional; consistency within the package favours the plain form.

_Suggestion:_ replace `{@link BundlerErrors.X}` with `{BundlerErrors.X}` across the four sites for consistency.

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1063 @@
...
+      case "morphoWithdrawCollateral": {
+        return BundlerAction.morphoWithdrawCollateral(chainId, ...args);
+      }
+      case "reallocateTo": {
+        return BundlerAction.publicAllocatorReallocateTo(chainId, ...args);
+      }
+      case "wrapNative": {
+        return BundlerAction.wrapNative(chainId, ...args);
+      }
+    }
```
Rubilmax, line 301: **[MEDIUM]** `BundlerAction.encode`'s `switch (type)` has no default branch and no `assertNever` after the switch. The function is declared `BundlerCall[]` but if a future contributor adds a new `ActionType` to `bundler/types.ts` and forgets a case, the function silently returns `undefined`. Biome's `useExhaustiveSwitchCases:error` catches this at lint time but is bypassable.

_Suggestion:_ add an `assertNever(type)` helper after the switch, or a `default:` branch that throws `BundlerErrors.UnexpectedAction(type, chainId)`, so the failure is loud at runtime even if lint is skipped.

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1063 @@
...
+   *
+   * const calls = BundlerAction.nativeTransfer(
+   *   1,
+   *   "0x0000000000000000000000000000000000000001",
+   *   "0x0000000000000000000000000000000000000002",
+   *   1_000000000000000000n,
+   * );
+   * ```
+   */
+  // biome-ignore lint/complexity/useMaxParams: TODO refactor to ≤2 params
```
Rubilmax, line 392: **[LOW]** 11 `// biome-ignore lint/complexity/useMaxParams: TODO refactor to ≤2 params` annotations across lines 290, 353, 398, 451, 512, 572, 623, 677, 738, 802, 867, 929, 989, 1038. Per §2.1's intent (no orphan TODO suppressions; suppressions carry an issue link and deletion plan), these orphan TODOs widen the suppression footprint indefinitely.

_Suggestion:_ either attach an issue link with a deletion plan to each, OR refactor the signatures to take options-bag parameters (one or two args max), which would also align with §3's discriminated-unions preference at the function-input layer.

packages/morpho-sdk/src/types/error.ts
```diff
@@ -1,6 +1,77 @@
...
 import type { Address } from "viem";
 
+/**
+ * Typed errors thrown while encoding supported Bundler3 actions.
+ *
+ * @remarks
+ * Import these classes through `@morpho-org/morpho-sdk/bundler` when handling
+ * failures from `BundlerAction`.
+ */
+export namespace BundlerErrors {
```
Rubilmax, line 11: **[MEDIUM]** The `BundlerErrors` namespace duplicates the class identities in `packages/bundler-sdk-viem/src/errors.ts`. Per §1 ("one place per error class") and §3 (errors are public API integrators pattern-match on via class identity), having two identities for the same conceptual error means `instanceof BundlerErrors.MissingSignature` from one package silently misses errors thrown from the other.

_Suggestion:_ in the same PR, delete the duplicate namespace from `bundler-sdk-viem/src/errors.ts` and re-export `BundlerErrors` from `@morpho-org/morpho-sdk` (or `/bundler`). bundler-sdk-viem-specific errors (`Bundle`, `MissingSwapData`, `MissingSkimHandler`, `UnskimedToken`) stay where they are.

packages/morpho-sdk/src/types/error.ts
```diff
@@ -1,6 +1,77 @@
...
+   * import { BundlerErrors } from "@morpho-org/morpho-sdk/bundler";
+   *
+   * if (error instanceof BundlerErrors.MissingSignature) {
+   *   // Attach the missing permit or Permit2 signature, then encode again.
+   * }
+   * ```
+   */
+  export class MissingSignature extends Error {
+    constructor() {
+      super("missing signature");
```
Rubilmax, line 27: **[MEDIUM]** `BundlerErrors.MissingSignature` throws `"missing signature"` — terse blame, not the instructive style the rest of `error.ts` uses (compare `BorrowExceedsSafeLtvError`: "Reduce borrow or increase collateral."). Violates §3 ("Error messages read like instructions, not blame").

_Suggestion:_ reword to identify which action and which asset is unsigned, with a recovery hint, e.g. `Signature is missing for this bundled action. Attach the permit / Permit2 signature on the matching requirement before encoding.`

packages/morpho-sdk/src/types/error.ts
```diff
@@ -1,6 +1,77 @@
...
+   * }
+   * ```
+   */
+  export class UnexpectedAction extends Error {
+    /**
+     * @param type - Unsupported Bundler3 action discriminator or name.
+     * @param chainId - Chain where the action was requested.
+     */
+    constructor(type: string, chainId: number) {
+      super(`unexpected action "${type}" on chain "${chainId}"`);
```
Rubilmax, line 49: **[MEDIUM]** `BundlerErrors.UnexpectedAction` message `unexpected action "${type}" on chain "${chainId}"` is terse and lowercase. Same §3 concern as MissingSignature.

_Suggestion:_ rewrite as `Bundler3 action "${type}" is not supported on chain ${chainId}. Remove the action or run the bundle on a chain where it is configured.`

packages/morpho-sdk/src/types/error.ts
```diff
@@ -1,6 +1,77 @@
...
+   * @example
+   * ```ts
+   * import { BundlerErrors } from "@morpho-org/morpho-sdk/bundler";
+   *
+   * if (error instanceof BundlerErrors.UnexpectedSignature) {
+   *   // Ask the user to sign a payload for the expected spender.
+   * }
+   * ```
+   */
+  export class UnexpectedSignature extends Error {
```
Rubilmax: **[MEDIUM]** `BundlerErrors.UnexpectedSignature` is defined, exported, JSDoc'd with an integrator-handler example, and constructed in a test, but no SDK source path ever throws it (grep confirms zero `throw new BundlerErrors.UnexpectedSignature(...)` sites). The legacy `bundler-sdk-viem` raised it for permit/permit2 signature-spender validation (signed `spender !== generalAdapter1`). The new `BundlerAction.approve2` overwrites `permitSingle.spender` with `generalAdapter1` (line 538) without checking the signed value, so the runtime guard is silently dropped. Dead public surface and a missing defense at the encoder boundary.

_Suggestion:_ pick one: (a) wire a defensive check at the encoder — assert `permitSingle.spender === undefined || isAddressEqual(permitSingle.spender, generalAdapter1)` (and the same for any permit/morpho-authorization paths) and throw `UnexpectedSignature` on mismatch, so user-constructed payloads with a wrong spender fail loudly instead of being silently overwritten; or (b) remove the class and its JSDoc example.

packages/morpho-sdk/src/types/error.ts
```diff
@@ -1,6 +1,77 @@
...
+    /**
+     * @param spender - Address unexpectedly authorized by the signature.
+     */
+    constructor(spender: Address) {
+      super(`unexpected signature consumer "${spender}"`);
```
Rubilmax: **[MEDIUM]** `BundlerErrors.UnexpectedSignature` message `unexpected signature consumer "${spender}"` is terse and blame-shaped. Same §3 concern.

_Suggestion:_ rewrite as `Signature authorizes unexpected spender ${spender}. Re-sign the payload for the expected GeneralAdapter1 address.`

packages/morpho-sdk/package.json
Rubilmax: It is to keep compatibility with existing sdks

packages/morpho-sdk/src/abis.ts
Rubilmax: we are deprecating the other package dumbass

packages/morpho-sdk/src/abis.ts
Rubilmax: downstream consumer support

packages/morpho-sdk/src/bundler/actions.ts
Rubilmax: Keep compatibility, ignored

packages/morpho-sdk/src/types/error.ts
Rubilmax: Keep error messages identical to avoid breaking consumption

packages/morpho-sdk/src/types/error.ts
Rubilmax: Keep error messages identical to avoid breaking consumption

packages/morpho-sdk/src/bundler/actions.ts
Rubilmax: lint is never skipped

packages/morpho-sdk/src/types/error.ts
Rubilmax: Keep error messages identical to avoid breaking consumption

packages/morpho-sdk/src/types/error.ts
Rubilmax: Expected during the process of deprecation

.changeset/drop-bundler-sdk-viem.md
```diff
@@ -0,0 +1,7 @@
+---
+"@morpho-org/morpho-sdk": minor
```
Rubilmax, line 2: **[HIGH]** Missing patch changeset for `@morpho-org/bundler-sdk-viem`. This PR ships a real behavior fix to its published `BundlerAction` surface in `packages/bundler-sdk-viem/src/BundlerAction.ts` (four `===`/`!==` → `isAddressEqual` conversions at the encodeBundle native-value detector, the nativeTransfer recipient/owner short-circuits, and the authorization-target guard). With the old `===`, integrators who passed lowercase/non-checksummed addresses to `bundler-sdk-viem` would silently under-count `tx.value` (encodeBundle never added the amount) and skip the bundler3 short-circuit in `nativeTransfer`. Per AGENTS.md §7, every semver-relevant change to published package source ships with a changeset.

_Suggestion: add `.changeset/<slug>.md` with `"@morpho-org/bundler-sdk-viem": patch` and a one-line note about the checksum-insensitive comparison fix._

packages/liquidity-sdk-viem/src/loader.test.ts
```diff
@@ -43,7 +43,7 @@ const vault = "0x0000000000000000000000000000000000000aaa";
 const loanToken = "0x0000000000000000000000000000000000000101";
 const targetCollateralToken = "0x0000000000000000000000000000000000000202";
 const sourceCollateralToken = "0x0000000000000000000000000000000000000303";
-const owner = "0x0000000000000000000000000000000000000bbb";
+const owner = "0x0000000000000000000000000000000000000bBB";
```
Rubilmax, line 46: **[MEDIUM]** Unrelated change bundled in this PR. Single owner-address checksum fix (`0x...bbb` → `0x...bBB`) is tangential to dropping `@morpho-org/bundler-sdk-viem`. AGENTS.md §8 requires one concern per PR.

_Suggestion: split into a separate PR._

packages/morpho-sdk/src/abis.ts
Rubilmax: **[HIGH]** ABI source-of-truth duplication. All 12 exported ABIs in this new file (`universalRewardsDistributorAbi`, `bundler3Abi`, `coreAdapterAbi`, `generalAdapter1Abi`, `ethereumGeneralAdapter1Abi`, `paraswapAdapterAbi`, `erc20WrapperAdapterAbi`, `aaveV{2,3}MigrationAdapterAbi`, `aaveV3OptimizerMigrationAdapterAbi`, `compoundV{2,3}MigrationAdapterAbi`) already exist in `packages/bundler-sdk-viem/src/abis.ts`. Verified byte-identical except for JSDoc comments. AGENTS.md §1 Modularity: "Single source of truth per concept: one place per ABI". Two copies will drift on the next protocol upgrade and silently desync calldata encoding.

_Suggestion: keep ABIs in one location only. Either (a) re-export the canonical ABIs from `@morpho-org/blue-sdk-viem`/`@morpho-org/bundler-sdk-viem` (the latter is already a devDep, so a runtime import would re-introduce the dep — which contradicts the PR goal), or (b) move the shared ABI source into a common low-level package both depend on (e.g. a new `@morpho-org/abis` or fold into `@morpho-org/blue-sdk`), or (c) flip the direction: make this file the canonical source and have `bundler-sdk-viem` re-export from `morpho-sdk` in the same PR (delete the duplicated copy in `bundler-sdk-viem/src/abis.ts`)._

packages/morpho-sdk/src/abis.ts
Rubilmax: **[MEDIUM]** 9 of 12 newly added ABIs are unused by `morpho-sdk` source. The encoder in `src/bundler/actions.ts` only references `bundler3Abi`, `coreAdapterAbi`, `generalAdapter1Abi`. The rest (`ethereumGeneralAdapter1Abi`, `paraswapAdapterAbi`, `erc20WrapperAdapterAbi`, `aaveV{2,3}MigrationAdapterAbi`, `aaveV3OptimizerMigrationAdapterAbi`, `compoundV{2,3}MigrationAdapterAbi`, `universalRewardsDistributorAbi`) widen the public surface for no current caller. MISSION.md values: "Delete before adding".

_Suggestion: drop the 9 unused ABIs until a real consumer in this package requires them._

packages/morpho-sdk/src/actions/marketV1/borrow.ts
```diff
@@ -106,10 +106,7 @@ export const marketV1Borrow = ({
     args: [marketParams, amount, 0n, minSharePrice, receiver, false],
   });
 
-  let tx = {
-    ...BundlerAction.encodeBundle(chainId, actions),
-    value: reallocationFee,
-  };
+  let tx = BundlerAction.encodeBundle(chainId, actions);
```
Rubilmax, line 109: **[HIGH]** Dropped `value: reallocationFee` override (`let tx = BundlerAction.encodeBundle(chainId, actions)` replaces `{...BundlerAction.encodeBundle(...), value: reallocationFee}`), but `borrow.test.ts` has no test covering the `reallocations` path — all tests assert `tx.value === 0n` for the no-reallocations path. Correctness of `tx.value === reallocationFee` is now solely guarded by the legacy fork test at `test/actions/marketV1/reallocations.test.ts:275-276`. Per AGENTS.md §5 (security invariants are tests), behavior previously enforced by the SDK source should have a colocated unit test.

_Suggestion: add a colocated unit test for `marketV1Borrow({ ..., reallocations: [...] })` pinning `tx.value` to the summed reallocation fee._

packages/morpho-sdk/src/actions/marketV1/supplyCollateralBorrow.ts
```diff
@@ -205,10 +207,7 @@ export const marketV1SupplyCollateralBorrow = ({
     args: [marketParams, borrowAmount, 0n, minSharePrice, receiver, false],
   });
 
-  let tx = {
-    ...BundlerAction.encodeBundle(chainId, actions),
-    value: (nativeAmount ?? 0n) + reallocationFee,
-  };
+  let tx = BundlerAction.encodeBundle(chainId, actions);
```
Rubilmax, line 210: **[HIGH]** Dropped `value: (nativeAmount ?? 0n) + reallocationFee` override. Existing colocated tests pin `tx.value === nativeAmount` (no-reallocations path) but no test exercises the combined `nativeAmount + reallocations` path. The combined-path equality `tx.value === nativeAmount + reallocationFee` is now derived entirely by the new `encodeBundle` algorithm and protected only by `test/actions/marketV1/reallocations.test.ts:578-585` (fork-only).

_Suggestion: add a colocated unit test combining a positive `nativeAmount` and a non-empty `reallocations` array, asserting `tx.value === nativeAmount + sum(reallocations.fee)`._

packages/morpho-sdk/src/actions/vaultV1/deposit.ts
```diff
@@ -176,10 +178,6 @@ export const vaultV1Deposit = ({
 
   let tx = BundlerAction.encodeBundle(chainId, actions);
```
Rubilmax, line 179: **[HIGH]** Dropped the `if (nativeAmount) { tx = { ...tx, value: nativeAmount }; }` block. Existing colocated tests in `deposit.test.ts` all assert `tx.value === 0n` (ERC-20-only paths); no test pins `tx.value === nativeAmount` for the native-deposit path. A regression in the new `encodeBundle` derivation for `nativeTransfer(bundler3, generalAdapter1, nativeAmount)` would silently produce wrong `msg.value` on the wrapping path.

_Suggestion: add a colocated unit test for `vaultV1Deposit({ args: { amount: 0n, nativeAmount, ... } })` against a wNative vault asserting `tx.value === nativeAmount`. Mirror the existing `marketV1/supplyCollateralBorrow.test.ts:81,109` pattern._

packages/morpho-sdk/src/actions/vaultV2/deposit.ts
```diff
@@ -176,10 +178,6 @@ export const vaultV2Deposit = ({
 
   let tx = BundlerAction.encodeBundle(chainId, actions);
```
Rubilmax, line 179: **[HIGH]** Same as `vaultV1/deposit.ts` — dropped `value: nativeAmount` override with no test pinning the equality. Existing tests cover only the ERC-20 path.

_Suggestion: add a colocated unit test for the wNative + nativeAmount path asserting `tx.value === nativeAmount`._

packages/morpho-sdk/src/bundler/actions.test.ts
```diff
@@ -0,0 +1,1251 @@
...
+        for (const call of calls) {
+          expect(call.to).toMatch(/^0x[0-9a-f]{40}$/i);
+          expect(call.data).toMatch(/^0x(?:[0-9a-f]{2})*$/i);
+          expect(call.value >= 0n).toBe(true);
+          expect(call.callbackHash).toMatch(/^0x[0-9a-f]{64}$/i);
+        }
+      }),
+      { numRuns: 100, seed: 669 },
+    );
+  });
```
Rubilmax, line 621: **[MEDIUM]** Property-based test (`encode accepts arbitrary supported actions deterministically`, 100 runs, seed 669) only asserts shape on individual encoded calls (`call.value >= 0n`, hex shape). It does NOT exercise `encodeBundle`'s `tx.value` derivation invariant — the load-bearing logic the production action layers now rely on.

_Suggestion: add a property test that for any sequence of `Action[]`, `tx.value` equals the sum of `nativeTransfer(*, bundler3, x)` pre-fundings (capped by downstream consumption) plus the excess of downstream `call.value` over `availableBundlerValue`._

packages/morpho-sdk/src/bundler/actions.test.ts
```diff
@@ -0,0 +1,1251 @@
...
+      data: tx.data,
+    });
+
+    expect(decoded.functionName).toBe("multicall");
+    const calls = decoded.args[0] ?? [];
+    expect(calls).toHaveLength(1);
+    expect(calls[0]?.value).toBe(5n);
+  });
+
+  test("encodeBundle uses native transfers to Bundler3 before adding call value", () => {
```
Rubilmax, line 767: **[MEDIUM]** Pre-funding tests cover only the case where pre-funded native equals the downstream call's `value` (5n vs 5n). The arithmetic branches in `encodeBundle` (lines 137-143 of `actions.ts`) — `call.value > availableBundlerValue` adds excess to `value`; otherwise drains `availableBundlerValue` — are not exercised under pre-funding > call.value, pre-funding < call.value across multiple value-carrying calls, or multiple chained pre-fundings drained by multiple downstream calls.

_Suggestion: add three explicit cases: (a) pre-fund 10n, single 5n call → tx.value = 10n; (b) pre-fund 5n, two 4n calls → tx.value = 8n (5 from pre-fund covers first 4, second 4 takes 1 from pre-fund + 3 new); (c) two pre-fundings 5n + 5n drained by 12n call → tx.value = 12n._

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1134 @@
...
+} from "viem";
+import { bundler3Abi, coreAdapterAbi, generalAdapter1Abi } from "../abis.js";
+import { BundlerErrors } from "../types/error.js";
+import type {
+  Action,
+  InputReallocation,
+  Permit2PermitSingle,
+} from "./types.js";
+
+export type {
```
Rubilmax: **[MEDIUM]** Double re-export of bundler types. This file re-exports `Action, ActionArgs, Actions, ActionType, Authorization, InputReallocation, Permit2PermitSingle, Permit2PermitSingleDetails` from `./types.js` (lines 28-37), and `bundler/index.ts` also does `export * from "./types.js"`. Two export paths to the same symbols on the public surface — refactor hazard.

_Suggestion: drop the `export type { ... } from "./types.js"` block in `actions.ts`; rely on the `bundler/index.ts` barrel._

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1134 @@
...
+  /** Whether Bundler3 should continue when the call reverts. */
+  readonly skipRevert: boolean;
+
+  /** Expected callback hash for calls that reenter Bundler3, or zero hash. */
+  readonly callbackHash: Hex;
+}
+
+const reenterAbiInputs = bundler3Abi.find(
+  (item) => item.type === "function" && item.name === "reenter",
+)!.inputs;
```
Rubilmax, line 50: **[MEDIUM]** Non-null assertion `bundler3Abi.find((item) => item.type === "function" && item.name === "reenter")!.inputs` will crash with `TypeError: Cannot read properties of undefined` at module load if the local `bundler3Abi` ever drifts (the duplication issue) and omits `reenter`. The error message would be hostile vs. a named, exported error class.

_Suggestion: replace the `!` with a typed assertion (e.g. throw `BundlerErrors.UnexpectedAction("reenter", -1)` or a new `LocalAbiInvariantError`) so an ABI desync surfaces with an actionable error message instead of a low-level TypeError._

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1134 @@
...
+   * ]);
+   * ```
+   */
+  export function encodeBundle(chainId: number, actions: Action[]) {
+    const {
+      bundler3: { bundler3, generalAdapter1 },
+    } = getChainAddresses(chainId);
+
+    let value = 0n;
+    let availableBundlerValue = 0n;
```
Rubilmax: **[MEDIUM]** Public footgun in `encodeBundle`. The pre-funding branch increments `value` and `availableBundlerValue` by `amount` whenever `nativeTransfer(*, bundler3, X)` is encountered with `owner !== bundler3 && owner !== generalAdapter1` — but no inner call is emitted (line 318 short-circuits). If an integrator constructs `[nativeTransfer(externalOwner, bundler3, hugeAmount)]` with no consuming downstream calls, the signed transaction sends `hugeAmount` to bundler3 with no inner consumption, leaving funds stranded (no refund path on bundler3). Morpho-sdk's own action layers never construct this shape, but `BundlerAction.encodeBundle` is publicly re-exported and now part of the SDK contract.

_Suggestion: at end-of-loop, throw a typed error when `availableBundlerValue > 0n` (residual unconsumed pre-funding), OR document the constraint in JSDoc with an explicit warning._

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1134 @@
...
+
+        // A transfer into Bundler3 emits no inner call; it pre-funds later
+        // value-carrying calls in the same multicall.
+        if (
+          !isAddressEqual(owner, bundler3) &&
+          !isAddressEqual(owner, generalAdapter1) &&
+          isAddressEqual(recipient, bundler3)
+        ) {
+          value += amount;
+          availableBundlerValue += amount;
```
Rubilmax: **[MEDIUM]** Algorithm is order-dependent in a way that can over-count `tx.value`. If a caller intermixes pre-funding actions AFTER value-carrying calls (e.g. `[reallocateTo(fee=10n), nativeTransfer(owner, bundler3, 30n)]`), the loop produces `tx.value = 40n` (the reallocateTo's 10n already added to `value`, then nativeTransfer's pre-funding adds 30n more to `value`), while the user's intent of 30n total would be hidden. The legacy `bundler-sdk-viem` encoder would produce `tx.value = 30n` here. All morpho-sdk action layers happen to place pre-funding first, so this is dormant — but as a public API the ordering contract should be documented or enforced.

_Suggestion: JSDoc `encodeBundle` to require pre-funding actions to precede their consumers, OR detect intermixed pre-funding and throw._

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1134 @@
...
+   *
+   * const recipient = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045";
+   *
+   * const calls = BundlerAction.encode(1, {
+   *   type: "wrapNative",
+   *   args: [1_000000000000000000n, recipient],
+   * });
+   * ```
+   */
+  export function encode(chainId: number, action: Action): BundlerCall[] {
```
Rubilmax, line 204: **[MEDIUM]** `encode` switch lacks an explicit `default` clause. Biome's `useExhaustiveSwitchCases` catches missing branches at compile time, but if a caller bypasses types (`{} as Action`, JSON-deserialized input), the function silently returns `undefined` and the `for (const call of BundlerAction.encode(...))` in `encodeBundle` throws `TypeError: undefined is not iterable`.

_Suggestion: add `default: { const _exhaustive: never = type; throw new BundlerErrors.UnexpectedAction(String(type), chainId); }`._

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1134 @@
...
+   *
+   * const calls = BundlerAction.nativeTransfer(
+   *   1,
+   *   sender,
+   *   recipient,
+   *   1_000000000000000000n,
+   * );
+   * ```
+   */
+  // biome-ignore lint/complexity/useMaxParams: TODO refactor to ≤2 params
```
Rubilmax, line 392: **[MEDIUM]** Thirteen `// biome-ignore lint/complexity/useMaxParams: TODO refactor to ≤2 params` suppressions on the new encoder exports (lines 306, 374, 422, 480, 546, 609, 663, 721, 787, 856, 926, 993, 1058, 1109). Biome `useMaxParams` is `error`-level. AGENTS.md §2 forbids `@ts-ignore` without an issue link and a deletion plan; treat repeated `biome-ignore TODO` similarly. Without an issue link, the TODO will rot.

_Suggestion: file a tracking issue and append the link (`TODO(#NNN)`) to each suppression, or refactor the offending signatures to single options-object form before merging._

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -0,0 +1,1134 @@
...
+      return [
+        {
+          to: generalAdapter1,
+          data: encodeFunctionData({
+            abi: coreAdapterAbi,
+            functionName: "nativeTransfer",
+            args: [recipient, amount],
+          }),
+          value: 0n,
+          skipRevert: false,
```
Rubilmax, line 416: **[MEDIUM]** When `owner === generalAdapter1`, `nativeTransfer` hardcodes `skipRevert: false` on the inner GA1 call, silently dropping the caller-supplied `skipRevert` parameter (declared at line 312). A caller invoking `BundlerAction.nativeTransfer(chainId, GA1, recipient, amount, true /* skipRevert */)` gets a hard-revert inner call. (This may be parity-preserving with `bundler-sdk-viem`, but on a freshly extracted public encoder the override deserves either honoring `skipRevert` or removing the parameter from this code path's type signature.)

_Suggestion: propagate the caller's `skipRevert` to the GA1-as-owner branch, or document the carved-out behaviour in the JSDoc._

packages/morpho-sdk/src/bundler/types.ts
```diff
@@ -0,0 +1,184 @@
...
+  readonly details: Permit2PermitSingleDetails;
+
+  /** Deadline timestamp for consuming the Permit2 signature. */
+  readonly sigDeadline: bigint;
+}
+
+/**
+ * Argument tuples for Bundler3 actions supported by `morpho-sdk`.
+ */
+export interface ActionArgs {
```
Rubilmax, line 67: **[MEDIUM]** `ActionArgs` interface keys (`nativeTransfer`, `erc20Transfer`, …, `wrapNative` — 14 entries) define tuple argument shapes and are part of the public discriminated union (`Action = Actions[ActionType]`). Per `docs/jsdoc-style.md` ("Class fields, getters, and methods each get their own JSDoc block"), each tuple key should carry a one-line description identifying the Bundler3 action and the meaning of each positional argument. Without per-field JSDoc, IDE autocomplete shows types only.

_Suggestion: add a one-line JSDoc on each key (`/** ERC-2612 permit signature targeting GeneralAdapter1. */`)._

packages/morpho-sdk/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from "./actions/index.js";
+export * from "./bundler/index.js";
```
Rubilmax: **[MEDIUM]** `export * from "./bundler/index.js"` pulls bundler low-level types onto the root public surface. Generic names like `Action`, `Actions`, `ActionArgs`, `ActionType`, `Authorization`, `InputReallocation` now co-exist with the SDK's existing high-level discriminators (`MarketV1BorrowAction`, `VaultV1DepositAction`, etc.) and the existing `MorphoAuthorizationAction`. `Action` is a very generic root-level name; combined with the dedicated `./bundler` subpath export the rationale for re-exporting from root weakens (consumers who need encoder types can already import from `@morpho-org/morpho-sdk/bundler`).

_Suggestion: drop the bundler re-export from `src/index.ts` and require explicit `import ... from "@morpho-org/morpho-sdk/bundler"` for the encoder surface; OR rename the generic types (e.g. `BundlerAction.Action`)._

packages/liquidity-sdk-viem/src/loader.test.ts
Rubilmax: Dont care noww

packages/morpho-sdk/src/types/error.ts
```diff
@@ -1,6 +1,56 @@
...
+   * @example
+   * ```ts
+   * import { BundlerErrors } from "@morpho-org/morpho-sdk";
+   *
+   * if (error instanceof BundlerErrors.MissingSignature) {
+   *   // Attach the missing permit or Permit2 signature, then encode again.
+   * }
+   * ```
+   */
+  export class MissingSignature extends Error {
```
Rubilmax, line 25: **[MEDIUM]** `BundlerErrors.MissingSignature` / `BundlerErrors.UnexpectedAction` are distinct class identities from the identically-named classes in `packages/bundler-sdk-viem/src/errors.ts`. Integrators currently using `bundler-sdk-viem` who upgrade to the morpho-sdk-only path and keep `error instanceof BundlerErrors.MissingSignature` checks (against the bundler-sdk-viem class) will silently miss errors thrown by the new morpho-sdk encoder. AGENTS.md §3: "One class per failure mode, exported, integrators pattern-match on it."

_Suggestion: either re-export the bundler-sdk-viem classes (defeats the purpose of dropping the dep) OR rename to clearly distinct names (e.g. `MorphoBundlerErrors.MissingSignature`) so the divergence is loud and integrators get a TypeScript error during migration._

packages/morpho-sdk/package.json
```diff
@@ -49,6 +51,16 @@
         "types": "./lib/esm/index.d.ts",
         "import": "./lib/esm/index.js",
         "require": "./lib/cjs/index.js"
+      },
```
Rubilmax, line 69: **[HIGH]** `publishConfig.exports` adds subpath exports `./abis` and `./bundler` but there is no root-level `exports` or `typesVersions` mapping in `package.json` and no `paths` mapping in `tsconfig.json` or root `tsconfig.json`. AGENTS.md §4 is explicit: "Subpath exports need both package exports and TS path mapping." Sister package `packages/test/package.json` declares both a dev `exports` block AND `typesVersions` for its subpaths. Consequence: in-workspace consumers cannot resolve `@morpho-org/morpho-sdk/bundler` or `@morpho-org/morpho-sdk/abis` against source under NodeNext; only the published artifact resolves. The 16 JSDoc `@example` blocks in `src/bundler/actions.ts` showing `import { BundlerAction } from "@morpho-org/morpho-sdk/bundler"` will fail when copy-pasted into a workspace consumer.

_Suggestion: add a top-level `exports` block in `package.json` (mapping `.`, `./bundler`, `./abis` to the `src/` entry points for dev mode) and a `typesVersions` map; mirror the pattern in `packages/test/package.json`._

Rubilmax:
> 📍 *Finding targets `packages/morpho-sdk/BUNDLER3.md:13` — the line is unchanged in this PR's diff so an inline review comment cannot be anchored. Posting as a top-level comment.*

**[MEDIUM]** Stale line reference `src/actions/vaultV1/deposit.ts#L82-L84` for the `getChainAddresses(chainId).bundler3.generalAdapter1` claim. The added `@throws {Permit2ExpirationMissingError}` JSDoc in `vaultV1/deposit.ts` shifts the actual destructuring; in HEAD it lives at L115-L118 (and was never at L82-L84 even before — those lines are inside the JSDoc).

_Suggestion: update to `#L115` or drop the line anchor._

packages/morpho-sdk/src/abis.ts
Rubilmax: Expected during deprecation

packages/morpho-sdk/src/abis.ts
Rubilmax: Keep consumer support

packages/morpho-sdk/src/bundler/actions.ts
Rubilmax: Will be caught by tests

packages/morpho-sdk/src/bundler/actions.ts
Rubilmax: Documented

packages/morpho-sdk/src/bundler/actions.ts
Rubilmax: Lint is enough

packages/morpho-sdk/src/types/error.ts
Rubilmax: Expected

Foulks-Plb (commented):
## Parallel PR Review (Claude) — focus: security / fund loss / tx flow

**Reviewed commit:** `9cb99a1b`  •  Base: `extract-public-reallocation`

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 0 |
| Low | 1 |

### Transaction-flow / `tx.value` audit (no fund-loss bugs found)

The core refactor — moving `tx.value` from explicit callsite overrides (`value: nativeAmount`, `value: reallocationFee`, `value: nativeAmount + reallocationFee`) to a derived value computed by the new `BundlerAction.encodeBundle` — was traced end-to-end:

- `nativeTransfer(bundler3, generalAdapter1, X)` emits one bundler call with `value=X`; `consumeCallValue` deficits the bundler-available balance and rolls the shortfall into `tx.value`. ✓
- `nativeTransfer(user, bundler3, X)` enters the prefund branch (`addBundlerPrefund`), which increments BOTH `state.value` and `availableBundlerValue`, so later value-carrying calls (e.g. `reallocateTo`) consume the prefund before adding to `tx.value`. ✓
- `reallocateTo` action carries `value: fee`; each `reallocateTo` correctly bumps `tx.value` by `fee` minus any remaining prefund. ✓
- Per-builder traces match the legacy explicit values:
  - `supplyCollateral` (native): `tx.value = nativeAmount` ✓
  - `borrow` (with reallocations): `tx.value = Σ fees` ✓
  - `supplyCollateralBorrow` (native + reallocations): `tx.value = nativeAmount + Σ fees` ✓
  - `repay` / `repayWithdrawCollateral`: `tx.value = 0n` ✓
  - `vaultV1` / `vaultV2 deposit` (native): `tx.value = nativeAmount` ✓
- Callback actions inside `morphoSupplyCollateral` / `morphoRepay` are correctly recursed in `encodeBundleAction`, so a `value > 0n` callback (e.g. a transfer in a callback) correctly contributes to `tx.value`. ✓

The parity test in `bundler/actions.test.ts` documents the new `tx.value = 130n` vs `legacyTx.value = 123n` for the same mixed bundle — the new encoder accounts for a `reallocateTo` fee whose value exceeds the remaining bundler prefund (the legacy `encodeBundle` did not, which is why the old callsites had to override `value` explicitly). Removing the explicit `value` override is therefore only safe *because* the new encoder is more accurate — both observations are true together.

Other correctness checks that came up while auditing the diff:

- The `bundler-sdk-viem` `=== → isAddressEqual` fix (separate changeset, included in this PR) is a real bug fix — the old `===` on `bundler3` / `generalAdapter1` recipients would miss non-checksummed inputs and either skip value-counting or skip the `recipient===bundler3` short-circuit in `nativeTransfer`. The new `morpho-sdk` encoder already uses `isAddressEqual` everywhere. `morpho-sdk` action builders source addresses from `getChainAddresses` (always canonical), so the old `===` bug never affected morpho-sdk outputs, but external bundler-sdk-viem callers could have hit it.
- `BundlerErrors.MissingSignature` / `UnexpectedAction` are now distinct classes from the ones in `@morpho-org/bundler-sdk-viem`. Anyone catching `BundlerErrors.X` from the old bundler-sdk-viem path will not match throws from the new morpho-sdk encoder via `instanceof`. The PR description notes `migration-sdk-viem` intentionally keeps the bundler-sdk-viem wiring, so this is the expected break, but worth flagging in release notes / migration guide.
- Note on scope: the `extract-public-reallocation` base branch (separate PR) is the one that removes the legacy 1h delay margin from `ReallocationData.getMarketPublicReallocations`. That is a behaviour change worth reviewing on the base PR, not this one.

### Findings

See inline comment.

_This is an automated parallel review. Findings list is conservative — only issues a reviewer would clearly act on in this PR._

Foulks-Plb:
**One finding (Low): ABI duplication** — `packages/morpho-sdk/src/abis.ts` ships 12 ABIs but the bundler only uses 3 of them (`bundler3Abi`, `coreAdapterAbi`, `generalAdapter1Abi`). The other 9 (URD, paraswap, Aave/Compound migrations, etc.) are dead code in this package and create a drift surface with `bundler-sdk-viem`. Suggestion: narrow to the 3 actually used, or add a comment naming the expected external consumer.

PR #673, opened 2026-05-18
ci(release): sign and harden release automation

Rubilmax:
## Summary

Updates the release workflows so release commits and package tags are created with a valid GitHub Actions identity, and hardens the release helper scripts around signed commits, rerun recovery, and branch edge cases.

## Details

The version PR workflow still lets the GitHub App own release branch updates, but the version commit is now created through GitHub's GraphQL `createCommitOnBranch` mutation instead of local `git commit`/`git push`. The mutation uses GitHub's `CommittableBranch.branchName` input field, preserves the version-file allowlist, and moves `changeset-release/<base>` to the GitHub-signed commit.

The version PR workflow keeps a single job, but now applies the requested minimum mitigations before minting the write-scoped App token: it records the trusted helper checksum and PATH before `pnpm run version`, verifies `scripts/release/create-version-commit.mjs` is unchanged afterward, truncates the current `$GITHUB_ENV` and `$GITHUB_PATH` command files, rejects non-sample files under `.git/hooks`, and forces the trusted PATH plus expected `RELEASE_BRANCH` on the write-token and PR steps.

The publish job sets `github-actions[bot]` as the repo-local git identity before `pnpm release`, so annotated package tags created by Changesets or rerun recovery have a valid tagger identity instead of failing with `Committer identity unknown`.

The push workflow documents why release staging branches are ignored: GitHub's commit mutation needs a short-lived branch ref for the signed commit before the workflow moves the real release branch to that commit. The ignore patterns are limited to `changeset-release/main-api-commit-*` and `changeset-release/next-api-commit-*` so unrelated branches under `changeset-release/` still run CI.

The release helper now rejects unsafe git-reported paths with control characters or lexical traversal, rechecks resolved real paths before reading files, rejects symlinked or non-file entries, constrains `RELEASE_BRANCH` to `changeset-release/main` or `changeset-release/next`, validates signed-commit temp refs against the same `changeset-release/main-api-commit-*` / `changeset-release/next-api-commit-*` staging patterns before any branch ref write, disables git hooks with `core.hooksPath=/dev/null`, uses `git push --no-verify`, and uses `--force-with-lease=<ref>:` when creating a previously missing release branch. The default temp branch name now uses the same deterministic `releaseBranch-api-commit-runId-runAttempt` format as the workflow entrypoint.

Publish rerun recovery now calls `scripts/release/compute-pending-tag.mjs` instead of embedding `node -e` in the workflow. Manifest reads are constrained to relative `packages/<name>/package.json` paths under the real worktree root and now reject absolute or lexical traversal input.

Release script coverage now includes `create-version-commit` main flow coverage, disallowed path errors, missing/poisoned env, GraphQL and REST failure paths, existing/create-only branch refs, temp ref validation, deterministic default temp ref naming, hook-disabled pushes, push cleanup behavior, delete warnings, in-repo symlinks, control-character paths, lexical traversal paths, and companion script happy/sad branches.

## Validation

- `pnpm exec biome check .github/workflows/version-pr.yml scripts/release/create-version-commit.mjs scripts/release/create-version-commit.test.js`
- `node --check scripts/release/create-version-commit.mjs`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/version-pr.yml")'`
- `pnpm exec vitest run --project scripts scripts/release/create-version-commit.test.js`
- `pnpm exec vitest run --project scripts`
- `pnpm exec vitest run --project scripts --coverage.enabled true --coverage.include 'scripts/release/**/*.{js,mjs}' --coverage.exclude 'scripts/**/*.test.{js,mjs}' --coverage.reporter text --coverage.reporter json-summary` (77 tests; 92.14% statements, 87.5% branches, 90.47% functions, 93.87% lines)
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/publish.yml")'`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/push.yml")'`
- Full root test suite not run; changes are scoped to release workflow scripts and YAML.

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1779097957246369)


.github/workflows/push.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - '**'
+      - '!changeset-release/**-api-commit-*'
```
Foulks-Plb: What are you trying to avoid with this negative filter?

Foulks-Plb (dismissed):

tarikbellamine:
@codex revoew

tarikbellamine (commented):
Security-focused review with severity rankings. Inline comments cover findings tied to specific lines. The following are out-of-diff or pre-existing items, called out here instead:

**P1 — extract and test publish-recovery `node -e`.** The inline ~15-line script at `publish.yml:97-117` (existing code from prior PR, not modified by this one) is release-critical and only fires on rerun-recovery. Move to `scripts/release/compute-pending-tag.mjs` with unit tests. Same anti-pattern called out previously on the github-release-body inline script that was extracted.

**P3 — GitHub App token expiry on long version runs.** The changelog token at `version-pr.yml:59-67` (existing, not modified here) is minted before `pnpm run version`. App installation tokens expire after 1 hour. For a large monorepo with many cross-PR references the changelog plugin resolves, this can expire mid-run. Likely operational, but worth flagging.

**Out-of-diff but gating effectiveness of this PR:**
- Branch protection still off on `main` and `next` (`main` → 404, `next` → `protected:true, enabled:false`). The whole point of signing commits is so a 'require signed commits' branch-protection rule can enforce them; without that rule, the upgrade buys verified UI badges and nothing else operationally. Signed-commit infrastructure is the *prerequisite* for actually requiring signed commits; closing the loop requires a repo-settings PR after this lands.
- `default_workflow_permissions: write` at repo level — flip to `read`.
- `can_approve_pull_request_reviews: true` — disable.
- No `environment:` gate on publish.
- Verify npm trusted publishing + provenance per package.

**Net assessment:** the structural moves in this PR are right (extracted helper with tests, split read-only changelog token, GitHub-signed commits, symlink/realpath validation). The biggest remaining risk isn't path traversal — it's that low-priv `pnpm run version` and high-priv `Create signed version commit` share a runner/worktree, which makes a compromised dep in the version-step graph directly weaponizable into repo writes (see inline on `version-pr.yml:93`). The privilege split is the single highest-leverage change.

.github/workflows/version-pr.yml
```diff
@@ -85,97 +85,12 @@ jobs:
...
-          if git ls-remote --exit-code --heads origin "${RELEASE_BRANCH}" >/dev/null 2>&1; then
-            git fetch origin "+${remote_release_ref}:${local_release_ref}"
-            expected_sha="$(git rev-parse "${local_release_ref}")"
-            git push --force-with-lease="${remote_release_ref}:${expected_sha}" origin "HEAD:${remote_release_ref}"
-          else
-            git push origin "HEAD:${remote_release_ref}"
-          fi
-
-          echo "has_version_changes=true" >> "$GITHUB_OUTPUT"
+        run: node scripts/release/create-version-commit.mjs
```
tarikbellamine: **P1 — Privilege boundary between `pnpm run version` and this step is non-existent; compromised dep in the version-step graph escalates directly to write-token capability.**

Same job, same runner, same worktree. Between `Version packages` (line 71) and this step, the following are all writable by any JS code `@changesets/cli` + `@changesets/changelog-github` + transitive (`node-fetch@2.7.0`, `dataloader@1.4.0`, `dotenv@8.6.0`) execute during versioning:

- `scripts/release/create-version-commit.mjs` itself — next step `node`-loads it from disk
- `scripts/release/detect-pending-changesets.mjs` and any other helper
- `.git/hooks/pre-commit`, `pre-push`, `pre-receive` — the helper's `git push` fires `pre-push` (your own test fixture at `create-version-commit.test.js:284-300` explicitly proves this works in production)
- `$GITHUB_ENV` (file-based — `appendFileSync(process.env.GITHUB_ENV, "RELEASE_BRANCH=main\n")` from any node code propagates to every later step, including this one)
- `$GITHUB_PATH` (same mechanism, hijacks PATH for later step `node`/`git` calls)
- `node_modules/**` (next step's `node` may resolve compromised modules)
- Background processes that outlive the step

Read-only changelog token reduces *token* blast radius. It does nothing for this attack class: the dep doesn't need a token, it needs to set up the worktree for this step to do its work with the write-scoped App token.

The path allowlist in `collectVersionChanges` only filters paths going INTO the signed commit. It does not protect any of the surfaces above.

**Recommended fix — structural, not a patch:**

- Split into two jobs.
- Low-priv job: runs `pnpm run version`, uploads `package.json` / `CHANGELOG.md` / `.changeset/*.md` / `.changeset/pre.json` files as an artifact. No App token, no write permissions, no GitHub API access beyond read.
- High-priv job: fresh `actions/checkout` at `github.sha` (gets unmutated helper code), downloads artifact, validates artifact path allowlist AND content (see comment on `scripts/release/create-version-commit.mjs:98`), then mints the write-scoped App token and creates the signed commit.

One structural change closes this finding, the `RELEASE_BRANCH`-from-`$GITHUB_ENV` poisoning path (see `create-version-commit.mjs:288`), and largely M1's compromised-dep-mutates-allowed-file path.

**Minimum mitigations if the structural fix isn't in scope:**
- Compute SHA-256 of `scripts/release/create-version-commit.mjs` before `pnpm run version`, verify unchanged before minting the write token.
- Run all `git` invocations in the helper with `-c core.hooksPath=/dev/null` and `git push --no-verify`.
- Clear `GITHUB_ENV`/`GITHUB_PATH` between low-priv and high-priv steps (write `> $GITHUB_ENV` before the App token mint).
- Inspect `.git/hooks/` for unexpected files before the write-token step.

These are patches, not boundaries — the structural split is the right answer.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+    assertPathInsideBase({
+      absolutePath: realpathSync(absolutePath),
+      basePath,
+      path,
+    });
+
+    additions.push({
+      contents: readFileSync(absolutePath).toString("base64"),
+      path,
+    });
```
tarikbellamine, line 125: **P1 — Path allowlist isn't enough; content of `package.json` should also be validated before sending to the signed commit.**

A compromised changelog plugin could inject `scripts.prepublishOnly` (or `postinstall`, or modify `bin`/`exports`/`files`/`main`) into a `packages/*/package.json` during `pnpm run version`. That file passes the path allowlist at line 22, gets base64-encoded here, ships into the GitHub-signed commit, rides the release PR (which reviewers tend to rubber-stamp since it's auto-generated version bumps), merges (no branch protection on main yet), and the published tarball runs the malicious lifecycle script when consumers install.

`@changesets/cli` shouldn't legitimately need to modify anything in `package.json` except `version` and possibly versioned internal-workspace dep entries (`workspace:*` style). Anything else is a smell.

Recommended validation (in the high-priv job per the privilege-split comment on `version-pr.yml:93`):

```js
function assertSafePackageJsonDiff(beforeContent, afterContent) {
  const before = JSON.parse(beforeContent);
  const after = JSON.parse(afterContent);
  const allowedKeyChanges = new Set(["version"]);
  const allowedDepBlocks = new Set(["dependencies", "devDependencies", "peerDependencies", "optionalDependencies"]);

  const allKeys = new Set([...Object.keys(before), ...Object.keys(after)]);
  for (const key of allKeys) {
    if (before[key] === after[key]) continue;
    if (allowedKeyChanges.has(key)) continue;
    if (allowedDepBlocks.has(key)) {
      // Only dep-version values may change; dep names must not.
      const beforeDeps = before[key] ?? {};
      const afterDeps = after[key] ?? {};
      const beforeNames = Object.keys(beforeDeps).sort().join(",");
      const afterNames = Object.keys(afterDeps).sort().join(",");
      if (beforeNames !== afterNames) throw new Error(`Disallowed dep-name change in ${key}.`);
      continue;
    }
    throw new Error(`Disallowed package.json field change: ${key}.`);
  }
}
```

The `before` content can come from `git show HEAD:<path>` (since the commit base is the workflow's `github.sha`). If `existsSync` is false (new package), accept the addition but flag it for review.

This is the second pillar of the defense — the privilege split on `version-pr.yml:93` reduces the attack surface; this validates the artifact content from a dep that's already past the boundary.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+  const fetchImpl = options.fetchImpl ?? fetch;
+  const pushReleaseBranch =
+    options.pushReleaseBranch ?? pushReleaseBranchWithLease;
+  const tempBranch =
+    options.tempBranch ??
+    `${options.releaseBranch}-api-commit-${randomUUID().slice(0, 8)}`;
+
+  if (tempBranch === options.releaseBranch) {
+    throw new Error("Temporary branch must differ from the release branch.");
+  }
```
tarikbellamine, line 161: **P2 — tempBranch construction has three related concerns:**

**1. Collision-clobber guard is too narrow.** The check at line 126 only rejects `tempBranch === options.releaseBranch`. `createOrUpdateBranchRef` (line 357) PATCH-force-overwrites whatever branch matches the temp name. If a future caller passes `releaseBranch: "main"` (intentionally or by mistake), tempBranch becomes `main-api-commit-<uuid>` — a name that *also* doesn't match `push.yml`'s `!changeset-release/**-api-commit-*` ignore-glob, so triggers a Release run, then gets force-overwritten silently. Defense in depth: assert `tempBranch.startsWith("changeset-release/")` and `tempBranch.includes("-api-commit-")` before issuing the force PATCH.

**2. Two defaults diverge (P3).** This one uses `randomUUID().slice(0, 8)`. `main()` at line 292-297 builds a deterministic name from `GITHUB_RUN_ID` + `GITHUB_RUN_ATTEMPT`. Only the latter runs in CI; this default is reachable only from tests or future callers. Pick one (recommend the deterministic form — easier to clean up stale branches if needed).

**3. 8-hex random suffix is short (P3).** ~65k birthday-collision. CI doesn't hit this because it uses deterministic naming, but tests do. Worth tightening to 16 hex if this default sticks around.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+          },
+          expectedHeadOid: options.baseSha,
+          fileChanges: options.fileChanges,
+          message: {
+            headline: commitMessage,
+          },
+        },
+      },
+    });
+    const commitOid = data.createCommitOnBranch.commit.oid;
```
tarikbellamine: **P2 — `commitOid` is consumed unvalidated.**

It's used as a git source ref at line 241 (`${options.commitOid}:${remoteReleaseRef}`), written to `$GITHUB_OUTPUT` (line 40), and logged to stdout (line 320). The only thing between "what GitHub's GraphQL response body contains" and "what we pass to `git push`" is the GraphQL schema enforced server-side. That's almost certainly fine, but defense in depth:

```js
const commitOid = data?.createCommitOnBranch?.commit?.oid;
if (typeof commitOid !== "string" || !/^[0-9a-f]{40}$/.test(commitOid)) {
  throw new Error(`Invalid commit OID returned by GitHub: ${JSON.stringify(commitOid)}`);
}
```

(Same applies to `existingRef.object.sha` in `createOrUpdateBranchRef` if that path were ever to use it — currently it isn't, but if a future change does, the validation belongs there too.)

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+          `${options.commitOid}:${remoteReleaseRef}`,
+        ],
+        { cwd: options.cwd },
+      );
+      return;
+    }
+
+    runGit(["push", "origin", `${options.commitOid}:${remoteReleaseRef}`], {
+      cwd: options.cwd,
+    });
```
tarikbellamine: **P1 — First-time release-branch push lacks lease protection.**

When `hasRemoteBranch` at line 226 returns false, we fall through to this non-force, non-lease push. TOCTOU window: between `hasRemoteBranch` and this push, a concurrent creator (workflow, bot, human with write) could create the release branch at an unrelated SHA. The non-force push then either:

- Fails (good — fail closed), OR
- Succeeds via fast-forward if the new remote SHA happens to be an ancestor of `commitOid` (e.g., it's the base SHA itself), silently extending someone else's branch.

The second case is rare but achievable when both runs derived from the same `github.sha`.

The test fixture at `create-version-commit.test.js:340-369` pre-creates the release branch before exercising `pushReleaseBranchWithLease`, so this fallthrough is also **not covered by any test**. Two issues, one root cause.

Fix: use `--force-with-lease=<ref>:` with empty expected SHA (modern git accepts this as 'create only'):

```js
runGit([
  "push",
  `--force-with-lease=${remoteReleaseRef}:`,
  "origin",
  `${options.commitOid}:${remoteReleaseRef}`,
], { cwd: options.cwd });
```

Then add a test covering this branch (release branch not present on remote).

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+}
+
+/**
+ * Runs the release commit workflow step.
+ *
+ * @param {{ apiBaseUrl?: string, cwd?: string, env?: NodeJS.ProcessEnv, fetchImpl?: typeof fetch, outputFile?: string }} options Runtime options.
+ * @returns {Promise<null | { commitOid: string, tempBranch: string }>} The commit result when changes exist.
+ */
+export async function main(options = {}) {
+  const cwd = options.cwd ?? process.cwd();
```
tarikbellamine, line 305: **P3 — `process.cwd()` default silently assumes invocation from repo root.**

If a future maintainer runs this script standalone for debugging from a subdirectory (`packages/foo/`), `collectVersionChanges` would call `git diff` from there and either error or return wrong paths. Not exploitable, but the failure mode is confusing.

Minor: validate that `cwd` contains a `.git/` directory before proceeding, or document the requirement in the JSDoc.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+
+  if (versionChanges.paths.length === 0) {
+    process.stdout.write("No version changes to commit.\n");
+    appendOutput(outputFile, getGitHubOutput({ hasVersionChanges: false }));
+    return null;
+  }
+
+  const token = readRequiredEnv(env, "GH_TOKEN");
+  const repository = readRequiredEnv(env, "GITHUB_REPOSITORY");
+  const releaseBranch = readRequiredEnv(env, "RELEASE_BRANCH");
```
tarikbellamine: **P1 — `RELEASE_BRANCH` is read from mutable job env; compromised `pnpm run version` can poison it via `$GITHUB_ENV`.**

`version-pr.yml:23` sets `RELEASE_BRANCH: changeset-release/${{ github.ref_name }}` at job level. But `$GITHUB_ENV` is a *file* that any step can append to, and appends propagate to every later step's `process.env`. A compromised changelog dep could do:

```js
fs.appendFileSync(process.env.GITHUB_ENV, "RELEASE_BRANCH=main\n");
```

Next time this script runs, `releaseBranch === "main"`. It then:
1. Creates temp branch `main-api-commit-<run>-<attempt>` — which does NOT match `push.yml`'s `!changeset-release/**-api-commit-*` ignore-glob, so the API branch creation fires a real Release workflow run as a secondary effect.
2. Force-with-lease pushes the signed commit to `refs/heads/main`.

With no branch protection on `main` today, the push succeeds. Next CI run sees a new commit on main → publish path runs → npm gets whatever the dep produced.

**Fix — three layers, apply all:**

```js
const ALLOWED_RELEASE_BRANCHES = new Set([
  "changeset-release/main",
  "changeset-release/next",
]);

const releaseBranch = readRequiredEnv(env, "RELEASE_BRANCH");
if (!ALLOWED_RELEASE_BRANCHES.has(releaseBranch)) {
  throw new Error(`Invalid RELEASE_BRANCH "${releaseBranch}".`);
}

// Also assert it matches the immutable GHA-provided ref:
const refName = readRequiredEnv(env, "GITHUB_REF_NAME");
if (releaseBranch !== `changeset-release/${refName}`) {
  throw new Error(`RELEASE_BRANCH "${releaseBranch}" does not match GITHUB_REF_NAME "${refName}".`);
}
```

Additionally: set `RELEASE_BRANCH` in the **step** env block (not the job env block), so it's resolved at step-start from the workflow-trigger value and doesn't read from a `$GITHUB_ENV` overwrite.

The privilege split on `version-pr.yml:93` mostly closes this too (fresh job env), but defense in depth — apply both.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+ * @returns {string} GitHub Actions output lines.
+ */
+export function getGitHubOutput(result) {
+  const output = [
+    `has_version_changes=${result.hasVersionChanges ? "true" : "false"}`,
+  ];
+
+  if (result.commitOid != null) {
+    output.push(`commit_sha=${result.commitOid}`);
+  }
```
tarikbellamine, line 54: **P3 — `commit_sha` step output is set but never consumed.**

`version-pr.yml:96` references `steps.version-commit.outputs.has_version_changes` but never `commit_sha`. Dead output. Harmless, but if the intent was to surface the signed commit OID into the release-PR body (a reasonable thing — surfaces the GitHub-verified attestation), the wire isn't connected.

Either drop the output here, or thread it into the `Open or update release PR` step's body.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+  const releaseBranch = readRequiredEnv(env, "RELEASE_BRANCH");
+  const baseSha = runGit(["rev-parse", "HEAD"], { cwd })
+    .toString("utf8")
+    .trim();
+  const tempBranch = [
+    releaseBranch,
+    "api-commit",
+    env.GITHUB_RUN_ID ?? "local",
+    env.GITHUB_RUN_ATTEMPT ?? "0",
+  ].join("-");
```
tarikbellamine: **P2 — Naming convention here is implicitly coupled to `push.yml`'s ignore-glob.**

This builds `${releaseBranch}-api-commit-${runId}-${attempt}` (e.g., `changeset-release/main-api-commit-12345-1`). `push.yml:11` filters out branches matching `!changeset-release/**-api-commit-*` to break the recursion when the API creates the staging branch. If either side drifts (script switches separator to `_staging_`, or someone narrows the glob — see comment on `push.yml:11`), the API branch creation triggers a real CI run mid-release.

Recommendation: export a constant from this file referencing the naming pattern, and import/reference it both from the workflow (via a comment) and a test that asserts the generated names match the pattern.

```js
export const TEMP_BRANCH_PATTERN = /^changeset-release\/(main|next)-api-commit-/;
```

Pair with a test that constructs a tempBranch from this module's logic and asserts `TEMP_BRANCH_PATTERN.test(tempBranch)`. That way a future change on either side fails CI loudly.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,574 @@
...
+  }
+
+  return value;
+}
+
+function readNullSeparatedGitOutput(output) {
+  return output
+    .toString("utf8")
+    .split("\0")
+    .filter((path) => path !== "");
```
tarikbellamine: **P3 — `readNullSeparatedGitOutput` doesn't sanitize control chars in returned paths.**

Only `sanitizeLogLine` filters them, and that's only at log time (line 540). Paths from git pass through to the GraphQL `path` field and `additions[].path` of the commit unchanged. GraphQL will reject most pathological cases server-side, but defense in depth: filter early.

```js
function readNullSeparatedGitOutput(output) {
  return output
    .toString("utf8")
    .split("\0")
    .filter((path) => path !== "")
    .map((path) => {
      if (/[\x00-\x1f\x7f]/.test(path)) {
        throw new Error(`Git returned path with control characters: ${JSON.stringify(path)}`);
      }
      return path;
    });
}
```

Minor; current attack surface is essentially nil because git itself rejects most weird paths.

scripts/release/create-version-commit.test.js
```diff
@@ -0,0 +1,420 @@
...
+
+import {
+  collectVersionChanges,
+  createSignedVersionCommit,
+  getGitHubOutput,
+  isAllowedVersionPath,
+  pushReleaseBranchWithLease,
+} from "./create-version-commit.mjs";
+
+const tempDirs = [];
```
tarikbellamine, line 23: **P2 — Test coverage gaps.**

The existing tests cover the happy paths well. Failure paths and the create-only branch path aren't covered, which matters here because the recovery/edge-case behavior is exactly what the security model relies on. Suggested additions:

1. **`main()` end-to-end** — currently only individual exports are tested. Wire it up with mocked `fetchImpl`, fake `GITHUB_OUTPUT`, env vars, and a real temp git repo. Verify the full flow: env reading, version detection, signed commit, push, output file written.

2. **`disallowedPaths.length > 0` throw in `main()`** — currently `collectVersionChanges` is tested for it but `main()` isn't. Verify the script exits non-zero and the disallowed paths appear in stderr (with control-char sanitization applied).

3. **`pushReleaseBranchWithLease` create-only path** (line 248-250 of the script). The current fixture `createReleaseRemoteFixture` pre-creates the release branch. Add a fixture variant where the release branch does NOT exist remotely; assert the script handles it correctly (and after the lease-fix in the inline comment on line 248, assert it uses `--force-with-lease=<ref>:`).

4. **`createSignedVersionCommit` failure paths**:
   - GraphQL response containing `errors: [...]`
   - REST `createOrUpdateBranchRef` returning 5xx
   - REST PATCH (update existing ref) failing
   - Mutation success but push fails — verify `finally` deletes temp branch
   - Delete failure — verify warning is written, doesn't fail the run

5. **In-repo symlinks** — the existing test covers symlinks pointing outside the repo. Add one for symlinks pointing within the repo (e.g., `packages/foo/package.json` → `packages/bar/package.json`). Should also throw (`isFile()` returns false on a symlink regardless of target).

6. **`$GITHUB_ENV` poisoning attempt** — set `process.env.RELEASE_BRANCH` to something that doesn't match the expected pattern, assert the script rejects it (after the validation fix in the inline comment on line 288).

7. **Control-character paths** — feed `collectVersionChanges` a git output containing paths with embedded `\x01` etc., assert it throws (after the filtering fix on line 475).

8. **Lexical traversal** — feed `collectVersionChanges` an output containing a path like `packages/../etc/passwd`, assert it throws.

.github/workflows/push.yml
```diff
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - '**'
+      # create-version-commit.mjs uses a short-lived API staging branch because
+      # GitHub's createCommitOnBranch mutation needs a branch ref to create the
+      # GitHub-signed version commit. The workflow moves the real release branch
+      # to that commit afterward, so skip CI/release runs for the staging branch.
+      - '!changeset-release/**-api-commit-*'
```
tarikbellamine: **Three findings here, P2 + P3:**

**P2 — Stale staging branches accumulate when the API delete in `finally` fails.** `create-version-commit.mjs:187-191` swallows delete errors with a stderr warning. Repeated failures (rate-limit window, transient 5xx) produce a slow-growing population of `changeset-release/**-api-commit-*` refs. Cosmetic until the next person needs to audit branches under the namespace. Worth a periodic cleanup workflow that purges `changeset-release/**-api-commit-*` older than 24h.

**P2 — Naming-pattern coupling is implicit.** This filter has to stay in sync with the temp-branch construction in `create-version-commit.mjs:122-124` and `:292-297`. Drift either way means either: (a) staging branch creation triggers a real CI run mid-release, or (b) some legitimate branch is silently skipped. See the comment on `create-version-commit.mjs:292-297` for a proposed shared-constant approach.

**P3 — Glob is broader than the script needs.** `!changeset-release/**-api-commit-*` matches *any* branch under `changeset-release/` ending with `-api-commit-*`, not just the two intended namespaces (main/next). Practical exposure today is small (no branch protection on main means anyone with write can already bypass CI by pushing to main), but it's a real CI blind spot — `changeset-release/anything-api-commit-foo` skips lint/build/test. Narrow:

```yaml
branches:
  - '**'
  - '!changeset-release/main-api-commit-*'
  - '!changeset-release/next-api-commit-*'
tags-ignore:
  - '**'
```

Pairs with the validation in `create-version-commit.mjs:288` so the script also constrains the allowed `RELEASE_BRANCH` values.

.github/workflows/publish.yml
```diff
@@ -42,6 +42,11 @@ jobs:
 
       - run: pnpm install --frozen-lockfile
 
+      - name: Configure git identity
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
```
tarikbellamine, line 78: **P1 — flagging an out-of-diff finding here for proximity:** the inline `node -e` recovery code at `publish.yml:97-117` (not modified by this PR) is the same anti-pattern that motivated the extraction of `create-version-commit.mjs` and `github-release-body.mjs`. It's ~15 lines of release-critical logic embedded in YAML, in the rerun-recovery branch — exactly the path you want tested. Extract to `scripts/release/compute-pending-tag.mjs` with unit tests.

Also related to the new `Configure git identity` step here: the recovery path runs `git tag "$tag" -m "$tag"` (annotated tag, needs identity). The fact that this PR adds the identity config means somebody noticed the recovery path needs it, which is great — but the recovery logic itself is still inline and untested. Land the extraction in a follow-up before relying on the recovery path in anger.

Rubilmax:
> **P1 — extract and test publish-recovery `node -e`.** The inline ~15-line script at `publish.yml:97-117` (existing code from prior PR, not modified by this one) is release-critical and only fires on rerun-recovery. Move to `scripts/release/compute-pending-tag.mjs` with unit tests. Same anti-pattern called out previously on the github-release-body inline script that was extracted.

Applying

> **P3 — GitHub App token expiry on long version runs.** The changelog token at `version-pr.yml:59-67` (existing, not modified here) is minted before `pnpm run version`. App installation tokens expire after 1 hour. For a large monorepo with many cross-PR references the changelog plugin resolves, this can expire mid-run. Likely operational, but worth flagging.

Let's see when it happens, if it happens

> **Out-of-diff but gating effectiveness of this PR:**
> 
> * Branch protection still off on `main` and `next` (`main` → 404, `next` → `protected:true, enabled:false`). The whole point of signing commits is so a 'require signed commits' branch-protection rule can enforce them; without that rule, the upgrade buys verified UI badges and nothing else operationally. Signed-commit infrastructure is the _prerequisite_ for actually requiring signed commits; closing the loop requires a repo-settings PR after this lands.
> * `default_workflow_permissions: write` at repo level — flip to `read`.
> * `can_approve_pull_request_reviews: true` — disable.
> * No `environment:` gate on publish.
> * Verify npm trusted publishing + provenance per package.

Will do after merge

scripts/release/create-version-commit.mjs
Rubilmax: I dont see the point

.github/workflows/push.yml
Rubilmax: Fixed P3. The others are overkill

scripts/release/create-version-commit.mjs
Rubilmax: Harmless, keeping it

scripts/release/create-version-commit.mjs
Rubilmax: Harmless, ignored

scripts/release/create-version-commit.mjs
Rubilmax: Harmless, ignored

scripts/release/create-version-commit.mjs
Rubilmax: Feels way overkill and not future-proof enough... I dont see much harm in modifying the package.json right before release

.github/workflows/version-pr.yml
Rubilmax: Separating jobs is too complex we need to share an artifact in-between. Instead i apply the "minimum mitigations"

scripts/release/compute-pending-tag.mjs
```diff
@@ -0,0 +1,159 @@
...
+    baseRef: options.baseRef,
+    cwd: options.cwd,
+    manifestPath: options.manifestPath,
+  });
+
+  if (previousManifest?.version === manifest.version) {
+    return undefined;
+  }
+
+  return `${manifest.name}@${manifest.version}`;
```
Rubilmax: `computePendingTag` returns `${manifest.name}@${manifest.version}` without checking that both fields are non-empty strings. A manifest missing `name` or `version` would silently yield a tag like `undefined@1.2.3`, which `publish.yml` then feeds into `git tag`. Validate both fields and throw an actionable error otherwise — mirroring the version checks already done in `github-release-body.mjs`'s `readReleasePackages`.

scripts/release/compute-pending-tag.mjs
```diff
@@ -0,0 +1,159 @@
...
+    throw new Error(`Invalid manifest path "${options.manifestPath}".`);
+  }
+
+  assertPathInsideBase({
+    absolutePath: realpathSync(manifestPath.absolutePath),
+    basePath: manifestPath.basePath,
+    manifestPath: options.manifestPath,
+  });
+
+  return JSON.parse(readFileSync(manifestPath.absolutePath, "utf8"));
```
Rubilmax, line 39: `readPackageManifest` does `JSON.parse(readFileSync(...))` with no try/catch. A present-but-malformed `package.json` throws a raw `SyntaxError` with no manifest path. Wrap the parse and rethrow including `options.manifestPath`, consistent with this file's other `Invalid manifest path "..."` errors.

scripts/release/compute-pending-tag.mjs
```diff
@@ -0,0 +1,159 @@
...
+
+  try {
+    return JSON.parse(
+      execFileSync("git", ["show", `${baseRef}:${manifestPath.relativePath}`], {
+        cwd: options.cwd,
+        encoding: "utf8",
+        stdio: ["ignore", "pipe", "ignore"],
+      }),
+    );
+  } catch {
```
Rubilmax: The bare `catch {}` in `readPreviousPackageManifest` swallows every failure of `git show ${baseRef}:...` — not only the intended "path did not exist at the base ref" case, but also malformed-JSON and unexpected git failures. Any of those is then treated as "package is new" and a tag is emitted. Narrow the catch to the expected unknown-revision/path git error and let parse / unexpected errors propagate.

Rubilmax (commented):

**Reviewed commit:** `13f1e938`

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 0 |
| Low | 12 |

Solid PR. `npx biome check scripts/release/` is clean and the colocated Vitest suite passes (77 tests). The signed-commit flow, allowlist enforcement, path-traversal/symlink/control-char validation, and token-scope ordering (read-only token for `pnpm run version`, write token only after the hardening step) are all sound and well tested. All findings below are low-severity polish: defensive error handling, helper deduplication, and consistency with the sibling release scripts.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,636 @@
...
+          },
+          expectedHeadOid: options.baseSha,
+          fileChanges: options.fileChanges,
+          message: {
+            headline: commitMessage,
+          },
+        },
+      },
+    });
+    const commitOid = data.createCommitOnBranch.commit.oid;
```
Rubilmax: `data.createCommitOnBranch.commit.oid` is dereferenced with no guard. `graphqlRequest` only throws on a populated `errors` array; a 200 response with an unexpected/`null` `data.createCommitOnBranch` shape would throw an opaque `TypeError`. Guard the path and throw an explanatory error that includes a summary of the response.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,636 @@
...
+      runGit(["fetch", "origin", `+${remoteReleaseRef}:${localReleaseRef}`], {
+        cwd: options.cwd,
+      });
+      const expectedSha = runGit(["rev-parse", localReleaseRef], {
+        cwd: options.cwd,
+      })
+        .toString("utf8")
+        .trim();
+
+      runGit(
```
Rubilmax: `pushReleaseBranchWithLease` has two near-identical `runGit(["push", "--no-verify", "--force-with-lease=...", "origin", ...])` invocations that differ only in the lease expected-sha suffix (`:${expectedSha}` vs empty). Compute the lease value once (existing SHA or empty string) and call `push` a single time to remove the duplication and the mid-function early `return`.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,636 @@
...
+}
+
+/**
+ * Pushes the signed temporary-branch commit to the release branch with lease protection.
+ *
+ * @param {{ commitOid: string, cwd: string, releaseBranch: string, remoteUrl?: string, repository: string, tempBranch: string, token: string }} options Push options.
+ * @returns {void}
+ */
+export function pushReleaseBranchWithLease(options) {
+  const originalOriginUrl = runGit(["remote", "get-url", "origin"], {
```
Rubilmax, line 243: `pushReleaseBranchWithLease` captures `git remote get-url origin` and restores it in `finally`. With `persist-credentials: false` the checkout URL is clean today, so this is safe — but the function is exported/reusable and round-tripping whatever `get-url` returns would re-emit a credential if origin ever carried an `x-access-token:` URL. Prefer reconstructing the clean URL from `options.repository` rather than capturing and restoring.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,636 @@
...
+    baseSha,
+    cwd,
+    fetchImpl: options.fetchImpl,
+    fileChanges: {
+      additions: versionChanges.additions,
+      deletions: versionChanges.deletions,
+    },
+    releaseBranch,
+    repository,
+    pushReleaseBranch: options.pushReleaseBranch,
```
Rubilmax, line 341: Object-key ordering nit: in this `createSignedVersionCommit({ ... })` call, `pushReleaseBranch` is placed after `releaseBranch`/`repository`, breaking the alphabetical key ordering used consistently throughout these scripts and their JSDoc option shapes. Move it between `fileChanges` and `releaseBranch`.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,636 @@
...
+    cwd: options.cwd,
+    stdio: options.stdio,
+  });
+}
+
+function formatIndentedList(paths) {
+  return paths.map((path) => `  ${sanitizeLogLine(path)}`).join("\n");
+}
+
+function sanitizeLogLine(value) {
```
Rubilmax: `sanitizeLogLine` here is a verbatim copy of the one in `detect-pending-changesets.mjs`. `getErrorMessage` and the path-containment guard (`assertPathInsideBase` / `isPathInside`) are likewise reimplemented across `create-version-commit.mjs`, `compute-pending-tag.mjs`, and `github-release-body.mjs`. Per AGENTS.md §1 (single source of truth), extract a shared `scripts/release/` helper module and import these from one place.

.github/workflows/version-pr.yml
```diff
@@ -75,114 +91,70 @@ jobs:
...
-          done < <(git ls-files --others --exclude-standard -z)
-
-          if [ "${#unstaged[@]}" -gt 0 ] || [ "${#untracked[@]}" -gt 0 ]; then
-            echo "::error::Versioning produced files outside the release allowlist:"
-            printf '  %s\n' "${unstaged[@]}" "${untracked[@]}"
+          actual_helper_sha="$(sha256sum scripts/release/create-version-commit.mjs | cut -d ' ' -f 1)"
+          if [ "${actual_helper_sha}" != "${EXPECTED_HELPER_SHA}" ]; then
+            echo "::error::scripts/release/create-version-commit.mjs changed during versioning."
             exit 1
           fi
```
Rubilmax: The hardening step scans `.git/hooks/` for unexpected files but does not detect a redirected `core.hooksPath` planted in `.git/config` during `pnpm run version`. `create-version-commit.mjs`'s own git calls are protected by `-c core.hooksPath=/dev/null` + `--no-verify`, so this isn't currently exploitable — but for completeness of the "detect hook tampering" intent, also assert `git config --get core.hooksPath` is empty/unset here.

.github/workflows/version-pr.yml
```diff
@@ -75,114 +91,70 @@ jobs:
...
+          app-id: ${{ secrets.VERSION_APP_ID }}
+          private-key: ${{ secrets.VERSION_APP_PRIVATE_KEY }}
+          permission-contents: write
+          permission-pull-requests: write
 
-          echo "has_version_changes=true" >> "$GITHUB_OUTPUT"
+      - name: Create signed version commit
+        id: version-commit
+        if: steps.pending-changesets.outputs.has_changesets == 'true'
+        env:
```
Rubilmax, line 153: `changeset-release/${{ github.ref_name }}` is re-interpolated inline in several steps (here, line 99, line 154) even though the job-level `env:` already defines `RELEASE_BRANCH` with the same value, and the hardening step additionally re-writes `RELEASE_BRANCH`/`TARGET_BRANCH` into `GITHUB_ENV` with identical values. Reuse the job-level env (`${{ env.RELEASE_BRANCH }}` / `$RELEASE_BRANCH`) and drop the redundant `GITHUB_ENV` rewrite to keep one derivation.

.github/workflows/push.yml
```diff
@@ -4,6 +4,12 @@ on:
   push:
     branches:
       - '**'
+      # create-version-commit.mjs uses a short-lived API staging branch because
+      # GitHub's createCommitOnBranch mutation needs a branch ref to create the
+      # GitHub-signed version commit. The workflow moves the real release branch
+      # to that commit afterward, so skip CI/release runs for main/next staging.
+      - '!changeset-release/main-api-commit-*'
```
Rubilmax, line 11: These branch-ignore globs encode the temp staging-branch naming scheme, which is independently encoded in `create-version-commit.mjs` (`TEMP_BRANCH_RE` and `buildTempBranchName`). A change to the script's naming would silently start triggering CI on staging branches unless this filter is updated in lockstep. The added comment helps; consider explicitly cross-referencing `create-version-commit.mjs` so the coupling is unmissable.

scripts/release/detect-pending-changesets.mjs
```diff
@@ -43,23 +43,23 @@ function sanitizeLogLine(value) {
...
+    writeOutput("Pending changesets:\n");
+    writeOutput(`${pendingChangesets.map(sanitizeLogLine).join("\n")}\n`);
   }
 
   return pendingChangesets;
 }
 
-export function main() {
-  reportPendingChangesets();
+export function main(options = {}) {
```
Rubilmax, line 50: Two consistency gaps with the new sibling scripts in this PR, both touching code modified here: (1) the exported functions in this file (`listPendingChangesets`, `getGitHubOutput`, `reportPendingChangesets`, `main`) still lack the JSDoc that `compute-pending-tag.mjs` / `create-version-commit.mjs` carry on every export; (2) the CLI entrypoint calls `main()` without the `try/catch` + `process.exitCode = 1` wrapper the other release scripts use. Per AGENTS.md §9, align touched code with the convention adopted by the new files.

Foulks-Plb:
## CRITICAL

**1. `$GITHUB_ENV` / `$GITHUB_PATH` not truncated before minting the write-scoped App token (`version-pr.yml`).** Violates `CLAUDE.md §10(b)`. Between `pnpm install --frozen-lockfile` (runs dep lifecycle scripts) and `Generate Release GitHub App Token`, anything can append `NODE_OPTIONS=--require /tmp/evil.js` to `$GITHUB_ENV` or shim PATH — payload then runs inside `Create signed version commit` with `GH_TOKEN` in env, exfiltrating the App token or producing arbitrary signed commits. Same gap in `publish.yml` between `pnpm install` and `pnpm release` (which holds the OIDC `id-token: write` for npm trusted publishing — worst case is publishing a malicious `@morpho-org/*` package with valid provenance).
**Fix:** add a step right before each write-token / privileged step that does `: > "$GITHUB_ENV"` and `: > "$GITHUB_PATH"` (and exits non-zero if either was non-empty, so tampering is visible rather than silently dropped).

**2. Trusted `PATH:` not forced on the privileged step.** Violates `CLAUDE.md §10`. `Create signed version commit` only sets `GH_TOKEN`; `create-version-commit.mjs` calls `execFileSync("git", …)` and resolves via inherited `PATH`. Same for `Open or update release PR` (`gh`, `git`).
**Fix:** on both privileged steps, set `env: PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` (and keep `RELEASE_BRANCH` explicit at step level, not only inherited).

## HIGH

**3. No `.git/hooks/` non-`.sample` rejection.** Violates `CLAUDE.md §10(c)`. Described in the PR body but not in the YAML. `core.hooksPath=/dev/null` inside `GIT_SAFE_CONFIG` only covers git invocations from the Node script — `Configure git identity` (`publish.yml`), `Open or update release PR` (`gh` reads git config), and any future out-of-script command still use default hooks.
**Fix:** before each write-token step, fail if `find .git/hooks -type f ! -name '*.sample' -print -quit` is non-empty.

**4. Integrity baseline too narrow (`maxdepth 1`, `*.mjs` only).** A future `scripts/release/internal/foo.mjs`, `.js`, `.cjs`, or imported `.json` is silently unprotected. `node_modules` integrity also relies entirely on the lockfile — no check that `pnpm install` didn't write outside `node_modules/`.
**Fix:** hash `find scripts/release -type f` (drop `maxdepth` and `-name '*.mjs'`), or walk the import graph from `create-version-commit.mjs` / `compute-pending-tag.mjs` and pin that explicitly.

## MEDIUM

**5. `pushReleaseBranchWithLease` writes the token into `.git/config` via `remote set-url`** for the duration of fetch+push (restored in `finally`). Low risk on a single-step runner but avoidable.
**Fix:** pass the credential per-invocation with `git -c http.https://github.com/.extraheader="AUTHORIZATION: bearer $GH_TOKEN" …` instead of `set-url`.

**6. `git tag \"$tag\" -m \"$tag\"` in `publish.yml`** is safe today only because `compute-pending-tag.mjs` constrains the value; brittle if the regex widens.
**Fix:** validate `\$tag` against `^@morpho-org/[a-z0-9-]+-v?[0-9].*$` in the workflow before passing to `git tag`, so the contract is enforced at the shell boundary too.

scripts/release/create-version-commit.mjs
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    assertPathInsideBase({
+      absolutePath: realpathSync(absolutePath),
+      basePath,
+      path,
+    });
+
+    additions.push({
+      contents: readFileSync(absolutePath).toString("base64"),
```
tarikbellamine: **P1 — Content-level `package.json` validation is missing.**

The path allowlist at `isAllowedVersionPath` (line 23) gates *which* files can change. The content here is base64-encoded as-is, with no inspection. A compromised changelog/version dep during `pnpm run version` can mutate `packages/foo/package.json` to add any of:

- `scripts.prepublishOnly` / `postinstall` / `preinstall` / `install`
- `publishConfig` (e.g., redirected `registry`)
- `bin` (puts attacker-controlled script on consumer `node_modules/.bin/`)
- `files`, `main`, `exports`, `module`, `types` (ships attacker-controlled files in the tarball)
- `packageManager` (later install-time hijack)
- Substituted dep names in `dependencies`/`peerDependencies` (typosquats)

All pass the path allowlist and ship into the signed commit.

**This is the shortest practical exploit chain in the threat model.** No privilege escalation, no PATH poisoning, no hook planting — just one compromised plugin and one PR reviewer who skims a routine version-bump diff. Hits npm on the next release, hits consumers on `npm install`.

**Recommended fix:** parse the before-image via `git show HEAD:<path>` (the pre-image SHA is already `baseSha` at line 306) and assert only allowed fields changed:

```js
const ALLOWED_TOP_LEVEL = new Set(["version"]);
const ALLOWED_DEP_BLOCKS = new Set([
  "dependencies",
  "devDependencies",
  "peerDependencies",
  "optionalDependencies",
]);

function assertSafePackageJsonChange(before, after, path) {
  const all = new Set([...Object.keys(before), ...Object.keys(after)]);
  for (const key of all) {
    if (JSON.stringify(before[key]) === JSON.stringify(after[key])) continue;
    if (ALLOWED_TOP_LEVEL.has(key)) continue;
    if (ALLOWED_DEP_BLOCKS.has(key)) {
      const beforeNames = Object.keys(before[key] ?? {}).sort().join(",");
      const afterNames = Object.keys(after[key] ?? {}).sort().join(",");
      if (beforeNames !== afterNames) {
        throw new Error(`Disallowed dep-name change in ${key} of ${path}.`);
      }
      continue;
    }
    throw new Error(`Disallowed package.json field change "${key}" in ${path}.`);
  }
}
```

Independent of any fix to the helper-script hash check (next comment) — that check doesn't protect file *contents*, only helper code.

.github/workflows/version-pr.yml
```diff
@@ -75,6 +100,26 @@ jobs:
...
+        run: |
+          set -euo pipefail
+
+          actual_helper_sha="$(
+            find scripts/release -maxdepth 1 -type f -name '*.mjs' -print0 |
+              sort -z |
+              xargs -0 sha256sum |
+              sha256sum |
+              cut -d ' ' -f 1
+          )"
```
tarikbellamine: **P1 — Hash verification is defeatable via PATH / `$GITHUB_PATH` shadowing.**

These five commands — `find`, `sort`, `xargs`, `sha256sum`, `cut` — are all PATH-resolved. A compromised dep running during `pnpm run version` can:

1. Write an evil binary to a tmp dir:
   ```js
   fs.mkdirSync("/tmp/x", { recursive: true });
   fs.writeFileSync(
     "/tmp/x/sha256sum",
     "#!/bin/sh\necho '<original-hash>  -'\n",
     { mode: 0o755 },
   );
   ```
   The original hash is computable by the dep itself (it has the unmutated files before tampering them), or readable from the prior privileged-baseline step's output file in `$RUNNER_TEMP/_runner_file_commands/`.

2. Prepend that dir to `$GITHUB_PATH` (which is a file on disk):
   ```js
   fs.appendFileSync(process.env.GITHUB_PATH, "/tmp/x\n");
   ```

3. This step starts with the poisoned `PATH`. Shadowed `sha256sum` returns the expected hash. Verify passes. Write token is minted at line 123. Tampered `create-version-commit.mjs` runs at line 138 with `contents: write` + `pull-requests: write` on the App.

`EXPECTED_HELPER_SHA` itself is protected (step-level `env:` overrides `$GITHUB_ENV` writes), but the *computation* of `actual_helper_sha` isn't. Same attack works on any of the five commands. Compute-in-Node doesn't fully fix this either — `node` is PATH-resolvable and gets shadowed the same way.

**Recommended fix — structural (best):** split into two jobs. Low-priv job runs `pnpm run version`, uploads only `packages/*/package.json`, `packages/*/CHANGELOG.md`, `.changeset/*.md`, `.changeset/pre.json` as an artifact. High-priv job fresh-checkouts at `github.sha` (gets unmutated helper), downloads + validates artifact, mints the write token, creates signed commit. Closes this entire class of attack for free, no PATH gymnastics needed.

**Recommended fix — patch (weaker, brittle):** absolute paths + reset PATH:
```bash
actual_helper_sha="$(
  env -i PATH=/usr/bin:/bin /usr/bin/find scripts/release -maxdepth 1 -type f -name '*.mjs' -print0 |
    /usr/bin/sort -z |
    /usr/bin/xargs -0 /usr/bin/sha256sum |
    /usr/bin/sha256sum |
    /usr/bin/cut -d ' ' -f 1
)"
```
Plus explicit `$GITHUB_ENV`/`$GITHUB_PATH` reset between the version step and this verify step. Brittle: adding any PATH-resolved tool to this step in the future re-opens the hole.

Until this is closed, the hash-check machinery is a speedbump, not a boundary. The whole U1 privilege-separation story rests on it.

.github/workflows/publish.yml
```diff
@@ -40,8 +40,36 @@ jobs:
...
+          )"
+
+          echo "helper_sha=${helper_sha}" >> "${GITHUB_OUTPUT}"
+
       - run: pnpm install --frozen-lockfile
 
+      - name: Configure git identity
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
tarikbellamine, line 76: **P1 — `Push package tags` step's `git push --atomic` (line 192) lets pre-push hooks fire with `GH_TOKEN` in env.**

Noting this on the `Configure git identity` step because that's the natural place to fix it (and what's in the diff). The actual vulnerable line is `publish.yml:192`.

The `create-version-commit.mjs` side already disables hooks (`GIT_SAFE_CONFIG = ["-c", "core.hooksPath=/dev/null"]` + `--no-verify` on the lease push). The publish-side bash git invocations don't. A compromised dep loaded by `pnpm release` at line 127 — `@changesets/cli publish` runs arbitrary plugin code with FS access — can plant `.git/hooks/pre-push` between `pnpm install` and the `Push package tags` step.

When `git push --atomic origin "${refs[@]}"` fires the hook:
- `GH_TOKEN=${{ github.token }}` is in env (workflow-level `contents: write` + `id-token: write`)
- Worktree is cwd
- Stdin is the list of refs being pushed

The hook script can exfiltrate `GH_TOKEN`, push additional commits/tags/branches under the App's identity, or use `id-token: write` to mint its own OIDC token for npm publishing.

**Cheapest fix — repo-level git config in this step (one new line):**
```yaml
- name: Configure git identity
  run: |
    git config user.name "github-actions[bot]"
    git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
    git config core.hooksPath /dev/null
```
This propagates to every git command in the job — no per-invocation flags, and it also covers the recovery-path `git tag` / `git rev-parse` calls in case future git versions or extensions add hook surfaces to them.

**Alternative — per-command at line 192:**
```bash
git -c core.hooksPath=/dev/null push --no-verify --atomic origin "${refs[@]}"
```

Mirrors the hardening the create-version-commit helper already applies.

tarikbellamine (commented):
Inline review covering the three practical P1 findings from this revision. The PR is structurally much tighter than the previous draft — most of the prior P1/P2/P3 list is resolved. The three findings below are what I'd want to see closed before treating this PR as an actual security boundary rather than preparation for one.

Not inline-commented (out-of-diff or repo-level, but still gating effective value):
- Branch protection still off on `main` / `next`. Until a 'require signed commits' rule exists, the signed-commit work in this PR is preparatory.
- `default_workflow_permissions: write` at repo level → `read`.
- `can_approve_pull_request_reviews: true` → off.
- No `environment:` gate on publish.
- Verify trusted-publishing + provenance per package on the npm registry side.

P2/P3 findings (publish-job privilege split, `compute-pending-tag` shape validation, `commitOid` hex validation, recovery-helper re-verify, `compute-pending-tag.mjs` git hook hardening, hash-check `-maxdepth 1` / `*.mjs` filter, M8/M10/M15 carry-overs) intentionally omitted from this inline pass — happy to add on request.

Rubilmax (commented):
LGTM

0xbulma:
## Local PR Review (post-LGTM follow-up) — :white_check_mark: ready to merge

Reviewed `release-bot-sign-commits` at `545fae52` against `main` (`b46cb87c..545fae52`, 33 files). Deduplicated against the 14 prior reviews + Rubilmax's automated Claude review + maintainer dispositions. The branch evolved substantially since the last reviewer snapshot at `13f1e938` — the privilege split, `package.json` content validation, `RELEASE_BRANCH` allowlist, lease-on-create push, and `env -i` publish hardening all landed and address the substantive prior P1s.

**Verdict: ready to approve and merge.** Neither finding below is blocking — the medium is a PR-scoping / description nit (the change itself is safe for the next release), and the low is a one-line completeness gap on an existing defense-in-depth guard in a code path that has no realistic tampering vector today. Both can land as follow-ups.

| Severity | Count |
|----------|-------|
| Critical | 0     |
| High     | 0     |
| Medium   | 1     |
| Low      | 1     |

### Medium (1) — non-blocking, follow-up suggested

- **[MEDIUM] `.changeset/config.json:15` — Undisclosed packaging-policy change mixed into a hardening PR.** Adding `"bumpVersionsWithWorkspaceProtocolOnly": true` and converting every `packages/*/package.json` peerDep from `workspace:^` to a literal pinned range is a packaging-policy change: Changesets will no longer cascade a major bump in one internal package to its dependents' peer ranges, so each downstream major must now be hand-authored. Per `AGENTS.md` §10 ("`.changeset/config.json` changes — flag for human review on every change"), this should have been explicitly called out in the PR description (it isn't; the description is entirely about signing/hardening), and it is also a §8 "one concern per PR" miss. All eight pinned floors match the currently-published `latest` on npm, so the *next* publish ships byte-equivalent peer ranges — the change is safe for the immediate release but quietly shifts the maintenance contract going forward.
  _Suggestion (follow-up): add a paragraph to the PR description acknowledging the propagation-policy flip, and note the new convention in the release runbook / `AGENTS.md` so future maintainers don't expect cascade behavior._

### Low (1) — non-blocking, follow-up suggested

- **[LOW] `.github/workflows/version-pr.yml:127` — Hook-tamper guards check `.git/hooks/` but not `core.hooksPath`.** The three `Verify write-token preconditions` / pre-commit / pre-PR guards (lines 132-140, 159-167, 181-189) check `find .git/hooks -type f ! -name '*.sample'`, which detects files planted directly under `.git/hooks/` but not a redirected `core.hooksPath` set in `.git/config`. In job 2's flow this is defense-in-depth only — the worktree comes from `actions/checkout` at trusted `github.sha` and no `pnpm install` runs in this job — so there is no realistic path to tampering with `.git/config` before the write-token mint. But the existing guard's stated intent is "detect hook tampering before any privileged git operation," and `core.hooksPath` is the documented way to bypass `.git/hooks/`. Rubilmax raised the same observation on the earlier revision at `version-pr.yml:111` (now-obsolete line); the comment was filed but the cross-step coverage didn't follow through.
  _Suggestion (follow-up): in each of the three guard blocks, add one line after the `find` check —_
  ```bash
  hooks_path="$(git config --local --get core.hooksPath || true)"
  if [ -n "${hooks_path}" ]; then
    echo "::error::core.hooksPath is set: ${hooks_path}"
    exit 1
  fi
  ```

### Deduplicated against past reviews (not raised here)

Resolved by commits since `13f1e938`:
- Privilege split between `pnpm run version` and the write-token step (two-job artifact split).
- `package.json` content validation (`assertSafePackageJsonChange` in `create-version-commit.mjs:572`).
- `RELEASE_BRANCH` allowlist (`readReleaseBranch`, line 511).
- `pushReleaseBranchWithLease` create-only case (`--force-with-lease=ref:` with empty SHA, line 270-290).
- Temp-branch shape validation (`assertTempBranch`, line 531).
- GraphQL `branchName` field (Devin, line 189).

Previously raised, accepted by maintainer:
- `compute-pending-tag.mjs` not validating `manifest.name`/`version`, bare `JSON.parse` — Rubilmax 2026-05-19, LGTM.
- PATH-shadowing of the helper-hash verification — maintainer accepted the `env -i` + `TRUSTED_PATH` patch over a structural split.

Checked, not real:
- Path-traversal in `body_file` from tag — `${tag//\//__}` strips every `/` before the path is built, so no traversal is reachable.

Skipped per trust model the maintainer has documented as acceptable (compromised `pnpm install` deps):
- `$HOME/.npmrc` poisoning via install-time lifecycle scripts forwarded through `env -i HOME=…`.
- `Record privileged-step baseline` running outside `env -i`.
- `apply-version-artifact.mjs` writing the leaf before `realpath`'ing it.

---
*Reviewer note: this is a local pre-PR review run, posted manually rather than via an automated bot.*

0xbulma (approved):

Foulks-Plb (approved):

tarikbellamine (approved):

PR #711, opened 2026-05-20
docs(agents): default to omitting Low-severity findings in PR review

0xbulma:
## Summary

Adds a structural nitpick guard to the shared per-agent contract in `.agents/lib/pr-review-base.md` (Step 5):

> **Default to omitting Low-severity findings.** Only include a Low finding when it would mislead a future reader or compound with another finding on the same code. Polish, wording, and stylistic preferences are not findings.

## Why

Audit of the existing review tooling found a gap. Scope-creep is well-handled (explicit "Stay in scope" clause + per-persona `out-of-scope:` frontmatter + aggregator path filter). But the nitpick guard is essentially absent:

- The base says *"Only actionable findings"* and never defines "actionable".
- The Step 6 aggregator path-filters out-of-scope findings but has no severity floor — Lows pass through untouched.
- Personas enumerate what *makes* a Low finding (e.g. `style-conventions` lists "JSDoc-only diff without a patch changeset") but never tell agents to *withhold* Lows by default.

Concretely, this surfaced after a docs-only PR review where the only thing keeping the output clean was passing `make sure to not scope creep or nitpick` as a manual directive. That should be the default, not user-supplied.

## How

One bullet, one file. Broadcasts to every launched persona (baseline + conditional) via the existing per-agent contract — no per-persona edits required. The clause allows compounding Lows on the same code path when they would mislead a future reader together, so legitimately-related low-severity signal still gets through.

## Test Plan

- [x] `git diff --check`
- [ ] Sanity-run `/pr-review-local` (or `/pr-review-gh`) on a recent PR after this lands and confirm Low-severity polish suggestions are now omitted by default.

## Changeset

Not needed — agent/review documentation only; no published package source changed.

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1779272376842079)


0xbulma:
Superseded by #688 — folding into the morpho-protocol persona PR instead so the persona rename and the Low-severity nitpick guard land together.

PR #712, opened 2026-05-20
feat: add Midnight SDK and shared protocol utilities

0xbulma:
## Summary

Adds the initial `@morpho-org/midnight-sdk` package and the shared `@morpho-org/morpho-ts` protocol utilities it depends on. The PR now covers the implementation, compatibility extraction, tests, docs, and release wiring, not just the original Midnight SDK TIB.

## What changed

- Adds `packages/midnight-sdk` as a Viem-based, framework-free package for Midnight protocol utilities, with pinned Midnight ABIs/contracts, constants, typed errors, market/position/offer models, tick and amount math, fetch helpers, deployless position reads, ratifier helpers, signature/payload/tree utilities, and a `@morpho-org/midnight-sdk/api` subpath for Midnight API book/quote/takeable-offer/validation helpers.
- Moves shared Blue/Midnight primitives into `@morpho-org/morpho-ts`: chain metadata, flat Morpho address/deployment registries, ABI subpath exports, fixed-point math helpers, bigint helpers, shared constants, and typed registry/math errors.
- Keeps existing Blue and main SDK import paths compatible by re-exporting the extracted chain, address, math, constant, and error surfaces from `@morpho-org/morpho-ts`.
- Updates maintained dependents and peer dependents so published packages resolve the extracted shared primitives used by the Blue compatibility layer.
- Adds unit/fetch/API coverage for the new Midnight SDK surfaces and the extracted `morpho-ts` primitives, plus README, TypeDoc, Vitest, Biome, Solidity compilation, package metadata, and changeset wiring.
- Adds `docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md` to document the package boundary and implementation plan.

## Release impact

- `@morpho-org/midnight-sdk`: new minor release.
- `@morpho-org/morpho-ts`, `@morpho-org/morpho-sdk`, and `@morpho-org/blue-sdk`: minor releases for the shared primitive extraction and compatibility surface.
- Maintained direct/peer dependents receive patch bumps where needed by the dependency audit.

## Links

- Linear project: https://linear.app/morpho-labs/project/midnight-sdk-ex-router-sdk-e398faa9785f/overview


docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Considered Alternatives
+
+### Alternative 1: Put Midnight addresses, ABIs, and constants in `blue-sdk`
+
+Extend `@morpho-org/blue-sdk`'s address registry and ABI surface with `midnight`, `midnightMempool`, `takeBundler`, and ratifier contracts.
+
+**Why rejected:** this makes `blue-sdk` a cross-protocol dumping ground. Midnight has its own address lifecycle, typed-data schema, rate/tick math, and ratifier rules. Keeping those in `blue-sdk` would blur package ownership, make future Midnight-only changes look like Blue changes, and contradict the "one reason to exist" package rule. A dedicated `midnight-sdk` mirrors the useful part of the `blue-sdk` pattern without coupling unrelated protocols.
+
+### Alternative 2: Folder named `marketV2/` (version-based, matches `marketV1`)
```
Foulks-Plb: Are we still sticking with Blue/Midnight as the canonical naming, or has that changed?

If Blue/Midnight is still the standard, should we start deprecating the older entities/functions and migrate everything to the new naming to stay consistent?

docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
```diff
@@ -0,0 +1,272 @@
...
+├── rateTick.ts                   # rate↔tick snapping, rate↔WAD-price
+├── typedData.ts                  # OFFER_ROOT_SIGNATURE_TYPES + buildOfferRootTypedData
+├── types.ts                      # Address/Hex-compatible protocol value types
+└── index.ts                      # explicit public barrel
+```
+
+```
+packages/morpho-sdk/src/
+├── actions/
+│   ├── midnight/
```
Foulks-Plb: Oh, it’s called Midnight here?

docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
```diff
@@ -0,0 +1,272 @@
...
+│   │   ├── lendMarket.ts         # pure transaction builder, exported as lendMarket from /midnight
```
Foulks-Plb: For the lendMarket actions, I find “Market” a bit redundant, wdyt?

docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
```diff
@@ -0,0 +1,272 @@
...
+│   │   ├── borrowMarket.ts
+│   │   ├── lendLimit.ts          # consumes pre-signed EIP-712 root
+│   │   ├── borrowLimit.ts        # consumes pre-signed EIP-712 root
+│   │   ├── setIsAuthorized.ts    # explicit per-user prerequisite action
```
Foulks-Plb: setIsAuthorized should be in requirements not a root level

docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
```diff
@@ -0,0 +1,272 @@
...
+├── ratifiers.ts                  # (v,r,s) ↔ EcrecoverRatifier data; ApprovalRatifier path
+├── rateTick.ts                   # rate↔tick snapping, rate↔WAD-price
+├── typedData.ts                  # OFFER_ROOT_SIGNATURE_TYPES + buildOfferRootTypedData
+├── types.ts                      # Address/Hex-compatible protocol value types
+└── index.ts                      # explicit public barrel
+```
+
+```
+packages/morpho-sdk/src/
+├── actions/
```
Foulks-Plb: what about supplyCollateral, withdraw, Repay ?

docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
```diff
@@ -0,0 +1,272 @@
...
+
+- Market orders → `TakeBundler.bundleTake{Buyer,Seller}Assets`
+- Limit orders → `MidnightMempool.submit`
+- Authorization (one-time per user) → `Midnight.setIsAuthorized`
+
+Validation errors are typed classes, never `throw new Error(…)` (CLAUDE.md §2.2, §3). Error classes that are purely protocol-shape failures live in `midnight-sdk` with unprefixed names; action-only failures may live in `morpho-sdk/actions/midnight/errors.ts`. `morpho-sdk/midnight` re-exports them unprefixed, and only root-level `morpho-sdk` exports may alias them if needed to avoid collisions.
+
+**Entity layer (reads on-chain state, returns `{ buildTx, getRequirements }`).** Mirrors `MarketV1` entity. `getRequirements()` returns the existing requirement union, extended with a `TypedDataSignatureRequirement { domain, types, primaryType, message }` variant for limit orders. The caller signs (via viem `walletClient.signTypedData`, wagmi `useSignTypedData`, or any EIP-712 signer), then calls `buildTx({ signature })`. For contract wallets, the entity instead returns an `ApprovalRatifierRequirement` describing the on-chain `approve(root)` call.
```
Foulks-Plb: Good point, we should also plan a fetchMidnight() function or smt like that, similar to fetchAccrualVaultV2().

Foulks-Plb: By the way, is it “accrual” for midnight?

Foulks-Plb:
We should also anticipate the impact of the new bundler in the SDKs.

docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
Foulks-Plb: We should also anticipate bundled actions like supplyCollateral + borrow

docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
Rubilmax: IMO we should start deprecating the v1/v2 naming yes, i got reminded multiple times that we should move to blue/midnight

docs/tibs/TIB-2026-05-20-midnight-instance-actions.md
Rubilmax: Yes, why not? So far it's internal structure

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,495 @@
...
+- market-order lending routes through `MidnightBundles.buyWithAssetsTargetAndWithdrawCollateral` with an empty `CollateralWithdrawal[]`, computes a min-units bound from the user's min rate, and relies on the bundler's per-take skip semantics;
+- supply-only collateral calls `Midnight.supplyCollateral` directly because `MidnightBundles` indexes `takes[0]` and cannot handle an empty take list;
+- limit-order flows use router payload encoding, router validation, Ecrecover-vs-Setter ratifier selection, `setIsAuthorized`, offer-root signing or root approval, and mempool submission.
+
+Keeping those details only in app code makes every downstream integration re-learn the same protocol edges. Folding them into `@morpho-org/morpho-sdk` is too premature: the action/entity shape, Blue-parity naming, and release contract for a high-level SDK integration still need more product and protocol mileage. The immediate decision is narrower: create a dedicated `@morpho-org/midnight-sdk` package that owns stable Midnight protocol utilities and can be used by the app today and by higher-level SDK packages later.
```
spennyp: Worry about _yet another SDK_. Who is the target audience for this SDK before it is mature enough to be included in `@morpho-org/morpho-sdk`? For sure the markets app, but will this be shared externall?

If external, I wonder if there is a way we can use `@morpho-org/morpho-sdk` but have an unstable release or sub-path export instead (maybe not worth the headache).

Alternatively, if it will be internal only until stable, an option is to put it into the `morpho-apps` monorepo as a package and not actually deploy. This also has nice properties in that it is easy for us to iterate upon once the app is consuming (note the actions are still changing with new protocol deploys, router updates, etc).

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,495 @@
...
+- Export typed errors for integrators to pattern-match on instead of throwing plain `Error`.
+- Cover the package with colocated unit tests, property-based tests for tick/price and unit/asset conversion, and viem mock-transport tests for fetch and requirement readers.
+
+**Non-Goals**
+
+- No `@morpho-org/morpho-sdk` source, package export, action, entity, or dependency changes in this TIB.
+- No Blue-style `client.midnight.market(...)` instance in this TIB.
+- No `@morpho-org/blue-sdk` or `@morpho-org/blue-sdk-viem` changes.
+- No React, wagmi, app `ActionFlow`, toast, label, analytics, or UI-state abstractions in `midnight-sdk`.
+- No quoter or orderbook client in this package. The package can define quote/take input shapes and build on-chain `Take[]`, but executable offers still come from the app or a future orderbook SDK.
```
spennyp: I think this SDK should support the existing router SDK functionality (tree construction, payload creation/compression, EIP-712 inputs), and we should depreciate the router SDK. This means router is just an API, and all utilities related to interacting with the protocol live here.

Further, I think the north star should be direct integration with the router API such that we can provide the best DX. A dev shouldn't need to go figure out how to integrate router API, fetch quote, post-process that to construct mempool payload, then post to chain. Ideally this should all be abstracted away in something like: `fetchQuote`, `takeOffers`. I think of this very much like DEX aggregator SDKs (although our SDK will do a bit more such that it doesn't need to trust the router).

spennyp (commented):
Overall aligned on extracting this logic from the app, this part is clearly worth doing an making generalized. Figured this would happen at some point, happy to see this initiative! 

One challenge I have is that for action payload creation, it might be worth considering flipping the default public API from low-level primitives to high level actions. Footguns should be opt-in, not the front door (i.e make it as low friction as possible, reduce the chances a dev can mess something up). 

The default public API I'd want for payload creation: a small set of async functions, one per action (`buildX(client, params)`, etc.), each owning live state fetching, the checks, requirement planning, rounding, and bundle construction, essentially what `markets-v2-app`'s `buildXActionFlow` does today. The caller wires nothing, passes no state, and can't get rounding or bounds wrong because it never touches them.

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
Rubilmax: Currently the `@morpho-org/morpho-sdk` offers "write" action flows and directly depends on `blue-sdk` for that, such that `blue-sdk` only offers "read" utilities and classes for read-only consumers

Because so far we did not agree to merge `blue-sdk` into `morpho-sdk` entirely, we dont want to force the `midnight-sdk` to do so. I would argue this is a broader refactoring that we should indeed consider, but is too early as of today considering we need the midnight sdk as a priority
I guess, with the current version of the TIB, it is too premature to consider implementing it in the fixed rate app. Instead, we are preparing a dedicated TIB to extract action flows from the fixed rate app to `morpho-sdk` in #769, such that the migration for the fixed rate app is the smallest effort possible. The goal is still to be stable from day 1 and release for integrators from day 1, hence why we'd need to keep it external to `morpho-apps`

It does have the benefit of easing the "prerelease" because we wont be advertising for this SDK and won't include it into `morpho-sdk` until it's proven stable

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
Rubilmax: Completely aligned on this. As a first step though, we need to focus on make orders which require less trivial build/encoding/submission utilities to connect to the router api. Then we will be able to include every other routes of the router api to improve DX

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,587 @@
...
+- `Take`
+- `TokenPermit`
+- `CollateralSupply`
+- `CollateralWithdrawal`
+- `MarketState`
+- `Position`
+- `RatifierInfo`
+- `MidnightAddresses`
+- `MidnightAddressRegistry`
+- quote-facing raw input shapes for converting app/router responses into `Offer` and `Take`
```
yum0e: nit: an update from this morning, we have renamed Take into TakeableOffer, it should not impact the first scope for the mempool payload but it will impact the quoter endpoint when we will support it

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,587 @@
...
+`midnight-sdk` exports Midnight-specific constants that app actions currently need or will need to
+validate inputs:
+
+- `CBP`
+- `MAX_TICK`
+- `PRICE_ROUNDING_STEP`
+- `DEFAULT_TICK_SPACING`
+- `MAX_COLLATERALS`
+- `MAX_COLLATERALS_PER_BORROWER`
+- `MAX_OFFERS_PER_TREE`
```
yum0e: there is no protocol limit on this, only on router and we will be increasing it as we have more traffic

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,587 @@
...
+Rounding names should match local SDK convention: `"Up"` and `"Down"`. Tests must cover the app-known boundary cases:
+
+- borrow market orders round max units up for `MidnightBundles.supplyCollateralAndSellWithAssetsTarget`;
+- lend market orders round min units down for `MidnightBundles.buyWithAssetsTargetAndWithdrawCollateral`;
+- `TickLib` prices are rounded to `PRICE_ROUNDING_STEP`;
+- `TickLib.priceToTick` returns the lowest spacing-aligned tick with price greater than or equal to the input price.
+
+### Offer and Take Utilities
+
+`OfferUtils.buildTakesFromOffers` should replace the app-local conversion in
```
yum0e: nit: as said above you can name `TakeableOffer` the struct `{units, offer, ratifierData}`

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,587 @@
...
+- validate a consistent market when callers choose to enforce it before encoding a bundle.
+
+Make-offer helpers should support the limit-order path without importing app code:
+
+- `OfferUtils.buildOffer` builds `Offer` values from market, side, tick, max assets, expiry, group, callback, receiver, and ratifier inputs;
+- generate or accept a group id via an injected random byte source;
+- `OfferUtils.buildOfferGroup` builds multiple offers with one shared group id, preserving order and validating only protocol group mechanics;
+- `OfferUtils.validateOfferGroup` enforces non-empty groups, same maker, same group, same side, same unit/asset caps with exactly one non-zero cap, and same loan token;
+- expose offer payload, Merkle root, proof, ratifier-data, root-approval, mempool-submission, and router-validation helpers behind stable SDK functions;
+- validate maker payloads through `MidnightRouterApi.validateMempoolPayload` or `MidnightRouterApi.validateMempoolItems`; the implementation uses a lightweight `fetch` boundary and does not add a runtime `@morpho-dev/router` dependency;
```
yum0e: you can call it `MidnightApi`, no need to name it with the router naming, integrators should not know our internal services that much

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,587 @@
...
+- keep offer publication/submission onchain through `OfferPayloadUtils.buildMempoolSubmissionCall`; the current Router API does not expose a submit endpoint.
```
yum0e: nit but I find the naming very dense, we might discuss later in details the exact interfaces, dont know if it is the right moment

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,587 @@
...
+### Signatures, Payloads, and Router Validation
+
+Make-offer flows need reusable utilities, but the SDK should not own a wallet or a UI wizard. It should expose grouped helpers:
+
+- `RatifierUtils.getRatifierInfo`
+- `OfferPayloadUtils.buildOfferPayload`
+- `OfferPayloadUtils.buildOfferTreeRoot`
+- `OfferPayloadUtils.buildOfferProof`
+- `OfferPayloadUtils.buildEcrecoverRatificationTypedData` or `OfferPayloadUtils.signEcrecoverRatification` with an injected `signTypedData` callback;
+- `OfferPayloadUtils.encodeEcrecoverRatifierData`
```
yum0e: to us, you should still be able to generate ratifierData without any notion of Payload. Same for offers, trees. Payload should not be tied to this, instead payload is the final way to interact onchain imo. It should completely live at the edge for devx (just before sending onchain, just after getting a log). It is just a communication mean. 

here is our current devx
```typescript
import { Payload, Group, Tree, EcrecoverRatifier, Api } from "@morpho-dev/router";

const api = Api.init(); // defaults to https://router-alpha.morpho.dev

const group = Group.create(offersWithoutGroup);
const tree = Tree.create({ groups: [group] });

// Validate before asking the wallet to sign — the router endpoint only
// inspects offer contents, so `api.validate` encodes the request with empty
// per-leaf ratifierData and a rejected batch never costs a signature.
const { issues } = await api.validate({ tree, chainId });
if (issues.length > 0) {
  throw new Error(`validation failed: ${issues.map((issue) => issue.rule).join(", ")}`);
}

const items = await EcrecoverRatifier.ratify({
  tree, wallet, chainId,
  // verifyingContract defaults to the chain's ecrecoverRatifier address
});
const payload = await Payload.encode(items);

await wallet.sendTransaction({
  to: mempoolAddress,
  data: payload,
});
```

yum0e: feels really bloated to put everything in the OfferPayloadUtils, it does not welcome composition

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,587 @@
...
+
+`MidnightRouterApi` is the router HTTP boundary. It accepts an optional string-or-`URL` `baseUrl` for the router API base URL, defaulting to `https://router.morpho.org`, plus injected `fetch` and `request` options for headers, abort signals, credentials, cache, and similar fetch settings. Custom base URLs are parsed with the standard `URL` API, normalized by clearing search/hash, and joined with SDK-owned endpoint paths. SDK-controlled fields remain fixed: URL path/query, HTTP method, JSON body, `Content-Type: application/json` on POST requests, and an exact `sdk-version` header equal to the `@morpho-org/midnight-sdk` package version. Validation returns SDK camelCase data `{ payload, valid, issues }`, where `valid` is derived from `issues.length === 0`. Rules are returned as `{ cursor, data }` with camelCase fields such as `chainId`, `callbackType`, `minTick`, `tickSpacing`, and `allowedLltvs`.
```
yum0e: you should be hitting api.morpho.org instead, router domain name services will not be advertized

yum0e: for the rules endpoint, I will take an additional look to have a format that can survive time, you can expect some changes until end of the week. I want to nail the format to be future proof

Rubilmax: which base url then?

Rubilmax: Removing interfaces specs for router api endpoints as it is premature

spennyp: https://api.morpho.dev/midnight/docs

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,613 @@
...
+  delegates to `ConsumableUnitsLib`
+- `fetchRatifierInfo`, which reads bytecode and delegates classification to
+  `RatifierUtils.getRatifierInfo`; it must preserve the EIP-7702 designator rule from the app: no
+  code, `0x`, or `0xef0100...` routes through Ecrecover; other code routes through Setter.
+- `fetchIsRootCanceled`
+- `fetchIsRootRatified`
+
+Fetch helpers must not mutate state, sign, submit transactions, or import UI/app code.
+
+### Requirement Plans
```
Rubilmax: Get rid of requirement planners

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,613 @@
...
+    +-- index.ts
+    +-- types.ts
+    +-- market/
+    |   +-- CollateralParams.ts
+    |   +-- Market.ts
+    |   +-- MarketUtils.ts
+    |   +-- Position.ts
+    |   +-- index.ts
+    +-- offers/
+    |   +-- Offer.ts
```
Rubilmax, line 95: A correct offer is currently bound to the type of ratifier that is deployed now, but maybe later we'll have multiple types of offers due to having multiple types of ratifiers thus multiple ways to validate offers

Double-check offer shape matches onchain so it is composable later on

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,613 @@
...
+    |   +-- index.ts
+    +-- requirements/
+    |   +-- approval.ts
+    |   +-- authorization.ts
+    |   +-- requirements.ts
+    |   +-- orderPlans.ts
+    |   +-- index.ts
+    +-- signatures/
+        +-- OfferTree.ts
+        +-- Payload.ts
```
Rubilmax, line 124: Payload does not necessarily corresponds to a readable/known offer

Rubilmax: Do we consider the current state of "Payload" as defined by the router api generic enough to never fragment due to usage preference?

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
Rubilmax: e.g. RateOffer

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,613 @@
...
+    |   +-- TakeAmountsLib.ts
+    |   +-- TickLib.ts
+    |   +-- index.ts
+    +-- fetch/
+    |   +-- midnight.ts
+    |   +-- index.ts
+    +-- queries/
+    |   +-- GetConsumableUnitsInputs.ts
+    |   +-- GetPosition.ts
+    |   +-- index.ts
```
Rubilmax, line 113: Nice to have

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,613 @@
...
+    +-- router/
+    |   +-- MidnightRouterApi.ts
+    |   +-- index.ts
+    +-- requirements/
+    |   +-- approval.ts
+    |   +-- authorization.ts
+    |   +-- requirements.ts
+    |   +-- orderPlans.ts
+    |   +-- index.ts
```
Rubilmax: tier3?

docs/tibs/TIB-2026-05-20-midnight-sdk-utilities.md
```diff
@@ -0,0 +1,613 @@
...
++-- contracts/
+|   +-- GetConsumableUnitsInputs.sol
+|   +-- GetPosition.sol
+|   +-- interfaces/
+|       +-- IMidnight.sol
++-- package.json
++-- tsconfig*.json
++-- src/
+    +-- abis.ts
+    +-- addresses.ts
```
Rubilmax: Move in morpho-ts to merge everything

packages/midnight-sdk/src/constants.ts
```diff
@@ -45,7 +45,7 @@ export const SETTLEMENT_FEE_BREAKPOINTS = [
  * console.log(MAX_TICK);
  * ```
  */
-export const MAX_TICK = 5820n;
+export const MAX_TICK = 6744n;
```
0xbulma, line 46: Was this reviewed ? Not sure to understand the implications.

packages/midnight-sdk/src/constants.ts
```diff
@@ -57,7 +57,7 @@ export const MAX_TICK = 5820n;
  * console.log(PRICE_ROUNDING_STEP);
  * ```
  */
-export const PRICE_ROUNDING_STEP = 1_000000000000n;
+export const PRICE_ROUNDING_STEP = 100_000000000n;
```
0xbulma, line 58: Same here

packages/midnight-sdk/src/constants.ts
Rubilmax: This is the correct value taken from TickLib: https://github.com/morpho-org/midnight/blob/main/src/libraries/TickLib.sol#L6

Unfortunately flagged unexpectedly by a prompt to review the tick spacing...

packages/midnight-sdk/src/constants.ts
Rubilmax: This is the correct value taken from TickLib: https://github.com/morpho-org/midnight/blob/main/src/libraries/TickLib.sol#L8

Unfortunately flagged unexpectedly by a prompt to review the tick spacing...

Foulks-Plb (dismissed):

packages/midnight-sdk/src/constants.ts
Rubilmax: So it appears they've changed 4d ago: https://github.com/morpho-org/midnight/commit/370e3a367baaa60ffa42deaa6a8fed9143aab639#diff-38303ef8a8f10e2b176f36cb72ee11168c1e10af793ef1217a6ad94b555c0b05

packages/midnight-sdk/src/constants.ts
Rubilmax: So it appears they've changed 4d ago: https://github.com/morpho-org/midnight/commit/370e3a367baaa60ffa42deaa6a8fed9143aab639#diff-38303ef8a8f10e2b176f36cb72ee11168c1e10af793ef1217a6ad94b555c0b05

packages/midnight-sdk/src/constants.ts
```diff
@@ -0,0 +1,226 @@
...
+ * Maximum Midnight tick.
+ *
+ * @example
+ * ```ts
+ * import { MAX_TICK } from "@morpho-org/midnight-sdk";
+ *
+ * console.log(MAX_TICK);
+ * ```
+ */
+export const MAX_TICK = 6744n;
```
yum0e, line 46: The SDK currently uses `MAX_TICK = 6744` and `PRICE_ROUNDING_STEP = 1e11`, while the current router uses `MAX_TICK = 5820` and `PRICE_ROUNDING_STEP = 1e12`.

This means the SDK can produce offers and prices that the router will not accept. The router is currently one contract version behind here, so for this SDK to be usable with the router, the constants and tick math should match the router’s current constants/logic.

packages/midnight-sdk/src/signatures/Payload.ts
```diff
@@ -0,0 +1,754 @@
...
+ * Maximum gzip-compressed item bytes accepted by the payload codec.
+ *
+ * @example
+ * ```ts
+ * import { Payload } from "@morpho-org/midnight-sdk";
+ *
+ * console.log(Payload.MAX_COMPRESSED_ITEMS_BYTES);
+ * ```
+ */
+export const MAX_COMPRESSED_ITEMS_BYTES = 1_000_000;
```
yum0e: The SDK currently allows `MAX_COMPRESSED_ITEMS_BYTES = 1_000_000`, then adds the payload header and attribution suffix on top. This makes the full payload exceed the router/onchain mempool cap of `1_000_000` bytes.

The full wire payload must be capped at `1_000_000` bytes. So `MAX_PAYLOAD_BYTES` should be `1_000_000`, and `MAX_COMPRESSED_ITEMS_BYTES` should be derived from it:

```ts
MAX_COMPRESSED_ITEMS_BYTES =
  MAX_PAYLOAD_BYTES - HEADER_BYTES - MAX_ATTRIBUTION_SUFFIX_BYTES
```

packages/midnight-sdk/src/signatures/TreeUtils.ts
```diff
@@ -0,0 +1,720 @@
...
+    if (structs.length === 0) {
+      throw new InvalidTreeError("Tree must not be empty.");
+    }
+
+    const offerStructs = padOfferStructs(structs);
+    const leaves = offerStructs.map(OfferUtils.hashStruct);
+    assertLeafOffers(offerStructs, leaves);
+
+    const height = Math.log2(offerStructs.length);
+    if (height > 20) throw new InvalidTreeHeightError(height);
```
yum0e, line 581: Open question: the SDK allows trees up to height `20`, while the current router hard-caps decoded payloads at `256` offers.

Do we want the SDK to prevent users from building/signing payloads that the current router will reject, or should the SDK keep the onchain limit and let router submission fail when the payload exceeds the router cap?

packages/midnight-sdk/src/offers/Offer.ts
```diff
@@ -0,0 +1,578 @@
...
+   *   group: apiGroup,
+   *   expiry: 2_100n,
+   *   ratifier,
+   *   maxAssets: 100n,
+   * });
+   * // offer satisfies Offer
+   * ```
+   */
+  public static create(params: BuildOfferParams): Offer {
+    const validated = OfferUtils.validateOfferParams(params);
```
yum0e, line 428: The SDK currently lets users create offers/markets that the router will reject later. Examples:

- empty collateral list
- unsorted or duplicate collaterals
- invalid LLTV / maxLif values
- `expiry > maturity`
- maturity not aligned to 15:00 UTC

Some of these are caught later during payload encoding, but `Offer.create`, `Tree.create`, and signing flows can still happen first. Should these objects be valid-by-construction, or is the current “validate later during payload encoding” behavior intentional?

packages/midnight-sdk/src/constants.ts
Rubilmax: Just a matter of desync, not blocking this PR to me since we will reach the same state at some point (and btw it's fine that the sdk produces offers that the router does not accept)

packages/midnight-sdk/src/signatures/TreeUtils.ts
Rubilmax: The SDK should keep the onchain limit and let router submission fail when the payload exceeds the router cap
The router is a constraint layer on top

packages/midnight-sdk/src/offers/Offer.ts
Rubilmax: Highly depends on the integration. If router validation happens before signing, everything's fine. Router rules can change over time and we dont want the sdk to be coupled to them

packages/midnight-sdk/src/offers/OfferUtils.ts
```diff
@@ -0,0 +1,771 @@
...
+   *   tick: 5_000n,
+   *   expiry: 3_600n,
+   *   ratifier: "0x0000000000000000000000000000000000004000",
+   *   maxUnits: 100n,
+   * });
+   * const offers = OfferUtils.validateOfferGroup({ offers: [offer] });
+   * console.log(offers.length);
+   * ```
+   */
+  export function validateOfferGroup(
```
yum0e, line 727: The SDK group validation does not fully match the router group rules. The router also requires offers in one group to share callback address/data, and offers in the same group + same market must not have overlapping `[start, expiry]` windows.

Today the SDK can build groups that the router will reject later. Should `Group.create` mirror the router group rules, or should this stay a `tree.mempoolValidate`-only check?

packages/midnight-sdk/README.md
```diff
@@ -0,0 +1,246 @@
...
+    tick: 5_000n,
+    maxAssets: parseUnits("250000", 6),
+  });
+  const lend100kUsdc = Offer.create({
+    ...commonOffer,
+    buy: true,
+    tick: 5_012n,
+    maxAssets: parseUnits("100000", 6),
+  });
+  const groupedLendOffers = Group.create([lend250kUsdc, lend100kUsdc]);
```
yum0e: This example builds an invalid group. The two grouped offers have different `maxAssets`, and they also share the same market/active window.

The router will reject this group, and `Group.create` should reject it as well because grouped offers must share the same cap mode and value. The example should either make these standalone offers or use a valid grouped-offer pattern.

packages/midnight-sdk/src/signatures/SetterRatifierUtils.ts
```diff
@@ -0,0 +1,267 @@
...
+   *   maxUnits: 100n,
+   * });
+   *
+   * const items = SetterRatifierUtils.ratify({
+   *   tree: Tree.create([offer]),
+   * });
+   * console.log(items.length);
+   * ```
+   */
+  export function ratify(params: {
```
yum0e, line 279: `SetterRatifierUtils.ratify` packages every offer in the tree without checking that the tree uses one maker and one ratifier.

The router rejects mixed-maker and mixed-ratifier payloads, and Setter root approval is keyed by maker. Should this mirror the Ecrecover path and reject mixed maker/ratifier trees before producing payload items?

packages/midnight-sdk/README.md
```diff
@@ -0,0 +1,246 @@
...
+const midnight = addresses[chainId].midnight!;
+const midnightBundles = addresses[chainId].midnightBundles!;
+const tokenPermitNone = { kind: 0, data: "0x" } as const;
+
+async function buildAskQuoteTakes(marketId: Hash, targetBuyerAssets: bigint) {
+  const quote = await MidnightApi.fetchBookQuote({
+    marketId,
+    side: "asks",
+    assets: targetBuyerAssets,
+    slippage: "0.25",
```
yum0e: This slippage value is not accepted by the router. The router allows slippage from `0.1` to `100` with at most one decimal place, so `"0.25"` will fail validation. Use something like `"0.3"` or `"0.5"` here.

packages/midnight-sdk/src/signatures/Payload.ts
Rubilmax: applied in #810

packages/midnight-sdk/src/offers/OfferUtils.ts
Rubilmax: It's fine to stay specific to mempoolValidate

packages/midnight-sdk/src/offers/Offer.ts
Rubilmax: applied the unsorted or duplicate collaterals check in #810

packages/midnight-sdk/README.md
Rubilmax: `Group.create` is already rejecting, but `toStructs` is not because it's only an encoding helper. Im updating its jsdoc to specify it

packages/midnight-sdk/src/signatures/SetterRatifierUtils.ts
Rubilmax: A mixed-maker Setter tree is onchain-valid if every maker has approved the same root. A blanket `SetterRatifierUtils.ratify` rejection for mixed makers would reject something the contracts can accept.

What is actually missing in the SDK is narrower:

1. `SetterRatifierUtils.ratify` should not imply that one root approval is enough for a mixed-maker tree. The docs should say approval is per maker, so mixed-maker trees require one approval per maker for that root.

2. If we want an SDK rejection, it should be about ratifier-family/address misuse, not mixedness by itself. `SetterRatifierUtils` emits Setter-formatted ratifier data, so offers pointing at an Ecrecover ratifier will fail onchain. To enforce that purely, the helper would need an expected Setter ratifier address parameter and reject offers whose `offer.ratifier` differs from it.

Rubilmax: applied in #810

packages/midnight-sdk/README.md
Rubilmax: applied in #810

packages/midnight-sdk/README.md
Rubilmax: exmaple updated in #810

packages/midnight-sdk/src/signatures/EcrecoverRatifierUtils.ts
```diff
@@ -0,0 +1,957 @@
...
+   * });
+   *
+   * const digest = EcrecoverRatifierUtils.digest({
+   *   tree: Tree.create([offer]),
+   *   chainId: 8453n,
+   * });
+   * console.log(digest);
+   * ```
+   */
+  export function digest(params: EcrecoverRatifierTypedDataParams) {
```
Foulks-Plb, line 571: **[security] EIP-712 digest is unverified against `typedData()` and the contract.** `digest()` and `typedData()` (:501) are two independent implementations of the same hash that must agree — and match the on-chain ratifier — yet the only test (`EcrecoverRatifierUtils.test.ts:223-230`) just asserts a 64-hex shape, and `treeTypeHashes` (:30) + the typehashes (`constants.ts:162-202`) are hardcoded with no derivation test. A single wrong/drifted constant silently invalidates every maker signature.

Fix (doable without deployment): assert `digest === hashTypedData(typedData)` for height 0/1/≥2, derive the typehashes in TS (`keccak256` of the canonical type strings) and compare to the table, and pin a contract-captured digest fixture before stable.

packages/midnight-sdk/src/api/MidnightApi.ts
```diff
@@ -0,0 +1,840 @@
...
+    const input = params;
+    const response = await requestMidnightApi({
+      ...input,
+      method: "POST",
+      path: "mempool/validate",
+      query: {
+        timestamp: input.timestamp,
+      },
+      body: {
+        chain_id: input.chainId,
```
Foulks-Plb, line 468: **[consistency] `chainId` is forwarded to the API unchecked.** Unlike the RPC path (where `getChainAddress` throws `UnsupportedChainIdError`), the API methods (`validateMempoolPayload` here, `fetchBooks` via `chain_ids`) forward the chain id with no validation, so a wrong id can return `valid: true` for a chain you won't broadcast on. Validate `chainId` against the supported-chain registry before the request and throw the same typed error as the RPC path.

Rubilmax: The chainId will be caught unsupported at some point before the offer is actually broadcasted anyway. This is overkill

packages/midnight-sdk/src/signatures/Payload.ts
```diff
@@ -0,0 +1,766 @@
...
+    throw new DecodeError(
+      "invalid offer bytes: expiry must be before or equal to maturity",
+    );
+  }
+  if (offer.tick > MAX_TICK) {
+    throw new DecodeError(
+      `invalid offer bytes: tick exceeds maximum ${MAX_TICK}`,
+    );
+  }
+  if ((offer.maxUnits === 0n) === (offer.maxAssets === 0n)) {
```
yum0e, line 499: The payload codec still accepts buy offers with a non-zero `receiverIfMakerIsSeller`. Router/core now rejects the same struct during payload decode, and the synced contract enforces `UnusedReceiverMustBeZero` on `take()`, so an SDK-encoded payload can pass local encode/decode but fail router validation or onchain.

Can we mirror the router check here in `assertApiValidOfferStruct`?

```ts
if (offer.buy && offer.receiverIfMakerIsSeller.toLowerCase() !== ZERO_ADDRESS) {
  throw new DecodeError(
    "invalid offer bytes: receiverIfMakerIsSeller must be zero for buy offers",
  );
}
```

This keeps the payload boundary aligned with the current protocol/router behavior.

bilbo524 (approved):

Foulks-Plb (approved):

PR #764, opened 2026-06-03
feat(evm-simulation)!: canonical per-account assetChanges

Foulks-Plb:
## Motivation

`assetChanges` had a different shape per backend: Tenderly returned an opaque per-call blob (`SimulationCall.assetChanges?: unknown`), while `eth_simulateV1` returned nothing at all. The field was backend-dependent and silently lost on the fallback path — impossible to read without knowing which backend had run.

## Solution

Both backends now normalize to a single canonical shape, exposed at the bundle level via `SimulationResult.assetChanges`.

- **New exported types**: `AssetChange` (`{ token, symbol?, decimals?, diff }` — one asset's net delta; native ETH uses viem's `ethAddress` sentinel) and `AccountAssetChanges` (`{ account, changes }`).
- `assetChanges` is the net per-token delta **grouped by account** over the whole bundle — the sender and every counterparty (the zero address is kept for mints/burns), sorted by address for deterministic, cross-backend output.
- **Tenderly**: the per-transfer payload is now schema-validated and reduced to `AccountAssetChanges[]`.
- **`eth_simulateV1` (fallback)**: `assetChanges` is derived from the emitted transfer logs plus each transaction's top-level `value` (payer debited, recipient credited, under `ethAddress`). Native ETH moved through internal calls (e.g. a `WETH.withdraw` refund) emits no log and is not captured on the fallback — use Tenderly for full native-ETH accounting.

## Architecture

```mermaid
graph TD
  Tenderly["Tenderly RPC<br/>per-transfer asset changes"] --> Norm["Net &amp; group by account"]
  V1["eth_simulateV1 fallback<br/>transfer logs + tx value"] --> Norm
  Norm --> Out["SimulationResult.assetChanges<br/>one AccountAssetChanges per account"]
```

## Breaking change

The opaque per-call `SimulationCall.assetChanges?: unknown` field is removed — read the bundle-level `SimulationResult.assetChanges` instead. Shipped as a `major` changeset.

## Notes

- The fallback path only needs an RPC that supports `eth_simulateV1`; no tracing or deployless calls are used, so it works on any compliant node.
- Unit + fork tests (at a pinned block), `tsc`, and Biome all pass. No other package depends on `evm-simulation`.

[View Slack thread](https://morpho-labs.slack.com/archives/C0B1C0ULKMK/p1780576356626329)


Foulks-Plb:
@prd-carapulse[bot] review the PR.
Focus devex, precision and consistency.

Foulks-Plb:
@prd-carapulse[bot] review the PR.
Focus devex, precision and consistency.

Foulks-Plb:
@prd-carapulse[bot] review the PR. Focus devex, precision and consistency.

Addressed the prior round:
- Fallback now derives native ETH from top-level tx `value` (under `ethAddress`); documented that ETH received via internal calls (e.g. WETH.withdraw refunds) is not captured there.
- `assetChanges` sorted by token address on both backends.
- Fork specs split into an opt-in `evm-simulation-fork` Vitest project (+ `test:fork` script) so the default unit project no longer needs `MAINNET_RPC_URL` at import time.

Foulks-Plb:
@codex review

Foulks-Plb:
@codex review

Foulks-Plb:
@prd-carapulse[bot] review the PR. Focus devex, precision and consistency.

packages/evm-simulation/src/simulate/simulate.spec.ts
```diff
@@ -43,6 +43,7 @@ const SPENDER: Address = "0x3333333333333333333333333333333333333333";
 function makeSuccessResult(logs: RawLog[] = []): RawSimulationResult {
   return {
     calls: [{ logs, status: true, returnData: "0x", gasUsed: 0n }],
+    assetChanges: [],
```
jinmel: can you add a test case where it has non empty `assetChanges`?

Foulks-Plb: Good call, thank you @jinmel

Rubilmax (approved):

PR #795, opened 2026-06-15
docs: specify EVM simulation state and safety checks

Foulks-Plb:
## Motivation

Transaction creation is moving to an independent write API. Both it and the frontend need common transaction checks. Tenderly recently omitted an asset change, exposing a provider dependency that is not clearly disclosed to users.

## Solution

Adds a [Proposed TIB](docs/tibs/TIB-2026-09-18-evm-simulation-state-override-verification.md) by @Foulks-Plb and @jinmel (revising the earlier 2026-06-15 draft in this PR) defining:

- `eth_simulateV1` only; remove Tenderly and fallback.
- Calldata-derived asset, permission, position end-state/diff and market checks across 24 flows.
- Pending authorizations (ERC-20 approval, ERC-2612 permit, Permit2, Morpho authorization) become explicit `stateDiff` overrides proven by read-back; no prepended approvals and no `ecrecover` override. Preview simulates the no-signature build; final accepts no authorizations. API uses preview; frontend uses both.
- Typed per-operation `OperationLimit` bounds with unit-fixed fields instead of a generic `ranges` metric, plus wallet, slippage, LLTV-buffer and signature-lifetime limits.
- Existing API/result compatibility, a Tenderly deprecation minor before the major, and a migration guide for the behavioral breaks.

Documentation only. Layout, coverage and diff checks pass; repository tests not run.


docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Layer A — Static decode (declared intent)
+
+What the user is *being asked to authorize*, checked **before** the bundle is sent.
+
+- **Per-chain trust-list of spenders/operators.** `GeneralAdapter1`, the bundler3 adapters, and `Permit2`, scoped **by `chainId`** (sourced from `@morpho-org/blue-sdk`, same pattern as the bundler3 set). An `approve`, an EIP-2612 `permit`, or a Permit2 `SignatureTransfer` whose **spender** falls outside the chain's trust-list is flagged.
+- **Intercept at the signature request, not only the calldata.** Off-chain signatures never produce a `Transfer` log, so a balance-only or calldata-only view misses them entirely. We decode the **typed-data being signed** — EIP-2612 `permit` and Permit2 `PermitTransferFrom` / `PermitBatchTransferFrom` — and check the granted `spender`/`amount`.
+- **Recursive batch decode.** The sensitive action is usually nested. We unwrap **Multicall3**, **Safe `multiSend`**, the **bundler3 multicall**, and **ERC-4337 `userOp` / `handleOps`**, recursively, and run every leaf call through the same checks.
```
0xbulma, line 61: **[High] Decode bundler3 `reenter()`/`callbackHash` callback sub-bundles, not just the outer `Call[]`.**

bundler3's sensitive leaves usually don't live in the outer multicall array — they arrive via `reenter(Call[])` during a callback (flashloan, Paraswap buy/sell) and are committed in the outer call only as an opaque `callbackHash` (bytes32). A multicall-array walk sees the hash, not the committed sub-calls, so it declares the bundle clean while the dangerous action hides in the callback — the exact "hidden inside a batch so the outer `to` looks trusted" threat from the Context section. Layer A should reconstruct the callback sub-bundle and verify it against the committed hash. Also handle `skipRevert=true` leaves (the intent decoder may assume a call executes when it can silently no-op).

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Proposed Solution
+
+Two verification layers, added as new pipeline stages, both consuming what the backends already return (decoded input for static, `stateDiff` + full logs for dynamic). Each check is independent, returns a **typed finding**, and is **opt-in per chain** via existing `SimulationConfig`.
+
+### Layer A — Static decode (declared intent)
+
+What the user is *being asked to authorize*, checked **before** the bundle is sent.
+
+- **Per-chain trust-list of spenders/operators.** `GeneralAdapter1`, the bundler3 adapters, and `Permit2`, scoped **by `chainId`** (sourced from `@morpho-org/blue-sdk`, same pattern as the bundler3 set). An `approve`, an EIP-2612 `permit`, or a Permit2 `SignatureTransfer` whose **spender** falls outside the chain's trust-list is flagged.
```
0xbulma: **[High] Exclude the Bundler3 *dispatcher* from the approval trust-list — trust only adapters.**

Per bundler3's security model: *"Bundler3 receives no approvals; it's a dispatcher and could steal funds. Individual adapters are safe to approve."* An `approve`/Permit2 grant whose spender is the bundler3 dispatcher is a red flag; the same grant to `GeneralAdapter1` is expected. Reuse risk: the retention trust-set is built from all `addresses.bundler3.*` values (`packages/evm-simulation/src/simulate/pipeline/bundler-retention.ts:44-48`), which *includes* the dispatcher itself. Reusing that set verbatim as the approval trust-list would bless approvals to the one address that must never hold them.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+- **Intercept at the signature request, not only the calldata.** Off-chain signatures never produce a `Transfer` log, so a balance-only or calldata-only view misses them entirely. We decode the **typed-data being signed** — EIP-2612 `permit` and Permit2 `PermitTransferFrom` / `PermitBatchTransferFrom` — and check the granted `spender`/`amount`.
```
0xbulma: **[High] Intercept EIP-7702 authorization tuples, not only EIP-2612 / Permit2.**

The most complete "zero balance change, future drain" is an EIP-7702 authorization delegating the EOA's code to an attacker contract — it hands over the *entire account*, dwarfs any single approval, and emits no Transfer/Approval log. 4337 / smart-account flows are explicitly in scope and the repo has zero 7702 handling today (net-new). An intent layer that catches off-chain grants but ignores 7702 has a hole exactly where the stakes are highest.

0xbulma: **[Medium] Permit2 coverage omits AllowanceTransfer (standing) mode and any amount/deadline policy.**

Listed: `SignatureTransfer`/`PermitTransferFrom`/`PermitBatchTransferFrom` (one-shot). Missing: Permit2 `IAllowanceTransfer.permit`/`permitBatch` (`approve2`) — the *standing* allowance with `amount`+`expiration`+`nonce`, the more dangerous future-drain vector, and one bundler3 supports. Also nothing checks unlimited (`type(uint256).max`) amounts or far-future deadlines even to a trusted spender. Trust-list membership ≠ safe amount/duration. Reuse `getPermit2PermitTypedData` from `blue-sdk-viem/src/signatures/permit2.ts`.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+- **Realized amount, not encoded amount.** Measure what the receiver actually got, so **fee-on-transfer / rebasing** tokens are accounted correctly.
+- **Slippage & fees within tolerance**, and **PublicAllocator fee is correct** (vault/market reallocation).
+- **No asset leaves to an unknown / unexpected address.** Every outflow recipient is either the user or an explicitly-allowed destination.
+- **Legitimate non-`from` recipients are allowed.** A receiver different from the sender is expected and fine for: vault `deposit`/`mint` → shares to `receiver`, bridges, smart-account / 4337 flows, and callback adapters. These must pass, not trip the "unknown recipient" check.
+- **State diff, not just balance diff.** A bundle can finish with **balance delta = 0 yet leave a standing allowance or authorization** → a future drain. We inspect **storage/state diffs and events** — `Approval` (ERC20 + Permit2), `AuthorizationSet` (Morpho), operator approvals — and assert that **no approval or authorization is granted as a side effect** beyond what the declared intent required.
```
0xbulma, line 74: **[High] Define the residual-allowance baseline — this check's correctness hinges on it.**

bundler3 flows legitimately rely on a standing Permit2 allowance to `GeneralAdapter1` (`approve2`/AllowanceTransfer) that persists across bundles by design. "No allowance/authorization granted as a side effect beyond what the declared intent required" needs an explicit policy separating expected residual adapter allowances from malicious side-effect grants — otherwise the check is either noisy (flags every legit standing allowance) or toothless. This is the single most important unspecified decision in the design; promote it to Open Questions.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+- **Recursive batch decode.** The sensitive action is usually nested. We unwrap **Multicall3**, **Safe `multiSend`**, the **bundler3 multicall**, and **ERC-4337 `userOp` / `handleOps`**, recursively, and run every leaf call through the same checks.
+- **Validate the inner actions of a trusted router.** A trusted entry point (e.g. `GeneralAdapter1`) can still be instructed to do an untrusted thing. For our routers we decode and validate the **internal action list**, not just the outer address.
+
+### Layer B — Dynamic simulation (effective result)
+
+What *actually happened*, from the post-simulation **state diff + events** — not balances alone.
+
+- **No net value loss, across every asset type.** Net diff ≥ expected, computed over **ERC20, native ETH, ERC-4626 vault shares, and debt/LP positions** — not only ERC20. (Today only ERC20 is parsed.)
```
0xbulma: **[Medium] "No net value loss across asset types" has no price basis — it can only check against the caller's declared expectation.**

For any asset-crossing flow (Paraswap swap, supplyCollateral A → borrow B), "net diff ≥ expected across all asset types" isn't computable without an exchange rate, but USD/oracle valuation is a Non-Goal. So the guarantee reduces to "matches the integrator-supplied expected amounts" — a malicious or buggy *declared expectation* passes. State plainly that Layer B verifies *conformance to declared intent*, not *fairness*, so integrators don't over-trust it.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+- **Realized amount, not encoded amount.** Measure what the receiver actually got, so **fee-on-transfer / rebasing** tokens are accounted correctly.
```
0xbulma: **[Medium] "Realized amount / every asset type" breaks on hostile ERC20s beyond fee-on-transfer/rebasing.**

Also in scope: ERC-777 `tokensReceived`/`tokensToSend` hooks (a reentrancy vector that also double-counts Transfer logs); double-entry-point tokens (one balance, two proxy addresses — the Compound/TUSD class) that defeat per-address balance/Transfer accounting; and return-false-not-revert tokens. A value-diff built on Transfer logs + balance diff inherits all of these. Scope which token classes are in/out, and degrade-to-warning on detected hooks (same discipline as the native-trace gap).

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+- **Slippage & fees within tolerance**, and **PublicAllocator fee is correct** (vault/market reallocation).
+- **No asset leaves to an unknown / unexpected address.** Every outflow recipient is either the user or an explicitly-allowed destination.
+- **Legitimate non-`from` recipients are allowed.** A receiver different from the sender is expected and fine for: vault `deposit`/`mint` → shares to `receiver`, bridges, smart-account / 4337 flows, and callback adapters. These must pass, not trip the "unknown recipient" check.
```
0xbulma, line 73: **[Low] Allow-list the resolved callback *target*, not the callback *role*.**

Treating "callback adapters" as a permitted non-`from` recipient role is a hole if the target is attacker-chosen. The resolved callback target (flashloan provider, Paraswap augustus) routes value and must itself be trust-listed — tie the allowance to the resolved target address, not the role.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Alternative 3: A new generic on-chain "tracer" backend
+
+Build our own EVM tracer instead of consuming Tenderly / `eth_simulateV1` state-diff output.
+
+**Why rejected:** enormous scope, duplicates what the backends already return, and breaks the package's I/O-at-the-edge rule. We already normalize backend output to `RawSimulationResult` — state diffs and full logs extend that, no new engine required.
+
+## Assumptions & Constraints
+
+- Both backends expose **state/storage diffs and full event logs** for a simulated bundle (Tenderly natively; `eth_simulateV1` via `stateDiff` + per-call logs). Where one backend is thinner (e.g. `eth_simulateV1` internal native transfers), the check **degrades to a typed warning**, never a silent pass — same discipline as the retention skip on unknown chains.
```
0xbulma: **[Medium] Native-ETH no-loss degrades to a warning on `eth_simulateV1` — for the highest-value asset.**

Internal native transfers (call-value, selfdestruct refunds) are invisible to `eth_simulateV1` and a prime exfil path through callback adapters. "Degrade to warning, never silent pass" is the right rule, but consider *requiring Tenderly* for native-bearing bundles rather than letting a user click through a warning on the most fungible, highest-value asset.

0xbulma (commented):
## Web3 threat-model review (scoped to intent gaps)

Reviewed as a design doc, focused on web3/EVM gaps in the threat model — grounded in the bundler3 contract (transient `initiator`, adapter fund-pull, the `reenter()`/`callbackHash` callback commitment, "the dispatcher holds no approvals, adapters do") and the current `evm-simulation` code. Inline comments mark specific gaps. Two cross-cutting points below.

**[Low] Name sim-vs-execution divergence as a Constraint/Non-Goal.** Every guarantee holds only at the simulated block; slippage tolerance is the sole guard against a bundle that's safe in simulation but sandwiched at inclusion (MEV / state drift). Not a flaw, but it belongs in Non-Goals so "no value loss" isn't read as an execution-time guarantee.

**Two strongest additions:** (a) make "decode the bundler3 `reenter`/`callbackHash` callback sub-bundle" an explicit Layer-A requirement (see comment on the recursive-batch-decode bullet); (b) promote "residual-allowance baseline policy" and "EIP-7702 authorization interception" from implicit to first-class Goals / Open Questions.

**Reusability:** most Layer-A primitives already exist — Permit2 + EIP-2612 (incl. DAI variant) typed-data in `blue-sdk-viem/src/signatures/{permit2,permit}.ts`, address registry via `getChainAddresses` (`blue-sdk/src/addresses.ts`, exposes `generalAdapter1`/`permit2`/`bundler3` per chain), ERC-4626 math in `blue-sdk/src/vault/VaultUtils.ts`, Morpho authorization typed-data in `blue-sdk-viem/src/signatures/manager.ts`. Only EIP-7702 and the bundler3 callback-bundle decoder are genuinely net-new.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+
+- **More batch formats** as they appear (new bundler versions, alternative account-abstraction entry points).
+- **Cross-chain / bridge intent** verification (destination-chain expectations) once bridge flows enter the SDK.
+- **USD/price-aware value diff** as an optional layer for integrators who want a single net-value number rather than per-asset deltas.
+- **Vault V2 adapter-specific checks** (per-adapter allow-lists) as the V2 adapter set grows.
+
+## Open Questions
+
+- Should trust-list checks be **hard errors or warnings by default**? Leaning: trust-list miss = error; backend-coverage gap (e.g. missing internal native trace) = warning.
+- For ERC-4626 share valuation, do we compare **shares** directly, or convert to assets via `convertToAssets` at the simulated post-state? (Affects how slippage is expressed for vault flows.)
```
0xbulma: **[Low] Open Question 2 has a security edge.** `convertToAssets` at the *simulated post-state* can be skewed by an intra-bundle donation/inflation move, so the vault value-diff should be robust to in-bundle share-price manipulation, not just read the post-state conversion (the root invariant set already calls out an inflation-attack guard). Reuse `VaultUtils.toAssets`/`toShares` from `blue-sdk/src/vault/VaultUtils.ts`.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,147 @@
...
+## Future Considerations
+
+- **More batch formats** as they appear (new bundler versions, alternative account-abstraction entry points).
+- **Cross-chain / bridge intent** verification (destination-chain expectations) once bridge flows enter the SDK.
+- **USD/price-aware value diff** as an optional layer for integrators who want a single net-value number rather than per-asset deltas.
+- **Vault V2 adapter-specific checks** (per-adapter allow-lists) as the V2 adapter set grows.
+
+## Open Questions
+
+- Should trust-list checks be **hard errors or warnings by default**? Leaning: trust-list miss = error; backend-coverage gap (e.g. missing internal native trace) = warning.
```
Rubilmax: agreed

Rubilmax (commented):
Great!

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
Foulks-Plb: Good point. But the model will be based on an allowlist rather than a denylist. The only contracts that will be authorized are the General Adapter and Permit2.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
Foulks-Plb: Good point, I’ll clarify.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
Foulks-Plb: True, I’ll clarify the Permit2 case.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
Foulks-Plb: This operates at the wallet/account level rather than the application level. I’m not sure this is something we can reliably verify at the app layer.

Foulks-Plb: Thinking about it more, I don’t think amount and expiration are part of our security model. GeneralAdapter is a static and trusted component, so the critical property is who receives the allowance, not its amount or duration.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
Foulks-Plb: Good point. Net diffs across asset types are too complex to handle properly at the moment. Let’s avoid introducing this notion for now and focus on verifying conformance to the declared expectations.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
Foulks-Plb: Hmm, maybe that’s out of scope for now. I’ve never seen any tokens other than standard ERC-20s on Morpho.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
Foulks-Plb: I just found a solution that eliminates the blind spot around native ETH utilization. We’ll be able to remove this limitation from the TIB.

Foulks-Plb: https://github.com/morpho-org/sdks/pull/803

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
Foulks-Plb: Maybe we shouldn’t implement inflation attack protection, it seems quite difficult to do properly. Unless someone has a quick and simple idea for how to implement it?

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,197 @@
...
+| **Status** | Proposed                                                        |
+| **Date**   | 2026-06-15                                                      |
+| **Author** | @foulques                                                       |
+| **Scope**  | Package: `evm-simulation`; consumers: Vaults frontend and write API |
+
+---
+
+## Context
+
+Transaction creation is moving from the `morpho-org/morpho-apps` frontend to an independent write
```
Foulks-Plb, line 14: Perfect

Foulks-Plb:
I think it would be helpful to make the “why” behind these protections more explicit — i.e. which threats or failure modes we are trying to protect against and why they matter.

This would make the intent behind the checks clearer and help our agents iterate on the design without losing the original security rationale.

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,229 @@
...
+balance/permission/position/market records, conversions and fees. Include **before/after, diff and
+actionDiff** (excluding accrual), including unchanged values; numeric deltas only. Failures throw typed errors.
+
+### Consumer limits
+
+| Optional `limits` field | Meaning / default |
+| --- | --- |
+| `maxSlippageBps` | Conversion slippage; default 3 bps |
+| `minLltvBufferBps` | Distance below liquidation threshold; default 50 bps |
+| `ranges` | Inclusive `{ subject?, metric, at?, min?, max? }`; `at` = `after` (default), `diff` or `actionDiff`. Subject = account + token/market/vault; omit when unambiguous |
```
jinmel: I think this is too broad definition. lets try to be more specific here to keep from agents generating helpers and error types for every different cases that `ranges` can have

docs/tibs/TIB-2026-06-15-evm-simulation-intent-effect-verification.md
```diff
@@ -0,0 +1,229 @@
...
+Intersect SDK defaults, calldata and caller bounds; callers only tighten checks. Reject malformed,
+unknown, ambiguous, inapplicable or contradictory bounds. Deterministic checks need no caller limits.
+
+### Position and market checks
+
+Cover affected positions, backing allocations, fee recipients and markets. Compare against a no-action
+baseline accrued to simulation time. Reconcile calls in order with protocol rounding/clamps and
+entrypoint-specific accrual/fee mints. Raw shares are exact; claim diffs need not equal cash flows.
+
+| State / metric | Required checks |
```
jinmel: add one more column for `error type` when the validation fails - or dedicate a new table

0xbulma (dismissed):

docs/tibs/TIB-2026-09-18-evm-simulation-state-override-verification.md
```diff
@@ -0,0 +1,482 @@
...
+| `blueAuthorizationSignature` | Morpho | `isAuthorized[authorizer][authorized]` | `message.isAuthorized` | As above, plus `Morpho.nonce(authorizer)` equals `message.nonce`, `deadline` within the lifetime bound, `domain.verifyingContract` is the registered Morpho | As above |
+
+- **Permit2 storage is not overridden.** The no-signature build never routes through Permit2, so the
+  Permit2 tuple, nonce advance and expiration are verified on the request in `preview` and on real
+  execution in `final`. The token-allowance override is the same authority
+  `(owner, token, spender, amount)` expressed where the `preview` form reads it.
+- **Storage keys are derived, then proven.** Morpho's storage layout is pinned in-package like its ABI.
+  Token allowance keys are discovered per token at verification time from candidate layouts and
+  confirmed by the read-back; a key that cannot be uniquely identified fails with
+  `UnsupportedVerificationFeatureError`. Every read-back runs against the same block and override set
```
Foulks-Plb, line 227: Nice. Could you make sure that all the tokens currently available in the app are supported by our simulation and this strategy?

Foulks-Plb: Like DAI, USDC proxy, Storage namespace ERC-7201,...

Foulks-Plb:
I’m worried the state override strategy might be too time-consuming, since we’d first need to fetch the slot and then run the simulation.

We should expose an option to consumers to either disable this feature or override the slot themselves.

docs/tibs/TIB-2026-09-18-evm-simulation-state-override-verification.md
jinmel: sure thing

jinmel:
@Foulks-Plb I reverted to prepending ERC-20 `approve()` eth_call before simulation because:
- the storage layout can be arbitrary 
- additional rpc call harms user experience

PR #803, opened 2026-06-17
feat(evm-simulation): capture internal native ETH via traceTransfers on eth_simulateV1

Foulks-Plb:
## What

Enable `traceTransfers` on the `eth_simulateV1` simulation backend so native-ETH moves are captured in `assetChanges` — including ETH moved through **internal calls** (e.g. a `WETH.withdraw` refund, a swap that pays out ETH), which previously emitted no log and was invisible on this path.

## Why

The `eth_simulateV1` backend derived native ETH only from each transaction's top-level `value`. ETH moved internally was silently dropped, so `assetChanges` was less complete than the Tenderly backend (the old JSDoc spelled this out as a known caveat: *"For full native-ETH accounting use Tenderly"*). This closes that gap — both backends now report the full net native-ETH delta.

## How

- **`eth-simulate-v1.ts`** — pass `traceTransfers: true` to `simulateCalls`. The node then synthesizes every native-ETH move (top-level *and* internal) as a `Transfer` event emitted from the native sentinel. `assetChanges` is now derived **entirely** from parsed transfer logs; the top-level `value` accounting is removed because `traceTransfers` already logs it and keeping both would double-count.
- **`parsing/transfers.ts`** — new `normalizeTransferToken` helper collapses the native sentinel (`0xeee…eee`, lowercase) to viem's `ethAddress` constant rather than its checksummed form, so native deltas key consistently across backends (`groupAssetChanges` keys tokens by exact string). Real ERC20 tokens stay checksummed. Tenderly emits no such logs, so the path is inert there.

## Validation

Verified empirically against a local **anvil 1.7.1** devnet that `eth_simulateV1` + `traceTransfers` emits native moves as `Transfer` logs at the `0xeee…` sentinel (= viem's `ethAddress`), confirming the address-normalization choice. Anvil is what the fork suite runs against.

- Unit tests: `traceTransfers` flag is set; native ETH accounted from synthetic logs; no double-count when a tx has `value` but no native log; parser normalizes the sentinel to `ethAddress`.
- Fork test (mainnet, opt-in): `WETH.withdraw` nets sender −1 WETH / +1 ETH — the internal refund the old path missed.
- `tsc --noEmit`, Biome, and the full `evm-simulation` unit suite (226 tests) pass.

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/803" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


Foulks-Plb:
@codex review

Rubilmax (dismissed):

Rubilmax (approved):

PR #815, opened 2026-06-23
feat(morpho-sdk): add Midnight action flows

Rubilmax:
## Summary

Add the concrete `morpho-sdk` Midnight action-flow implementation on top of the prerequisite SDK split in #829.

This PR now contains only:
- `client.morpho.midnight(chainId)` entity wiring
- `MorphoMidnight` action outputs for Midnight flows
- pure `actions/midnight` transaction builders and focused unit tests
- the implementation changeset for the Midnight action flows

The shared SDK prerequisites, Blue requirement-helper refactors, Midnight SDK package changes, and `midnight-api` re-export path are isolated in #829. The TIB remains isolated in #769.

## Stack

1. #769 - Midnight action-flow TIB
2. #829 - SDK prerequisites for the action flows
3. #815 - Concrete Midnight action-flow implementation

## Validation

- `pnpm --filter @morpho-org/morpho-sdk build`
- `pnpm --filter @morpho-org/morpho-sdk test -- packages/morpho-sdk/src/actions/midnight packages/morpho-sdk/src/entities/midnight` was attempted, but the morpho-sdk Vitest project still loads fork-backed suites that require `MAINNET_RPC_URL`; 364 tests passed before env-gated suites failed during setup.


packages/morpho-sdk/src/actions/midnight/borrowMarket.ts
```diff
@@ -0,0 +1,92 @@
...
+  MidnightBorrowMarketParams,
+  MidnightCollateralSupply,
+} from "./types.js";
+
+/** Encodes the borrow-market Midnight bundle. */
+export const midnightBorrowMarket = (
+  params: MidnightBorrowMarketParams,
+): Readonly<Transaction<MidnightBorrowMarketAction>> => {
+  if (params.collateralAssets < 0n) {
+    throw new NegativeMidnightAmountError(
```
Rubilmax: Blue actions use blue-specific errors with blue-specific `Market` for example. Why? tbd with @Foulks-Plb

packages/morpho-sdk/src/actions/midnight/takeBorrow.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+  if (params.maxUnits < 0n) {
+    throw new NegativeMidnightAmountError("maxUnits", params.maxUnits);
+  }
+  if (params.takes.length === 0) {
+    throw new EmptyMidnightTakesError();
+  }
+
+  const marketId = MarketUtils.toId({
+    market: params.market,
```
Rubilmax: why provide a market as params? it's already included in takes

packages/morpho-sdk/src/constants.ts
```diff
@@ -8,6 +8,24 @@ export {
   SECONDS_PER_YEAR,
   TransactionType,
 } from "@morpho-org/blue-sdk";
+export {
+  ALLOWED_LLTVS,
```
Rubilmax: needs rebase since it was removed

Rubilmax (commented):
partial review

packages/morpho-sdk/src/actions/midnight/takeBorrow.ts
Rubilmax: I would keep `market` explicit here. It matches the Blue action boundary where pure action encoders receive the target market params instead of deriving them from nested execution items.

For Midnight it is not only duplicate metadata: `borrowMarket` uses the explicit market to derive the intended market id, validate that every take belongs to that market, and resolve the collateral token for the collateral `TokenPermit`. If we derived from `takes[0].offer.market`, the first quote take would become the source of truth and the action could no longer distinguish “wrong take for the requested market” from “requested market inferred from the wrong take”.

The entity already fetched the market from `marketId`, so passing it down is cheap and keeps the pure action encoder deterministic. I would not change this unless we add a separate convenience wrapper that derives the market from takes and still performs the same consistency validation.

Rubilmax:
@codex

packages/morpho-sdk/src/entities/midnight/midnight.ts
```diff
@@ -0,0 +1,1017 @@
...
+    const offers = params.offers.map((offer) =>
+      Offer.create({
+        market: offer.market,
+        buy: params.buy,
+        maker: params.accountAddress,
+        start: offer.start,
+        expiry: offer.expiry,
+        tick: offer.tick,
+        ratifier: ratifier.ratifier,
+        maxAssets: params.loanAssets,
```
yum0e: This maker flow is very narrow for router usage. Every offer gets `maxAssets: params.loanAssets`, and below this all offers go into one shared `Group.create(offers)`. That prevents router-style offers with `maxUnits`, `reduceOnly`, callbacks, custom receiver, fee cap, or multiple groups/standalone offers.

Could this accept fuller `Offer.create` inputs, or an explicit group/tree shape, instead of forcing this one layout?

packages/morpho-sdk/src/entities/midnight/midnight.ts
```diff
@@ -0,0 +1,1017 @@
...
+      throw new UnknownMidnightCollateralError({
+        market: "provided market",
+        collateralIndex,
+      });
+    }
+
+    if (params.offers.length === 0)
+      throw new EmptyMidnightMakeOfferInputsError();
+
+    const prepared = await this.prepareOffers({
```
yum0e: I think this should validate that `params.market` matches every offer market. Right now collateral is supplied to `params.market`, but `prepareOffers` only checks that offers share the same loan token. A user could supply collateral to market A and publish borrow offers for market B.

The take flows already do market-id checks for takeable offers; this path should do the same kind of check.

packages/morpho-sdk/src/actions/midnight/takeBorrow.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (actualMarketId.toLowerCase() !== marketId.toLowerCase()) {
+      throw new MidnightTakeableOfferMarketMismatchError({
+        index,
+        expectedMarket: marketId,
+        actualMarket: actualMarketId,
+      });
+    }
+  }
+
+  const midnightBundles = getChainAddress(params.chainId, "midnightBundles");
```
yum0e, line 94: Router migration blocker: this resolves deployments from the global chain registry. Router previews/local tests use runtime deployment addresses, so `chainId` alone is not enough.

Same issue in other Midnight builders, for example:
- [authorization.ts](https://github.com/morpho-org/sdks/blob/0b6d7a2e3712d9939ae0f150bebf5603de275760/packages/morpho-sdk/src/actions/midnight/authorization.ts#L25)
- [ratifyRoot.ts](https://github.com/morpho-org/sdks/blob/0b6d7a2e3712d9939ae0f150bebf5603de275760/packages/morpho-sdk/src/actions/midnight/ratifyRoot.ts#L129)
- [submitOffers.ts](https://github.com/morpho-org/sdks/blob/0b6d7a2e3712d9939ae0f150bebf5603de275760/packages/morpho-sdk/src/actions/midnight/submitOffers.ts#L24)
- [supplyCollateral.ts](https://github.com/morpho-org/sdks/blob/0b6d7a2e3712d9939ae0f150bebf5603de275760/packages/morpho-sdk/src/actions/midnight/supplyCollateral.ts#L38)

Could these params accept address overrides and only fall back to the registry when omitted?

packages/morpho-sdk/src/actions/midnight/repayWithdrawCollateral.ts
```diff
@@ -0,0 +1,111 @@
...
+    chainId: params.chainId,
+  });
+  const market = MarketUtils.toStruct(params.market);
+  const midnightBundles = getChainAddress(params.chainId, "midnightBundles");
+  const receiver = params.receiver ?? params.onBehalf;
+  const collateralWithdrawals: readonly MidnightCollateralWithdrawal[] =
+    params.withdrawCollateralAssets > 0n
+      ? [
+          {
+            collateralIndex: params.collateralIndex ?? 0n,
```
yum0e, line 88: If `withdrawCollateralAssets > 0`, can we validate that `collateralIndex` exists on the market before building the tx? Other flows resolve the collateral index earlier, but this path can build a transaction that only fails onchain for an unknown index.

packages/morpho-sdk/src/actions/midnight/mempoolSubmitOffers.ts
```diff
@@ -0,0 +1,42 @@
...
+  readonly payload: Hex;
+}
+
+/** Encodes the Midnight mempool payload submission transaction. */
+export const midnightSubmitOffers = (
+  params: MidnightSubmitOffersParams,
+): Readonly<Transaction<MidnightSubmitOffersAction>> => {
+  const midnightMempool = getChainAddress(params.chainId, "midnightMempool");
+
+  return deepFreeze({
```
yum0e: Unlike the other action builders, this does not apply `metadata` with `addTransactionMetadata`. Is that intentional? If not, offer submissions lose the same attribution/metadata added to the other Midnight transactions.

packages/morpho-sdk/src/actions/midnight/ratifyRoot.ts
```diff
@@ -0,0 +1,52 @@
...
+export interface MidnightRatifyRootParams {
+  readonly chainId: number;
+  readonly maker: Address;
+  readonly root: Hex;
+  readonly isRootRatified?: boolean;
+  readonly metadata?: Metadata;
+}
+
+/** Encodes `SetterRatifier.setIsRootRatified(maker, root, true)`. */
+export const midnightRatifyRoot = (
```
bilbo524: nit: Unlick other actions, this is not on midnight contract so in order to create a clear mental model for users I would change a bit the naming convention. Maybe something like `setterRatifierRatifyRoot`

packages/morpho-sdk/src/actions/midnight/submitOffers.ts
```diff
@@ -0,0 +1,42 @@
...
+  readonly root: Hex;
+  readonly maker: Address;
+  readonly ratifier: Address;
+  readonly ratifierType: "ecrecover" | "setter";
+  readonly offers: number;
+  readonly payload: Hex;
+}
+
+/** Encodes the Midnight mempool payload submission transaction. */
+export const midnightSubmitOffers = (
```
bilbo524: nit: Same as `ratifyRoot`, here I would push for something like `mempoolSubmitOffers`

bilbo524 (commented):
It seems that the ecrecover CancelRoot action is missing

packages/morpho-sdk/src/actions/midnight/supplyCollateralTakeBorrow.ts
```diff
@@ -0,0 +1,129 @@
...
+        expectedMarket: marketId,
+        actualMarket: actualMarketId,
+      });
+    }
+  }
+
+  const midnightBundles = getChainAddress(params.chainId, "midnightBundles");
+  const market = MarketUtils.toStruct(params.market);
+  const collateralIndex = params.collateralIndex ?? 0n;
+  const collateral = market.collateralParams[Number(collateralIndex)];
```
Foulks-Plb: Why you dont use market.getCollateralParamsByIndex(collateralIndex) ?

packages/morpho-sdk/src/client/morphoViemExtension.ts
```diff
@@ -51,6 +56,10 @@ function createMorphoNamespace(
       }
       return new MorphoBlue(namespace, marketParams, chainId);
     },
+
+    midnight(chainId: number) {
+      return new MorphoMidnight(namespace, chainId);
```
Foulks-Plb, line 61: 🔥

packages/morpho-sdk/src/entities/midnight/midnight.ts
```diff
@@ -0,0 +1,965 @@
...
+          : {};
+
+    const [position, block] = await Promise.all([
+      fetchAccrualPosition(this.client.viemClient, {
+        ...params.parameters,
+        deployless: this.client.options.supportDeployless,
+        marketId: params.marketId,
+        user: params.accountAddress,
+      }),
+      getBlock(this.client.viemClient, blockParameters),
```
Foulks-Plb: I’m not sure how the fixed-rate app does it, but for Blue we fetch the block beforehand and pass it as a parameter to getPositionData.

packages/morpho-sdk/src/entities/midnight/midnight.ts
```diff
@@ -0,0 +1,965 @@
...
+    ]);
+
+    return position.accrueInterest(block.timestamp);
+  }
+
+  async getOffersData(params: GetOffersDataParams): Promise<OffersData> {
+    validateChainId(this.client.viemClient.chain?.id, this.chainId);
+    const tree = Tree.create(params.tree);
+    const ratifier = tree.offers[0]!.ratifier;
```
Foulks-Plb: Do we need a JS assertion to avoid this non-null assertion ?

packages/morpho-sdk/src/actions/midnight/encodeMidnightTokenPermit.ts
```diff
@@ -0,0 +1,125 @@
...
+ * const permit = encodeMidnightTokenPermit({
+ *   token: loanToken,
+ *   owner: taker,
+ *   spender: midnightBundles,
+ *   amount: 1_000_000n,
+ *   signatures,
+ * });
+ * ```
+ */
+export const encodeMidnightTokenPermit = (
```
Foulks-Plb: can we move it in the requirements folder ?

packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Permit.ts
```diff
@@ -68,8 +79,24 @@ export const encodeErc20Permit = async (
   }
 
   const {
+    midnightBundles,
     bundler3: { generalAdapter1 },
   } = getChainAddresses(chainId);
+  const spender = params.spender ?? generalAdapter1;
```
Rubilmax: could be required param but it's a breaking change

packages/morpho-sdk/src/actions/requirements/midnight/getMidnightBundlesRequirements.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (chainAddresses.permit2 != null) {
+      const permit2Allowance = await readContract(params.viemClient, {
+        address: params.token,
+        abi: erc20Abi,
+        functionName: "allowance",
+        args: [params.owner, chainAddresses.permit2],
+      });
+      const nonceBytes = new Uint8Array(32);
+      if (globalThis.crypto?.getRandomValues == null) {
+        throw new CryptoUnavailableError("Permit2 unordered nonce generation");
```
Rubilmax: need crypto for random nonce?

Rubilmax (commented):
Almost ready

packages/morpho-sdk/src/actions/midnight/repayWithdrawCollateral.ts
```diff
@@ -0,0 +1,159 @@
...
+      params.withdrawCollateralAssets,
+    );
+  }
+  if ((params.referralFeePct ?? 0n) < 0n) {
+    throw new NegativeMidnightAmountError(
+      "referralFeePct",
+      params.referralFeePct ?? 0n,
+    );
+  }
+  if ((params.deadline ?? maxUint256) < 0n) {
```
Rubilmax: possibly undefined deadline??

packages/morpho-sdk/src/actions/midnight/repayWithdrawCollateral.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+  const collateralWithdrawals =
+    params.collateralWithdrawals ??
+    (params.withdrawCollateralAssets > 0n
+      ? [
+          {
+            collateralIndex: params.collateralIndex ?? 0n,
+            assets: params.withdrawCollateralAssets,
+          },
+        ]
```
Rubilmax: 2 kinds of input interface? Seems very complex

packages/morpho-sdk/src/actions/midnight/supplyCollateralTakeBorrow.ts
```diff
@@ -0,0 +1,153 @@
...
+  if (params.maxUnits < 0n) {
+    throw new NegativeMidnightAmountError("maxUnits", params.maxUnits);
+  }
+  if ((params.referralFeePct ?? 0n) < 0n) {
+    throw new NegativeMidnightAmountError(
+      "referralFeePct",
+      params.referralFeePct ?? 0n,
+    );
+  }
+  if ((params.maxContinuousFee ?? maxUint256) < 0n) {
```
Rubilmax: why possibly undefined max continuous fee?

packages/morpho-sdk/src/actions/midnight/supplyCollateralTakeBorrow.ts
```diff
@@ -0,0 +1,153 @@
...
+        signatures: params.signatures,
+      }),
+    },
+  ];
+  const reduceOnly = params.reduceOnly ?? false;
+  const receiver = params.receiver ?? params.taker;
+  const referralFeePct = params.referralFeePct ?? 0n;
+  const referralFeeRecipient = params.referralFeeRecipient ?? zeroAddress;
+  const maxContinuousFee = params.maxContinuousFee ?? maxUint256;
+  const deadline = params.deadline ?? maxUint256;
```
Rubilmax: define prior to checking them so the default fallback is only declared once

Rubilmax (commented):
partial review

packages/morpho-sdk/src/entities/midnight/midnight.ts
```diff
@@ -0,0 +1,1058 @@
...
+  if (amount < 0n) throw new NegativeMidnightAmountError(label, amount);
+};
+
+const assertPositiveAmount = (label: string, amount: bigint) => {
+  if (amount <= 0n) throw new NonPositiveMidnightAmountError(label, amount);
+};
+
+const validateMarketData = (market: Market, chainId: number) => {
+  validateChainId(Number(market.chainId), chainId);
+};
```
Rubilmax, line 147: seems duplicated multiple times over the source code

packages/morpho-sdk/src/entities/midnight/midnight.ts
```diff
@@ -0,0 +1,1058 @@
...
+
+/** Parameters for building and validating Midnight offer data. */
+export interface GetOffersDataParams {
+  readonly accountAddress: Address;
+  readonly offers: TreeInput;
+  readonly validation?: OfferValidationParams;
+}
+
+/** Prepared Midnight maker-offer data derived from a tree-like offer set. */
+export interface OffersData {
```
Rubilmax: a lot of types are declared in this file. Is it similar to blue? feels like would deserve a file on their own

packages/morpho-sdk/src/actions/midnight/takeBorrow.ts
Rubilmax: Addresses can be hydrated at runtime, so it should not be a blocker

packages/morpho-sdk/src/actions/midnight/types.ts
```diff
@@ -0,0 +1,19 @@
+import type { OfferStruct } from "@morpho-org/midnight-sdk";
+import type { Hex } from "viem";
+
+/** Protocol-shaped collateral supply used by Midnight bundle calls. */
+export interface MidnightCollateralSupply {
+  readonly collateralIndex: bigint;
+  readonly assets: bigint;
+  readonly permit: {
+    readonly kind: 0;
```
Rubilmax: Use PermitKind for this

packages/morpho-sdk/src/actions/midnight/authorization.ts
```diff
@@ -0,0 +1,76 @@
...
+ * import { midnightAuthorization } from "@morpho-org/morpho-sdk";
+ *
+ * const tx = midnightAuthorization({
+ *   chainId: 8453,
+ *   authorized: midnightBundles,
+ *   onBehalf: user,
+ * });
+ * ```
+ */
+export const midnightAuthorization = (
```
Rubilmax:
```suggestion
export const midnightSetIsAuthorized = (
```

?

Rubilmax (commented):
LGTM

Rubilmax:
@codex

packages/morpho-sdk/src/entities/midnight/midnight.ts
```diff
@@ -0,0 +1,928 @@
...
+    assertNonNegativeAmount("minUnits", params.minUnits);
+    assertNonNegativeAmount("deadline", params.deadline);
+
+    const market = params.marketData;
+    const midnightBundles = getChainAddress(this.chainId, "midnightBundles");
+
+    return {
+      getRequirements: async () => {
+        const requirements: ActionRequirement[] = [
+          ...(await getMidnightApprovalRequirements({
```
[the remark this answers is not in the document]
Rubilmax: if the quote is stale, we cant do nothing

Rubilmax:
@codex

packages/morpho-sdk/src/actions/midnight/supplyCollateralTakeBorrow.ts
```diff
@@ -0,0 +1,158 @@
...
+        params.loanAssets,
+        params.maxUnits,
+        params.taker,
+        false,
+        params.taker,
+        collateralSupplies,
+        params.takeableOffers,
+        0n,
+        zeroAddress,
+        maxUint256,
```
Foulks-Plb, line 120: is it intentional about maxContinuousFee ?

Rubilmax: Yes, intentional for the initial Markets app migration. `maxContinuousFee` is a taker-side cap, separate from the offers’ `continuousFeeCap`. The current app passes no bounded cap, so the TIB specifies `maxUint256` and defers expanding the public action parameters. If we want this SDK surface to provide fee-slippage protection now, it should become an explicit required parameter across every take flow, like `deadline`.

Rubilmax: It will come with the PR later, #857

packages/morpho-sdk/src/entities/midnight/midnight.ts
```diff
@@ -0,0 +1,967 @@
...
+  ): Promise<AccrualPosition> {
+    validateChainId(this.client.viemClient.chain?.id, this.chainId);
+    const parameters = params.parameters ?? {};
+    const blockParameters =
+      parameters.blockNumber != null
+        ? { blockNumber: parameters.blockNumber }
+        : parameters.blockTag != null
+          ? { blockTag: parameters.blockTag }
+          : {};
+    const block = await getBlock(this.client.viemClient, blockParameters);
```
Foulks-Plb: The block should be fetched outside the action so that all fetches use the same block.

packages/morpho-sdk/test/helpers/midnight.ts
```diff
@@ -0,0 +1,24 @@
+import { ChainId } from "@morpho-org/blue-sdk";
+import { registerCustomAddresses } from "@morpho-org/morpho-ts";
+import type { Address } from "viem";
+
+export const midnightForkAddresses = {
```
Foulks-Plb: Why not use the ones from morpho-ts?

packages/morpho-sdk/src/entities/midnight/midnight.ts
Rubilmax: In blue, we pass the blockNumber as input param and nothing guarantees multiple instances are fetched at the same block if "latest" is provided (or no input). So i applied the same here in midnight

Foulks-Plb: Oh in that case, it’s a mistake in Blue. Do you know exactly which fetch in Blue?

Foulks-Plb (approved):

packages/morpho-sdk/src/entities/midnight/midnight.ts
Rubilmax: `getReallocationData` and `getPositionData` because it is part of how the fetchers work: they dont query the latest block beforehand if "latest" is provided. IMO this requires a deeper refactoring and since it did not break anything so far im not sure it's worth it

Foulks-Plb: Sorry, I’m not sure I understand here. My previous point was about removing the getBlock call from the getPositionData function in the Midnight entity.

The block is fetched on the consumer side so all data is aligned to the same block.

I don’t see similar logic in this function (getReallocationData & getPositionData). Do you have a URL? It looks like it already receives the block as a parameter, which is the correct behavior.

Foulks-Plb:
@codex review

.changeset/fixed-rate-offer-chain.md
```diff
@@ -0,0 +1,15 @@
+---
+"@morpho-org/midnight-sdk": minor
```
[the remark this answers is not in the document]
Rubilmax: Because the package is young enough, this feels unnecessary to me. The only changes are to conform to blue's interface where user and marketId are passed and it feels ok to do this change this early

Foulks-Plb (approved):

Foulks-Plb (dismissed):

Foulks-Plb (approved):

PR #823, opened 2026-06-26
fix(blue-sdk-viem): repair deployless GetVault EIP-5267 decode

Foulks-Plb:
## Motivation

The deployless `GetVault` query reverted on **every** MetaMorpho vault (v1.0 and v1.1) while decoding the EIP-5267 domain: reading the high-level `eip712Domain()` struct return hits a Solidity via-IR decoding regression that reverts on valid domains. Because `fetchVault`/`fetchAccrualVault` default to `deployless: true`, they silently fell back to multicall (so tests stayed green and users still got results), while the single-RPC-call deployless fast path was dead and `deployless: "force"` threw. Verified the revert reproduces on direct public RPCs and on a high-gas anvil fork (so it was never an RPC gas-cap issue).

## Solution

`GetVault.sol` now decodes the raw `eip712Domain()` returndata as a tuple — the same workaround already applied to `GetToken` — and additionally reads `lostAssets` (MetaMorpho V1.1) so the deployless and multicall paths return byte-identical `Vault` state. Regenerated `GetVault.ts`, wired `lostAssets` through the deployless branch of `fetchVault`, and added regression tests asserting `deployless: "force"` succeeds and equals the multicall path for both vault versions. Includes a patch changeset for `blue-sdk-viem` and its direct maintained dependents (`morpho-sdk`, `wdk-protocol-lending-morpho-evm`).


Rubilmax (approved):
Very good find, can we extend the tests added to all kinds of deployless calls?

PR #833, opened 2026-06-29
feat(morpho-sdk): integrate setAuthorizationWithSig with array buildTx

Foulks-Plb:
## Motivation

Bundled Blue paths that operate on behalf of the user (`borrow`, `withdraw`, `supplyCollateralBorrow`, `repayWithdrawCollateral`, `refinance`) previously required a standalone `setAuthorization` transaction before the bundle, forcing an extra onchain step even when the integrator can collect offchain signatures.

## Solution

When `supportSignature` is enabled, the authorization requirement is now returned as a signable `Requirement`; signing it yields an `AuthorizationRequirementSignature` that `buildTx` folds into the bundle as a `setAuthorizationWithSig` call, removing the separate transaction. Entity `buildTx` now accepts an array of signatures (`buildTx(signatures?: readonly RequirementSignature[])`) — split internally via the new `isPermitSignature` / `isAuthorizationSignature` guards — so permit and authorization signatures travel together. New public surface includes `encodeAuthorization`, `getAuthorizationAction`, the `morphoSetAuthorizationWithSig` bundler action, the `AuthorizationRequirementSignature` / `PermitRequirementSignature` types, and `BundlerErrors.UnexpectedSignature`; the `wdk-protocol-lending-morpho-evm` dependent is updated accordingly. This is a breaking `buildTx` signature change (single → array) so the changeset marks `@morpho-org/morpho-sdk` as **major**.


.changeset/morpho-sdk-set-authorization-with-sig.md
```diff
@@ -0,0 +1,14 @@
+---
+"@morpho-org/morpho-sdk": major
+"@morpho-org/wdk-protocol-lending-morpho-evm": minor
+---
+
+Add offchain `setAuthorizationWithSig` support for Morpho Blue bundled paths and switch entity `buildTx` to accept an array of signatures.
```
Foulks-Plb, line 7: **[Medium] Peer-dependent audit triggered by the `major` bump — `liquidity-sdk-viem` is missing.**

`@morpho-org/morpho-sdk` goes `major` here (4.2.0 → 5.0.0). Per AGENTS.md §4/§7, a package bump obligates auditing every package that declares it as a peer dependency. `@morpho-org/liquidity-sdk-viem` (maintained, not frozen) declares `"@morpho-org/morpho-sdk": "^3.0.0"` in `peerDependencies` and consumes it at runtime (`PublicReallocation`, `ReallocationData` in `src/loader.ts`).

That range `^3.0.0` is in fact **already stale** against the current 4.x — but this major bump to 5.0.0 is exactly the §7 trigger point where the author must decide: widen the peer range and add `liquidity-sdk-viem` to this changeset (at least a patch), or explicitly document why it's excluded. Right now it's silently absent.

packages/morpho-sdk/src/actions/blue/authorizationSignature.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        shares: 0n,
+        receiver: USER,
+        minSharePrice: 0n,
+        authorizationSignature,
+      },
+    });
+
+    const calls = decodeBundle(tx.data);
+    const inner = decodeFunctionData({ abi: blueAbi, data: calls[0]!.data });
+    expect(calls[0]!.to).toBe(morpho);
```
Foulks-Plb, line 95: **[High] No fork-based integration test for the signed-authorization bundle path.**

This file (and the other new unit tests) only use a *fake static* signature and decode calldata — they never submit a `setAuthorizationWithSig` bundle against real state. The existing fork tests for the affected paths (`test/actions/blue/{borrow,withdraw,refinance,supplyCollateralBorrow,repayWithdrawCollateral}.test.ts`) all still submit authorization as a *standalone* `sendTransaction(authorization)` and then call `buildTx()` with no signature.

Per AGENTS.md §5 + §2 rule 6, `setAuthorizationWithSig` correctness depends on real onchain state — the signed `nonce` must match Morpho's current `nonce(owner)`, and `isAuthorized` gating is live — so it must be exercised via an Anvil fork at a pinned block, not just calldata decoding. Add a `createViemTest` fork test for at least one authorization-bearing path (e.g. `borrow`) that: reads the requirement with `supportSignature: true`, signs it, calls `buildTx([authorizationSignature])`, submits the bundle, and asserts GeneralAdapter1 ends up authorized in-bundle and the operation succeeds.

packages/morpho-sdk/src/entities/blue/blue.ts
```diff
@@ -672,6 +697,7 @@ export class MorphoBlue implements BlueActions {
             receiver,
             minSharePrice,
             reallocations,
+            authorizationSignature: signatures?.find(isAuthorizationSignature),
```
Foulks-Plb: **[Medium] `supportSignature` path silently degrades when the authorization signature is absent.**

With `supportSignature: true`, `getRequirements()` returns a signable `Requirement` that the integrator must sign and fold back in. But `buildTx` derives the auth purely via `signatures?.find(isAuthorizationSignature)` (here and at lines 825, 1148, 1255, 1485) with no guard. If the integrator calls `buildTx()`, `buildTx([])`, or `buildTx([onlyPermitSig])`, `.find` returns `undefined`, the action layer omits the `setAuthorizationWithSig` call entirely, and a fully-built, signable bundle is produced **without authorization** — so the on-behalf `morphoBorrow`/`morphoWithdraw` reverts opaquely on-chain after gas is spent. That's the exact opaque-revert failure `getAuthorizationAction`'s own JSDoc claims to prevent.

Unlike the `supportSignature: false` path (where auth is a separate user-submitted tx, so the bundle never *assumes* in-bundle auth), this path moves auth into the bundle yet leaves its presence unverified. The entity is the only layer that knows auth was required (it called `getMorphoAuthorizationRequirement`). Consider surfacing a typed error when a required authorization signature is missing rather than silently building an unauthorized bundle. (Acknowledged this mirrors the existing permit-signature fallback pattern — but the permit path falls back to `erc20TransferFrom`, whereas here there is no fallback.)

packages/morpho-sdk/src/types/action.ts
```diff
@@ -375,16 +428,44 @@ export function isRequirementAuthorization(
...
   );
 }
+
+/**
+ * Narrows a {@link RequirementSignature} to a permit / Permit2 token-approval signature.
+ *
+ * @param signature - The signed requirement to test.
+ * @returns `true` when `signature.action.type` is `"permit"` or `"permit2"`.
+ */
+export function isPermitSignature(
```
Foulks-Plb, line 457: **[Low] New exported type guards `isPermitSignature` / `isAuthorizationSignature` have no direct unit test.**

They're only exercised indirectly through `buildTx`'s `.find(...)` calls. Per AGENTS.md §5 (every exported function has a unit test), add a small colocated `src/types/action.test.ts` asserting both branches for each guard (`'permit'`/`'permit2'` → permit-true/auth-false, `'authorization'` → auth-true/permit-false). Pure functions — no client or fork needed.

packages/wdk-protocol-lending-morpho-evm/src/morpho-protocol-evm.ts
```diff
@@ -659,7 +665,7 @@ export default class MorphoProtocolEvm extends LendingProtocol {
    */
   async getBorrowRequirements(
     options: MorphoBorrowOptions,
-  ): Promise<RequirementAuthorization[]> {
+  ): Promise<(RequirementAuthorization | RequirementSignatureRequest)[]> {
```
Foulks-Plb, line 677: **[Low] Stale `@returns` on `getBorrowRequirements`.**

The return type was widened to `Promise<(RequirementAuthorization | RequirementSignatureRequest)[]>`, but the JSDoc `@returns` still says only "Authorization requirements." Update it to note it can now also surface a signable authorization requirement (`RequirementSignatureRequest`) when offchain signatures are enabled.

Foulks-Plb (commented):

**Reviewed commit:** `9feabee`

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 1 |
| Medium | 2 |
| Low | 2 |

The protocol wiring is solid: 8 of 8 personas verified the `Authorization` struct mapping (`owner`→`authorizer`), the `v: yParity + 27` signature reconstruction (correct for both compact and full sigs), the `nonce(address)` read, `skipRevert: false` fail-safe on the folded path, action ordering, the chainId-bound EIP-712 domain (no cross-chain replay), full barrel re-exports, and action-layer purity (builders stay sync; signing is confined to the `encode/` boundary). Findings below are about a missing end-to-end fork test, a silent-degradation footgun on the new signature path, a peer-dependent changeset gap, and minor coverage/doc items.

_Automated parallel review (event=COMMENT — no approve/request-changes). 8 personas fired, 0 failed._

Foulks-Plb:
@codex review

Foulks-Plb:
@codex review

packages/morpho-sdk/src/actions/signatures/getBlueAuthorizationAction.ts
```diff
@@ -0,0 +1,57 @@
...
+  } = getChainAddresses(chainId);
+
+  if (!isAddressEqual(authorized, generalAdapter1)) {
+    throw new BundlerErrors.UnexpectedSignature(authorized);
+  }
+
+  return {
+    type: "morphoSetAuthorizationWithSig",
+    args: [
+      { authorizer: owner, authorized, isAuthorized, nonce, deadline },
```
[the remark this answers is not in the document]
Foulks-Plb: keep this operation unrestricted so it can be used for revoke or authorize.

Foulks-Plb:
@prd-carapulse[bot], please audit this PR with a focus on the security of the transaction flow.

Please pay particular attention to:

Zero possibility of fund loss
Signature validation and handling
Race conditions / TOCTOU issues
Replay attack vectors
Approval and allowance safety
Asset flow correctness (no unexpected transfers)
Reentrancy assumptions and callback safety
Error handling and failure modes (no partial or inconsistent state)
Edge cases and invariant violations

approve if all good

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -870,6 +883,76 @@ export namespace BundlerAction {
...
+    chainId: number,
+    authorization: Authorization,
+    signature: Hex,
+    skipRevert = true,
+  ): BundlerCall[] {
+    const {
+      morpho,
+      bundler3: { bundler3 },
+    } = getChainAddresses(chainId);
+    const { r, s, yParity } = parseSignature(signature);
```
Rubilmax: is a `Signature` more often handled than a hex-encoded signature? Study viem's api interface to determine which of the 2 is the most often returned/used/handled as a developer
Then pass the most consumed type as input `signature` (hex or `Signature` object)

packages/morpho-sdk/src/bundler/actions.ts
```diff
@@ -870,6 +883,76 @@ export namespace BundlerAction {
...
+
+    if (isAddressEqual(authorization.authorized, bundler3)) {
```
Rubilmax: isn't it supposed to be the general adapter?

packages/morpho-sdk/src/actions/blue/supply.ts
```diff
@@ -29,7 +29,7 @@ export interface BlueSupplyParams {
     onBehalf: Address;
     /** Maximum supply share price (in ray). Slippage protection against inflation attacks. */
     maxSharePrice: bigint;
-    requirementSignature?: RequirementSignature;
+    requirementSignature?: PermitRequirementSignature;
```
Rubilmax, line 38: missing jsdocs (in multiple places actually)

packages/morpho-sdk/src/actions/signatures/getBlueAuthorizationAction.ts
Rubilmax, line 1: `getBlueAuthorizationAction`

Rubilmax: What's the difference between this and `getMorphoAuthorizationRequirement`, besides this being a signature? Why does it deserve the name of `action` in this case?

Im afraid we're missing the opportunity to merge and simplify things here

packages/morpho-sdk/src/actions/blue/repayWithdrawCollateral.ts
```diff
@@ -41,7 +43,14 @@ export interface BlueRepayWithdrawCollateralParams {
...
     maxSharePrice: bigint;
-    requirementSignature?: RequirementSignature;
+    /** Optional pre-signed permit/permit2 approval for the loan-token transfer. */
+    requirementSignature?: PermitRequirementSignature;
+    /**
+     * Optional signed Morpho authorization. When provided, a `setAuthorizationWithSig` call is
+     * prepended to the bundle so GeneralAdapter1 is authorized in-bundle instead of via a
+     * standalone `setAuthorization` transaction.
+     */
+    authorizationSignature?: AuthorizationRequirementSignature;
```
Rubilmax, line 53: Same spirit: 2 kinds of signatures for 2 pre-pended actions in the resulting array. Probably the opportunity to have this system scale (also arguable: we don't foresee any possible need for scalability in here)

Rubilmax:
Here is an extract of a prompt, that is interesting:

> `getRequirementsAction` is the build-time encoder for signed ERC-20 spending requirements. Given a signed `permit` or `permit2`, it validates that the signed `asset` and `amount` match the transfer being built, then emits the bundler actions that both apply the approval and pull tokens into the recipient, usually `generalAdapter1`: [getRequirementsAction.ts](/Users/rubilmax/conductor/workspaces/sdks/london-v1/packages/morpho-sdk/src/actions/requirements/getRequirementsAction.ts:22).
> 
> It emits:
> 
> - `permit` + `erc20TransferFrom` for ERC-2612 permits
> - `approve2` + `transferFrom2` for Permit2
> - typed errors if asset/amount/Permit2 expiration don’t line up
> 
> `getAuthorizationAction` is related, but it is not the same requirement class. It consumes a signed Morpho authorization and emits one `morphoSetAuthorizationWithSig` action, after checking the signature authorizes the chain’s `GeneralAdapter1`: [getAuthorizationAction.ts](/Users/rubilmax/conductor/workspaces/sdks/london-v1/packages/morpho-sdk/src/actions/requirements/getAuthorizationAction.ts:9).
> 
> I would not fold `getAuthorizationAction` into `getRequirementsAction` as it exists today. The current split is correct because:
> 
> - token requirements need `asset`, `amount`, and `recipient`
> - authorization needs `chainId` and validates `authorized === generalAdapter1`
> - some flows need only permit, some only authorization, and some both
> - the type system already models the distinction with `PermitRequirementSignature` vs `AuthorizationRequirementSignature`, then `selectRequirementSignatures()` splits them before `buildTx`: [action.ts](/Users/rubilmax/conductor/workspaces/sdks/london-v1/packages/morpho-sdk/src/types/action.ts:397)
> 
> At the entity API level, authorization already can be “part of requirements”: for flows that need both, `getRequirements()` returns ERC-20 requirements plus `getMorphoAuthorizationRequirement()`, and `buildTx()` accepts both signatures: [blue.ts](/Users/rubilmax/conductor/workspaces/sdks/london-v1/packages/morpho-sdk/src/entities/blue/blue.ts:1135).
> 
> **The main issue is naming. `getRequirementsAction` sounds like it handles every signed requirement, but it only handles permit/Permit2 token-spend requirements. A clearer name would be something like `getPermitRequirementActions` or `getTokenRequirementActions`. If you want one public convenience helper, I’d add a new wrapper that dispatches by `signature.action.type`, rather than expanding this helper’s responsibility.**

packages/morpho-sdk/src/actions/signatures/getBlueAuthorizationAction.ts
Rubilmax: Suggested reply:

`getMorphoAuthorizationRequirement` is the async resolver: it reads `isAuthorized`/`nonce` and returns either `null`, a standalone `setAuthorization` tx, or a signable `Requirement`.

`getAuthorizationAction` is the pure build-time encoder for the already-signed requirement: it validates the signed `authorized` target and turns it into the bundler `morphoSetAuthorizationWithSig` action.

So the split is:

- requirement helper: “does the user need to do/sign anything?”
- action helper: “given the signature, what bundler action goes into `encodeBundle`?”

I agree the naming is a bit broad. It mirrors `getRequirementsAction`, but `getSignedAuthorizationAction` or `getAuthorizationRequirementAction` would probably make the boundary clearer. I don’t think these should merge with `getMorphoAuthorizationRequirement`, though, because that would mix async RPC resolution with pure action encoding.


Probably worth separating action getters and requirements here. Action getters should not stand in the `requirements` folder

packages/morpho-sdk/src/bundler/actions.ts
Rubilmax: In midnight, i union Hex and Signature to be compatible with everything and it seems not complex

packages/morpho-sdk/src/bundler/actions.ts
Foulks-Plb: It’s the General Adapter that manage your position, so throwing when it’s Bundler3 is correct?
However, the throw could instead be authorized != GeneralAdapter.

Foulks-Plb: To stay consistent in the SDK, since we have another check like this elsewhere.

packages/morpho-sdk/src/bundler/actions.ts
Foulks-Plb: Done in 15e0ae57 — `permit`, `approve2`, and `morphoSetAuthorizationWithSig` now accept `Hex | Signature`, normalized through a single `toSignatureHex` helper. viem hands you a `Hex` from `signTypedData`/`signMessage` and a `Signature` object from the low-level `sign`, so both are accepted with no conversion at the call site.

packages/morpho-sdk/src/actions/blue/supply.ts
Foulks-Plb: Added in 15e0ae57 — documented the previously-bare `args` fields across the Blue builders (`requirementSignature`, `onBehalf`, `receiver`, `amount`, `borrowAmount`, `user`, `collateralAmount`).

packages/morpho-sdk/src/actions/signatures/getBlueAuthorizationAction.ts
Foulks-Plb: Moved in 15e0ae57 — both `getAuthorizationAction` and `getRequirementsAction` now live in a new `actions/signatures/` folder, out of `requirements/`. The split is: `requirements/` = async resolvers that read on-chain state, `signatures/` = pure encoders that turn an already-signed requirement into the bundler `Action`s prepended to a bundle. Kept the names since they return an `Action`, and did not merge them with `getMorphoAuthorizationRequirement` for the reason you noted (would mix async RPC resolution with pure encoding).

packages/morpho-sdk/src/actions/blue/repayWithdrawCollateral.ts
Foulks-Plb: Kept the two explicit branches for now. Only two prerequisites are ever possible on these paths (permit + authorization), the two encoders need different context (`getRequirementsAction` needs asset/amount/recipient, `getAuthorizationAction` does not), and the ordering of the prepended actions is load-bearing. A generic signature->actions dispatcher would add indirection for scalability we don't foresee, so leaving it as-is — happy to revisit if a third signature kind ever appears.

Rubilmax (approved):

Rubilmax (approved):

Rubilmax (approved):

PR #845, opened 2026-07-03
feat(blue-sdk-viem): add mega deployless fetch accrual vaultV2

Foulks-Plb:
## Motivation

`fetchAccrualVaultV2` makes a chain of sequential RPC round-trips imposed by the VaultV2 architecture — it fetches the vault, then resolves and fetches each adapter, then each adapter's Morpho Blue markets or wrapped MetaMorpho V1 vault. Deployless bytecode reads can collapse this waterfall into a single call.

## Solution

Adds `fetchAccrualVaultV2Deployless`, a deployless-only reader (no multicall fallback) that traverses the entire accrual tree on-chain via a new `GetAccrualVaultV2` Solidity query and returns a fully-hydrated `AccrualVaultV2` from one `eth_call`. The result is behaviourally identical to `fetchAccrualVaultV2` (same `maxDeposit`/`maxWithdraw`/`accrueInterest`/per-adapter `realAssets`); the nested MetaMorpho V1 vault omits only two capacity-irrelevant optional fields (`eip5267Domain`, `publicAllocatorConfig`) to keep the query bytecode small (~13.5 KB, well under limits). Covered by unit tests for the decoder and typed-error mapping, plus fork tests comparing the two readers across all three adapter types (MorphoVaultV1, MorphoMarketV1, MorphoMarketV1AdapterV2).


Foulks-Plb:
@codex review

packages/blue-sdk-viem/test/AccrualVaultV2Deployless.test.ts
```diff
@@ -0,0 +1,211 @@
...
+  );
+}
+
+describe("fetchAccrualVaultV2Deployless", () => {
+  vaultV2Test(
+    "matches fetchAccrualVaultV2 for a MorphoVaultV1 liquidity adapter",
+    async ({ client }) => {
+      const [deployless, sequential] = await Promise.all([
+        fetchAccrualVaultV2Deployless(vaultV2VaultV1, client),
+        fetchAccrualVaultV2(vaultV2VaultV1, client),
```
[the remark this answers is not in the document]
Foulks-Plb: already done

Foulks-Plb:
@codex review

Foulks-Plb:
@prd-carapulse[bot], test this against the current production Vault V2s and tell me the average size of the objects returned by this single deployless call? 
Also, check whether the payload size stays within typical EVM/RPC limits?

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -394,6 +414,26 @@ export async function fetchAccrualVaultV2(
 ) {
   parameters.chainId ??= await getChainId(client);
 
+  const { deployless = true } = parameters;
+
+  // The entire accrual tree can be read in a single deployless call. When it succeeds there is
+  // nothing left to fetch, so return early and skip the sequential multicall fan-out below.
+  if (deployless) {
+    try {
+      return await fetchAccrualVaultV2Deployless(address, client, parameters);
```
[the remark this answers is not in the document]
Foulks-Plb: Fixed in e8ad840b (commit db5866e6). The deployless `GetAccrualVaultV2` query now reads the nested MetaMorpho V1 vault's `eip5267Domain` and PublicAllocator config (vault-level `admin`/`fee`/`accruedFee` and per-market flow caps) in the same single `eth_call`, so the default path stays at parity with the multicall path without any extra sequential reads.

Details:
- Added `eip5267Domain` to `VaultV1Config` (read via the `eip712Domain()` staticcall workaround, same as `GetVault`), `publicAllocatorConfig` + a `hasPublicAllocator` flag to `VaultV1Response`, and `flowCapMaxIn`/`flowCapMaxOut` to `VaultV1MarketAllocation`. The chain PublicAllocator address is threaded into the query and its VaultV1 helpers.
- The vault-level config is gated on `isAllocator` via `hasPublicAllocator`, so the decoder returns `undefined` (not a zeroed struct) when the vault has not set the PublicAllocator, exactly matching `fetchVault`'s multicall gate. Per-market flow caps are read whenever the chain has a PublicAllocator, matching `fetchVaultMarketPublicAllocatorConfig`.
- The fork test now asserts these fields for exact parity; a mock-based unit test covers the decode.

packages/blue-sdk-viem/contracts/vault-v2/GetAccrualVaultV2.sol
```diff
@@ -0,0 +1,418 @@
...
+            IMorphoMarketV1Adapter typed = IMorphoMarketV1Adapter(adapter);
+            res.parentVault = typed.parentVault();
+            res.skimRecipient = typed.skimRecipient();
+
+            uint256 length = typed.marketParamsListLength();
+            res.marketV1Positions = new MarketV1Position[](length);
+            for (uint256 i; i < length; ++i) {
+                Id id = _id(typed.marketParamsList(i));
+                res.marketV1Positions[i].position = morpho.position(id, adapter);
+                res.marketV1Positions[i].market = _queryMarket(morpho, id, adaptiveCurveIrm);
```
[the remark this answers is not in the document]
Foulks-Plb: Fixed in e8ad840b. The `MorphoMarketV1Adapter` branch of the deployless query now carries the adapter's configured `marketParams` through the response instead of deriving them from `idToMarketParams`.

Details:
- Added `MarketParams marketParams` to the `MarketV1Position` struct, populated from `typed.marketParamsList(i)` (read directly from the adapter), so the params survive even for a market listed on the adapter but not yet created in Morpho.
- The decoder now rebuilds `VaultV2MorphoMarketV1Adapter.marketParamsList` from `entry.marketParams` rather than the (potentially zeroed) `MarketResponse.marketParams`.
- Added a unit test with a market whose `idToMarketParams` is zeroed but whose adapter params are non-zero, asserting `marketParamsList` keeps the real params.

Note: the per-position `AccrualPosition.marketId` still derives from the market object (built from `idToMarketParams`), which matches the old multicall path — its `fetchMarket` also reads `idToMarketParams`, so both paths produce the same (empty) market for a not-yet-created market. Only `marketParamsList` was losing data.

Foulks-Plb:
@codex review

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,346 @@ export async function fetchAccrualVaultV2(
...
+type MarketQueryResponse =
+  AdapterQueryResponse["marketV1Positions"][number]["market"];
+
+/** @internal Mirrors the `AdapterType` enum encoded by `GetAccrualVaultV2.sol`. */
+const AdapterType = {
+  Unknown: 0,
+  MorphoVaultV1: 1,
+  MorphoMarketV1: 2,
+  MorphoMarketV1AdapterV2: 3,
+} as const;
```
Rubilmax, line 514: use ts enum

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,346 @@ export async function fetchAccrualVaultV2(
...
+type AccrualVaultV2QueryResponse = ContractFunctionReturnType<
+  typeof getAccrualVaultV2Abi,
+  "view",
+  "query"
+>;
+/** @internal One adapter entry of the deployless `GetAccrualVaultV2.query` response. */
+type AdapterQueryResponse = AccrualVaultV2QueryResponse["adapters"][number];
+/** @internal One market entry of the deployless `GetAccrualVaultV2.query` response. */
+type MarketQueryResponse =
+  AdapterQueryResponse["marketV1Positions"][number]["market"];
```
Rubilmax: are this type really useful? why not inline?

Rubilmax (commented):
it feels the agent over-factorized the code here, as it does often: a lot of things can be inlined and it feels easier to review
anyway it's not blocking as it's all internal

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Good call — inlined

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: I'd keep the `as const` map here rather than a TS `enum`, for a concrete reason beyond house style (§3 favours `as const` for protocol/ABI literals): `adapter.adapterType` comes back from the `eth_call` decoded by viem as a plain `number`. A nominal `enum` type wouldn't accept that `number` without a cast (`x as AdapterType`) at every comparison — which §2 bans — whereas the `as const` map yields a literal union that the decoded `number` matches directly. It also avoids the numeric-enum runtime reverse-mapping and keeps the module tree-shakeable (§4).

Happy to add `satisfies Record<string, number>` for extra safety if you want, but I left it untouched.

Foulks-Plb:
On the over-factorization: I audited the new code and the rest holds up on call-site count — `toMarket` has 3 sites, `toAccrualAdapter` is a ~150-line switch I wouldn't want inlined at its 2 sites, and the error helpers are shared with the multicall path. The one genuine over-factorization was the type-alias chain, so I trimmed it to what earns its keep (details in the two inline threads). Let me know if you want me to push the cursor further.

Foulks-Plb:
@codex review

Rubilmax (commented):

**Reviewed commit:** `d47ae08`

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 2 |
| Medium | 3 |
| Low | 6 |

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,341 @@ export async function fetchAccrualVaultV2(
...
+ * );
+ * ```
+ */
+// biome-ignore lint/complexity/useMaxParams: TODO refactor to ≤2 params
+export async function fetchAccrualVaultV2Deployless(
+  address: Address,
+  client: Client,
+  parameters: DeploylessFetchParameters = {},
+) {
+  parameters.chainId ??= await getChainId(client);
```
Rubilmax: **High:** `fetchAccrualVaultV2Deployless` mutates the caller-owned `parameters` object by assigning `chainId`, violating AGENTS.md §2's ban on input mutation. Reusing that object with another client will retain the first client's chain ID; derive a new parameters object instead.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,341 @@ export async function fetchAccrualVaultV2(
...
+
+    default:
+      throw new UnsupportedVaultV2AdapterError(adapter.adapter);
+  }
+}
+
+/**
+ * Fetches the full VaultV2 accrual tree in a single deployless call.
+ *
+ * Unlike {@link fetchAccrualVaultV2}, which chains sequential reads (vault, then each adapter, then
```
Rubilmax: **High:** This public JSDoc says `fetchAccrualVaultV2` chains sequential reads, but this diff makes that function use the deployless reader by default and reserve sequential reads for fallback or explicit opt-out. Rewrite the comparison so it accurately distinguishes this no-fallback entry point from the default wrapper.

packages/blue-sdk-viem/contracts/vault-v2/GetAccrualVaultV2.sol
```diff
@@ -0,0 +1,495 @@
...
+import {IMorphoMarketV1AdapterV2} from "./interfaces/IMorphoMarketV1AdapterV2.sol";
+import {IMorpho, Id, MarketParams, Market, Position} from "../interfaces/IMorpho.sol";
+import {IMetaMorpho, MarketConfig, PendingUint192, PendingAddress} from "../interfaces/IMetaMorpho.sol";
+import {IOracle} from "../interfaces/IOracle.sol";
+import {IAdaptiveCurveIrm} from "../interfaces/IAdaptiveCurveIrm.sol";
+import {IERC20} from "../interfaces/IERC20.sol";
+import {IERC20Permit, Eip5267Domain} from "../interfaces/IERC20Permit.sol";
+import {IPublicAllocator, FlowCaps} from "../interfaces/IPublicAllocator.sol";
+
+error UnknownOfFactory(address factory, address vault);
```
Rubilmax, line 20: **Medium:** This query duplicates errors, `Token`/`VaultV2Allocation` shapes, base VaultV2 reads, and liquidity-allocation logic already owned by `GetVaultV2.sol`. That creates two sources of truth for the same deployless state mapping and lets the standard and accrual fetchers drift. Extract the shared Solidity types and base-query logic into one module used by both queries.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,341 @@ export async function fetchAccrualVaultV2(
...
+    address: adapter.adapter,
+    parentVault: adapter.parentVault,
+    skimRecipient: adapter.skimRecipient,
+  };
+
+  switch (adapter.adapterType) {
+    case AdapterType.MorphoVaultV1: {
+      const { morphoVaultV1, vaultV1 } = adapter;
+
+      const vault = new Vault({
```
Rubilmax, line 549: **Medium:** The `MorphoVaultV1` branch constructs `Vault` directly without preserving its normalized 18 share-token decimals. `VaultToken` therefore inherits `Token`'s default of 0, while the sequential path constructs through `VaultConfig` and returns 18. This breaks the promised deployless/multicall parity; construct the nested vault through `VaultConfig` or otherwise preserve the invariant.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,341 @@ export async function fetchAccrualVaultV2(
...
+ *   vaultV2Address,
+ *   client,
+ * );
+ * ```
+ */
+// biome-ignore lint/complexity/useMaxParams: TODO refactor to ≤2 params
+export async function fetchAccrualVaultV2Deployless(
+  address: Address,
+  client: Client,
+  parameters: DeploylessFetchParameters = {},
```
Rubilmax: **Medium:** This deployless-only API accepts `DeploylessFetchParameters`, whose public contract says `deployless: false` selects multicall, but the function always performs the bytecode call and ignores that choice. Use `FetchParameters` or another type that omits `deployless` so callers cannot request unsupported behavior.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -394,6 +416,26 @@ export async function fetchAccrualVaultV2(
...
 
+  const { deployless = true } = parameters;
+
+  // The entire accrual tree can be read in a single deployless call. When it succeeds there is
+  // nothing left to fetch, so return early and skip the sequential multicall fan-out below.
+  if (deployless) {
+    try {
+      return await fetchAccrualVaultV2Deployless(address, client, parameters);
+    } catch (error) {
+      if (deployless === "force") throw error;
```
Rubilmax, line 433: **Low:** The modified wrapper covers generic fallback and deployless success, but not the new `deployless: "force"` branch or the typed deterministic-error branches that must avoid fallback. Add focused tests for those control-flow paths.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,341 @@ export async function fetchAccrualVaultV2(
...
+    rateAtTarget:
+      response.marketParams.irm === adaptiveCurveIrm
+        ? response.rateAtTarget
+        : undefined,
+  });
+}
+
+/** @internal Rebuilds one accrued adapter from a deployless `GetAccrualVaultV2` adapter response. */
+// biome-ignore lint/complexity/useMaxParams: internal decoder threading chain-resolved addresses
+function toAccrualAdapter(
```
Rubilmax, line 534: **Low:** `toAccrualAdapter` is a new local non-exported helper with only two call sites, contrary to the root §1 three-call-site extraction threshold. Restructure the liquidity and regular adapter decoding to avoid the two-use local extraction.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,341 @@ export async function fetchAccrualVaultV2(
...
+ * @param parameters.account - Optional account passed to the viem call.
+ * @param parameters.blockNumber - Optional block number for a historical read.
+ * @param parameters.blockTag - Optional block tag for a historical read.
+ * @param parameters.stateOverride - Optional viem state override.
+ * @param parameters.chainId - Optional chain id; defaults to `getChainId(client)`.
+ * @returns The hydrated `AccrualVaultV2` entity with asset balance, accrued liquidity and regular
+ *   adapters, and force-deallocate penalties.
+ * @throws {UnknownFactory} when the configured chain has no VaultV2 factory.
+ * @throws {UnknownOfFactory} when `address` is not a VaultV2 from the configured factory.
+ * @throws {UnsupportedVaultV2AdapterError} when the vault or one of its adapters uses an unsupported
```
Rubilmax, line 726: **Low:** The `@throws` inventory omits RPC/decode failures rethrown from `readContract`. The canonical JSDoc guide requires downstream external errors to be documented in qualified form; add the appropriate `@throws {viem.BaseError}` entry.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
```diff
@@ -444,3 +486,341 @@ export async function fetchAccrualVaultV2(
...
+    _totalAssets: response._totalAssets,
+    totalSupply: response.totalSupply,
+    virtualShares: response.virtualShares,
+    maxRate: response.maxRate,
+    lastUpdate: response.lastUpdate,
+    address,
+    adapters: response.adapters.map((adapter) => adapter.adapter),
+    liquidityAdapter: response.liquidityAdapter,
+    liquidityData: response.liquidityData,
+    liquidityAllocations: response.isLiquidityAdapterKnown
```
Rubilmax, line 812: **Low:** All direct deployless fixtures use a known, nonzero liquidity adapter, leaving the new `isLiquidityAdapterKnown === false` and `hasLiquidityAdapter === false` decoding paths untested. Add fixtures for both branches.

packages/blue-sdk-viem/contracts/vault-v2/GetAccrualVaultV2.sol
```diff
@@ -0,0 +1,495 @@
...
+                morphoMarketV1AdapterV2Factory,
+                morpho,
+                adaptiveCurveIrm,
+                publicAllocator
+            );
+        }
+    }
+
+    /// @dev Computes the configured liquidity adapter's cap allocations, mirroring `GetVaultV2`.
+    function _queryLiquidityAllocations(
```
Rubilmax, line 237: **Low:** Several new private helpers have fewer than three call sites: `_queryLiquidityAllocations`, `_queryAdapter`, `_queryVaultV1`, `_queryVaultV1Allocation`, `_id`, and `_queryEip5267Domain`. This violates the root §1 local-helper threshold; inline or restructure them, or document any compiler-driven exception in the nearest package guidance.

packages/blue-sdk-viem/contracts/vault-v2/GetAccrualVaultV2.sol
```diff
@@ -0,0 +1,495 @@
...
+        ) {
+            res.adapterType = uint8(AdapterType.MorphoMarketV1);
+            IMorphoMarketV1Adapter typed = IMorphoMarketV1Adapter(adapter);
+            res.parentVault = typed.parentVault();
+            res.skimRecipient = typed.skimRecipient();
+
+            uint256 length = typed.marketParamsListLength();
+            res.marketV1Positions = new MarketV1Position[](length);
+            for (uint256 i; i < length; ++i) {
+                MarketParams memory marketParams = typed.marketParamsList(i);
```
Rubilmax, line 330: **Low:** The not-yet-created `MorphoMarketV1` market case is tested only with a fabricated decoded response. Add an Anvil test configuring such a market so the contract-side `marketParamsList`/`idToMarketParams` behavior and generated bytecode are exercised.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Rubilmax: A morpho vault always has decimals defined so it's not an actual issue. But for consistency i would fix this still

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Rubilmax: Use `FetchParameters`

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Rubilmax: Use fork tests if not already btw

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Fixed in 5eee172d. Neither `fetchAccrualVaultV2Deployless` nor the `fetchAccrualVaultV2` wrapper assigns `chainId` onto the caller's argument anymore. The deployless reader resolves `chainId` into a local and threads it through a local `readParameters` copy; the wrapper destructures `{ deployless, ...parameters }` so it defaults `chainId` on its own rest copy (same pattern already used by `fetchVaultV2`). Reusing one `parameters` object across clients no longer retains the first client's chain id.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Fixed in 5eee172d. The JSDoc no longer says `fetchAccrualVaultV2` chains sequential reads. It now states the wrapper defaults to this same single deployless call and only falls back to sequential multicall reads when the call fails, while this entry point is deployless-only — it always performs the single `eth_call` and never falls back.

packages/blue-sdk-viem/contracts/vault-v2/GetAccrualVaultV2.sol
Foulks-Plb: Acknowledged — the `Token`/`VaultV2Allocation` structs and the base VaultV2 reads are indeed shared with `GetVaultV2.sol`. I've left them duplicated deliberately for now: (1) each deployless query compiles to a single self-contained contract whose bytecode we pin in-package, and the accrual query's response shape and traversal genuinely diverge (full adapter tree, nested MetaMorpho V1, PA config) rather than reusing `VaultV2Response`; (2) extracting a shared base pulls directly against the sibling Low findings on this same file asking to *inline* the small private helpers, so the net single-source-of-truth win is mostly the struct declarations. Since it's internal-only (matching your earlier note that the over-factorization here isn't blocking), I kept the two queries independently readable. I can extract a shared `*.sol` types/library module in a follow-up if you'd rather converge them.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Fixed in 5eee172d. The nested MetaMorpho V1 vault is now built through `new VaultConfig({ ... })` (like the multicall `fetchVault` path), so the share token keeps its normalized 18 decimals instead of `Token`'s default of 0. As you said, a Morpho vault always defines decimals so it wasn't an actual runtime bug — but the two paths are now consistent and the fork parity test covers it.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Fixed in 5eee172d — `fetchAccrualVaultV2Deployless` now takes `FetchParameters`, so callers can no longer pass a `deployless` toggle that this deployless-only reader would silently ignore.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Fixed in 5eee172d. Added wrapper unit tests covering the previously-untested control flow: `deployless: "force"` rethrows without falling back, and each deterministic typed error (`UnknownFactory`, `UnknownOfFactory`, `UnsupportedVaultV2AdapterError`) is rethrown without a multicall retry. Each asserts by mocking only the single-call response, so a fallback attempt would surface a different error and fail the test.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Kept as-is (with rationale). `toAccrualAdapter` is the single ~90-line `switch` that decodes every adapter kind into the discriminated `IAccrualVaultV2Adapter` union; it has two call sites (the liquidity adapter and each regular adapter). Inlining it into both would duplicate the whole switch and hurt reviewability — the opposite of what the three-call-site rule is meant to prevent. Per §9 I'd treat this as a documented local exception rather than force a worse shape. Happy to add a one-line note to the package `AGENTS.md` if you'd like it recorded.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Fixed in 5eee172d. Added `@throws {viem.BaseError}` to `fetchAccrualVaultV2Deployless` (deployless read/decode failure, no fallback) and to `fetchAccrualVaultV2` (a read failing with no fallback left), per `docs/jsdoc-style.md` §5.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Foulks-Plb: Fixed in 5eee172d. Added a fork test (`matches fetchAccrualVaultV2 for a vault without a liquidity adapter`) that deploys a bare VaultV2 — no liquidity adapter, no regular adapters — exercising both `hasLiquidityAdapter === false` and `isLiquidityAdapterKnown === false` against the real contract and asserting full parity with the multicall path. Also added mock decode unit tests for the `isLiquidityAdapterKnown: false` and `hasLiquidityAdapter: false` branches.

packages/blue-sdk-viem/contracts/vault-v2/GetAccrualVaultV2.sol
Foulks-Plb: The only honest justification left is readability: the helpers split one large `query`/`_queryAdapter` into named steps. `_queryAdapter` genuinely has two call sites (liquidity adapter + the adapter loop). Since inlining is essentially free here, I'm happy to fold the single-call-site helpers in if you'd prefer that — just say the word and I'll inline + recompile the pinned bytecode.

packages/blue-sdk-viem/contracts/vault-v2/GetAccrualVaultV2.sol
Foulks-Plb: The decode side of this case is covered by the `keeps a MorphoMarketV1 adapter's configured params for a market not yet created` unit test (adapter `marketParams` non-zero, `idToMarketParams` zeroed). An on-chain Anvil fixture would need to add a market to a `MorphoMarketV1Adapter`'s `marketParamsList` without creating it in Morpho, which the current test utils (`deployMorphoMarketV1Adapter`) don't expose; the created-market path is already exercised end-to-end by the existing fork tests. I can add the util + Anvil fixture in a follow-up if you want the contract-side `marketParamsList`/`idToMarketParams` branch run on a fork too.

packages/blue-sdk-viem/src/fetch/vault-v2/VaultV2.ts
Rubilmax: Fine because internal

Rubilmax (approved):

PR #865, opened 2026-07-10
fix(evm-simulation): block native ETH retained by bundler3 on Tenderly

Foulks-Plb:
## Context

Cantina **finding 1440** — *`simulate()` reports no funds retained by bundler3 while native ETH is retained* (SDK-277, Sev 2 High). Confirmed as a real code bug on `main`.

`assertNoBundlerRetention` only inspected parsed `Transfer` logs. Native ETH **emits no event log**, so it never reaches `transfers` on the **Tenderly primary backend** — Tenderly derives native value into `assetChanges` instead. As a result, native ETH stuck in a bundler3 address silently passed the "never bypassable" retention guard on the default path. The `eth_simulateV1` fallback caught it (via `traceTransfers` synthetic `ethAddress` logs), but that is the fallback, not the primary path.

Verified against Tenderly docs + EVM fundamentals: a native ETH transfer produces no `LOG`, so a log-only parser structurally cannot see raw native retention on Tenderly.

## Fix

- `assertNoBundlerRetention` now reads **native ETH from `assetChanges`** (the normalized, cross-backend source of truth — both backends populate it with `ethAddress`), while **ERC20/WETH retention keeps coming from transfer logs** (unchanged behavior).
- **No double-counting on `eth_simulateV1`**: native moves exist there both as a synthetic `ethAddress` transfer log *and* in the `assetChanges` derived from it. Native transfer logs are only used to backfill bundler endpoints **absent** from `assetChanges`.
- `assetChanges` is now a required parameter of the internal check; `simulate()` passes `result.assetChanges` through.

## Tests

Added regression coverage in `bundler-retention.spec.ts` and an end-to-end case in `simulate.spec.ts`:

- **Tenderly path**: native ETH present only in `assetChanges` (no logs) → throws `BlacklistViolationError` (the exact finding-1440 scenario).
- **eth_simulateV1 path**: native ETH in both a log and `assetChanges` → counted **once** (`netRetained` == amount, not doubled).
- Fallback (native log, empty `assetChanges`) → throws.
- Native pass-through / below-dust / non-bundler account → no throw.
- Mixed ERC20-log + native-assetChanges retention reported together.

**Proof the tests have teeth**: with the `assetChanges` path disabled, exactly the 3 native-ETH Tenderly regression tests fail; restored → `234 passed`. `tsc --noEmit`, `biome check`, and the full `evm-simulation` suite pass.

Patch changeset included. No package depends on `@morpho-org/evm-simulation`, so no dependency cascade.

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/865" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/native-eth-bundler-retention.md
+++ b/.changeset/native-eth-bundler-retention.md
@@ -1,0 +1,16 @@
+---
+"@morpho-org/evm-simulation": patch
+---
+
+fix(evm-simulation): detect native ETH retained by bundler3 on the Tenderly backend
+
+`assertNoBundlerRetention` only inspected parsed `Transfer` logs, so native ETH
+retained by a bundler3 address slipped through the guard on the Tenderly primary
+backend — native ETH emits no event log, and Tenderly derives it into
+`assetChanges` rather than synthetic transfer logs (Cantina finding 1440).
+
+The retention check now also reads native ETH from `assetChanges` (the
+cross-backend source of truth), while ERC20/WETH retention keeps coming from
+transfer logs. Native transfer logs (the `eth_simulateV1` synthetic sentinel)
+are only used as a fallback for bundler addresses absent from `assetChanges`, so
+native moves are never double-counted on `eth_simulateV1`.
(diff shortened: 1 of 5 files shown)


Foulks-Plb:
@codex review

Rubilmax (approved):

PR #866, opened 2026-07-10
feat: default shared-liquidity target utilization to 90% and deprecate its tuning surface

Foulks-Plb:
## Motivation

The shared-liquidity algorithm exposes configurable supply-target and source-market withdrawal utilizations (via `PublicAllocatorOptions`, `ReallocationComputeOptions`, and `liquidity-sdk-viem`'s `LiquidityParameters` / the API's `targetWithdrawUtilization`). We want both knobs to converge on a single 90% default and to stop advertising them as tunable, without breaking existing integrators.

## Solution

Non-breaking (minor): the defaults move to 90% and the whole tuning surface is `@deprecated` but still fully honored until the next major.

**`@morpho-org/morpho-sdk`** (minor):
- `DEFAULT_SUPPLY_TARGET_UTILIZATION` and `DEFAULT_WITHDRAWAL_TARGET_UTILIZATION` are now both 90% (previously 90.5% and 92%). Callers passing no override trigger reallocations at 90% and cap phase-1 source-market withdrawals at 90%; the aggressive phase-2 fallback still drains to 100% as a last resort.
- `PublicAllocatorOptions.maxWithdrawalUtilization` / `defaultMaxWithdrawalUtilization` and `ReallocationComputeOptions.supplyTargetUtilization` / `defaultSupplyTargetUtilization` are `@deprecated`. The per-market → default → constant resolution chain is unchanged, so explicit overrides behave exactly as before.

**`@morpho-org/liquidity-sdk-viem`** (minor):
- `LiquidityParameters` and the `LiquidityLoader` `parameters` constructor argument are `@deprecated`. The loader no longer consults the Morpho API's per-market `targetWithdrawUtilization`; the ceiling defaults to 90% via `morpho-sdk`. Explicitly passed `parameters` overrides are still forwarded and honored.
- The `@morpho-org/morpho-sdk` peer range moves to `^5.4.0`, since the 90% default ceiling ships with that release (an older peer would silently keep the 92% default now that the API value is ignored).
- The pinned-block fork tests' expected withdrawal plans are re-recorded against the 90% ceiling.

**`@morpho-org/wdk-protocol-lending-morpho-evm`** (patch): dependency-resolution cascade only.

Tests, JSDoc, and the changeset land with the change.

### DIFF
--- a/.changeset/shared-liquidity-target-utilization-90.md
+++ b/.changeset/shared-liquidity-target-utilization-90.md
@@ -1,0 +1,29 @@
+---
+"@morpho-org/morpho-sdk": minor
+"@morpho-org/liquidity-sdk-viem": minor
+"@morpho-org/wdk-protocol-lending-morpho-evm": patch
+---
+
+Default the shared-liquidity target utilization to 90% and deprecate all
+customization of it. This is non-breaking: the tuning surface stays in place and
+explicit overrides are still honored until the next major.
+
+**`@morpho-org/morpho-sdk`**:
+
+- `DEFAULT_SUPPLY_TARGET_UTILIZATION` and `DEFAULT_WITHDRAWAL_TARGET_UTILIZATION`
+  are now both 90% (previously 90.5% and 92%). Callers that pass no explicit
+  override now trigger reallocations at 90% and cap phase-1 source-market
+  withdrawals at 90%. The aggressive fallback still drains to 100% as a last resort.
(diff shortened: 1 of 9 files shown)


Foulks-Plb:
@codex review

Foulks-Plb:
@codex review

0xbulma (dismissed):

remiroyc (approved):

0xbulma (approved):

PR #868, opened 2026-07-10
docs(evm-simulation): sync README API surface with exports

Foulks-Plb:
## Motivation

The `@morpho-org/evm-simulation` README had drifted from the package's actual public surface: the "API surface" list omitted four exported types (`AccountAssetChanges`, `AssetChange`, `SimulationCall`, `RawLog`), and neither the overview nor the usage example mentioned `assetChanges`/`calls`, which are now core fields of `SimulationResult`.

## Solution

Reorganized the API surface list into Config / Input / Result groups so it matches `src/index.ts` exactly, updated the overview to mention net per-account balance changes, and expanded the usage example to destructure all four result fields with a one-line description of each. Documentation-only change — no runtime source touched, so no changeset is required.

### DIFF
--- a/packages/evm-simulation/README.md
+++ b/packages/evm-simulation/README.md
@@ -2,8 +2,8 @@
 
 ## Overview
 
-EVM transaction simulation engine for Morpho — bundle execution preview
-and transfer parsing.
+EVM transaction simulation engine for Morpho — bundle execution preview,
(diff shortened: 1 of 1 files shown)


Foulks-Plb:
@codex review

Rubilmax (approved):

PR #873, opened 2026-07-13
fix(blue-sdk-viem): skip Permit2 call in deployless holding query when absent

Foulks-Plb:
## Problem

The deployless `GetHolding` query touched Permit2 in two places, unconditionally:

```solidity
res.erc20Allowances = ERC20Allowances({
    ...
    permit2: token.allowance(account, address(permit2)),   // ERC20 allowance granted to Permit2
    ...
});
res.permit2BundlerAllowance = permit2.allowance(account, address(token), generalAdapter1);  // calls Permit2
```

On chains that have no Permit2 deployment, `fetchHolding` defaults `permit2` to `address(0)` (`src/fetch/Holding.ts`). The second line then calls the **Permit2 contract at `address(0)`** (no code) → the typed external call reverts on Solidity's `extcodesize` guard, so the whole deployless query reverts.

Consequences:
- with `deployless: true` (the default, used by `useHolding` / `morpho-sdk`), the deployless read reverts and **silently falls back to multicall** — deployless never helps on those chains;
- with `deployless: "force"`, it **throws** outright.

The multicall fallback already handles this: it returns `0` for both Permit2 allowances **without any read** when the chain has no Permit2 (`getValue(chainAddresses, "permit2") == null → 0n`, and the guarded `chainAddresses.permit2 != null` branch). The deployless path did not, so the two paths diverged.

## Fix

Mirror the fallback in the query contract: gate **both** Permit2 reads on a single `hasPermit2` flag, leaving the Permit2 ERC20 allowance and the bundler allowance at their zero defaults when Permit2 is absent.

```solidity
bool hasPermit2 = address(permit2) != address(0);
res.erc20Allowances = ERC20Allowances({
    morpho:          token.allowance(account, morpho),
    permit2:         hasPermit2 ? token.allowance(account, address(permit2)) : 0,
    generalAdapter1: token.allowance(account, generalAdapter1)
});
if (hasPermit2) {
    res.permit2BundlerAllowance = permit2.allowance(account, address(token), generalAdapter1);
}
```

Gating the first read too (per the Codex review) keeps deployless byte-for-byte equal to the fallback (`0` without a read) and avoids a residual revert on tokens that reject zero-address spender reads.

`src/queries/GetHolding.ts` was regenerated via the repo's compile pipeline (`pnpm --filter @morpho-org/blue-sdk-viem compile`). **The ABI is unchanged**; only the bytecode changed (its metadata tail also moved from solc `0.8.27` → `0.8.35`, the version pinned in the root `package.json`). No other query artifacts are touched.

### Why `0` and not `undefined`

`Holding.erc20Allowances.permit2` (`bigint`) and `permit2BundlerAllowance` (`Permit2Allowance`) are non-optional, so `0` is the type's "empty" value and matches what the fallback has always returned. Modelling Permit2 absence as `undefined` would be a breaking (major) change and is redundant here — unlike per-token signals such as `erc2612Nonce`, Permit2 availability is a per-chain fact consumers already read authoritatively from `getChainAddresses().permit2` (e.g. the `if (permit2)` gate in `getGeneralAdapterRequirements`). Out of scope for this patch.

## Tests

Added a fork test in `test/Holding.test.ts` that invokes the raw deployless `query` on the mainnet fork with `permit2 = address(0)`, approves `morpho` (6) and `generalAdapter1` (4), and asserts the call resolves (no revert) with `erc20Allowances == { morpho: 6n, permit2: 0n, generalAdapter1: 4n }` and a zeroed `permit2BundlerAllowance`. Before this change the query reverts. The 54 mocked unit tests in `src/fetch/fetch.test.ts` still pass locally; fork tests run in CI (no local RPC).

## Dependent audit (per §4/§7)

- `@morpho-org/blue-sdk-viem` — **patch** (this changeset).
- `@morpho-org/morpho-sdk`, `@morpho-org/wdk-protocol-lending-morpho-evm` — direct `workspace:^` runtime dependents; auto-patched by Changesets (`updateInternalDependencies: "patch"` + `bumpVersionsWithWorkspaceProtocolOnly: true`).
- `@morpho-org/liquidity-sdk-viem` — peer `^5.0.0`, already accepts the 5.1.3 → 5.1.4 patch; no range change.
- Frozen deprecated peers (`blue-sdk-wagmi`, `bundler-sdk-viem`, `migration-sdk-viem`, `simulation-sdk-wagmi`) excluded from the cascade.


Foulks-Plb:
@codex review

Rubilmax (approved):

andrefgf:
Nice catch on the `extcodesize` guard. The half of this I find more interesting is the silent one.

With `deployless: true` the revert is swallowed and the fallback returns the same numbers — `0` for both allowances, no read. So the observable output is identical whether deployless worked or died. The only thing that changed is *which path ran*, and no assertion on the result can see that. On the affected chains deployless was doing nothing, for however long, and everything stayed green.

Two questions:

1. Does anything in CI assert that the deployless path actually *executed*, rather than that the value came back right? A spy that fails the test if the fallback fires under `deployless: true` would turn this class from silent into loud.

2. You note the 54 mocked tests in `fetch.test.ts` stayed green and only a hand-written fork test caught it — and fork tests don't run locally. Is per-chain fork coverage practical at the chain count you're at now, or is CI-only the accepted trade-off?

Asking because I work on real-vs-mocked gaps in wallet and dApp flows, and the guard here is per-chain state (`getChainAddresses().permit2`) — so every chain you add is another chance for the two paths to disagree quietly.

PR #875, opened 2026-07-13
refactor(morpho-sdk): unify input validation errors

Rubilmax:
## What changed

- Add `NegativeInputError` and `NonPositiveInputError` to `morpho-sdk`, with readonly `field` and `value` context.
- Replace operation-specific scalar-bound errors across Blue, Midnight, and Vault flows.
- Remove 25 redundant public error classes while retaining domain-specific relational errors.
- Update tests, JSDoc, package guidance, and add a major changeset.

## Why

Scalar validation had accumulated function- and protocol-specific error classes for equivalent negative or non-positive input failures. This made consumers handle multiple class identities for the same corrective action. The shared errors tighten the public API while preserving enough field-level context to identify the invalid input.

## Impact

This is a breaking `morpho-sdk` API change. Consumers handling any removed scalar-bound class should migrate to `NegativeInputError` or `NonPositiveInputError` according to the documented field constraint.

This PR is stacked on #815 so the Midnight action flows land independently from the error-surface consolidation.

## Validation

- `pnpm --filter @morpho-org/morpho-sdk exec tsc --noEmit --pretty false`
- Biome check on all changed TypeScript files
- Focused morpho-sdk suite: 210 tests passed
- Wider source suite: 458 tests passed; 22 fork suites require `MAINNET_RPC_URL`
- Pre-commit Biome and address lint hooks


Foulks-Plb:
### 🔸 Point to fix (devex / maintainability — §1)

In the `blue.ts` entity, the `repay` (l.896-911) and `repayWithdrawCollateral` (l.1123-1138) methods gain a **new validation block** that **duplicates** checks already performed ~30 lines below in the existing (still present) resolution block:

| New block (added) | Resolution block (existing, unchanged) |
|---|---|
| `params.shares <= 0n` → `NonPositiveInputError` (l.897) | `shares <= 0n` → `NonPositiveInputError` (l.936, 1173) |
| `amount < 0n` → `NegativeInputError` (l.902) | `amount < 0n` → `NegativeInputError` (l.965, 1197) |
| `amount + nativeAmount <= 0n` → `NonPositiveInputError` (l.905) | `repayAssets <= 0n` → `NonPositiveInputError` (l.969, 1201) |

The checks in the second block become **unreachable** for these exact conditions. This is duplicated logic that has to be maintained in two places (two sources of truth for the same rule), which §1 (one responsibility, no dead path) forbids — and I confirmed via `origin/main` that these blocks are **new in this PR** (the base only had the `nativeAmount < 0n` guard at that spot).

**Suggestion:** pick a single place. If the intent is to validate inputs *before* the state reads (`MissingAccrualPositionError` / `validateAccrualPosition`) — which is a good goal — then remove the now-dead checks in the resolution block (l.936-938, 965-966, 969-970 for `repay`; l.1173-1175, 1197-1199, 1201-1203 for `repayWithdrawCollateral`). Otherwise, drop the new block.

Rubilmax:
> **Suggestion:** pick a single place. If the intent is to validate inputs _before_ the state reads (`MissingAccrualPositionError` / `validateAccrualPosition`) — which is a good goal — then remove the now-dead checks in the resolution block (l.936-938, 965-966, 969-970 for `repay`; l.1173-1175, 1197-1199, 1201-1203 for `repayWithdrawCollateral`). Otherwise, drop the new block.


1. we need it at the lowest level to make sure pure actions inputs are checked (because actions are root-exported)
2. we need it at the highest level to avoid async or costly state reads

We agreed that we prefer keeping both for now

The following can be considered though:
- drop it at the lowest level and remove root exports of pure actions
- is the major bump required if it's only about changing the errors thrown?

Foulks-Plb (dismissed):

Foulks-Plb (dismissed):

Foulks-Plb (approved):

PR #884, opened 2026-07-16
feat(morpho-sdk): allow borrow/withdraw up to lltv

Foulks-Plb:
## Context

Users can no longer increase a borrow while their LTV is still below the LLTV. The builder rejects any resulting position above `LLTV − DEFAULT_LLTV_BUFFER` (0.5%), which removed the ability — present in the app before — to borrow closer to LLTV behind a risk acknowledgment. The product decision is to make the 0.5% offset a UI default, not a hard build-time floor.

## Change

`validatePositionHealth` (borrow) and `validatePositionHealthAfterWithdraw` (withdraw) now validate against the **true LLTV** instead of `LLTV − DEFAULT_LLTV_BUFFER`:

- Only positions that would reach or exceed the LLTV (i.e. immediately liquidatable) are rejected.
- The `+ 1n` share↔asset rounding guard and the strict `>` comparison are kept, so a position is **never** built at or above LLTV.
- `DEFAULT_LLTV_BUFFER` stays exported as a recommended **UI default** (e.g. a "max borrow" button); integrators can let users borrow closer to LLTV behind an explicit acknowledgment.
- `BorrowExceedsSafeLtvError`'s message updated to match.

## Safety

Reviewed against Morpho Blue `_isHealthy`:

- Healthy is inclusive (`maxBorrow >= borrowed`), liquidatable is strict (`borrowed > maxBorrow`). The SDK's `maxSafeBorrowAfter = wMulDown(collateralValue, lltv)` is now bit-identical to the contract's `maxBorrow` (same roundings, scale, lltv).
- With `+ 1n` + strict `>`, the builder rejects iff `borrowed ≥ maxBorrow`, so a built position is always strictly `< maxBorrow` → never liquidatable at build time.
- The contract re-accrues interest and re-checks health atomically at execution, so the SDK can never cause an unhealthy on-chain borrow. The removed buffer guarded **no** protocol invariant — it was purely a UX margin against interest/oracle drift, which now lives as the UI default.

## Tests

`validate.test.ts`: added tests that borrowing/withdrawing into the former buffer band (just under LLTV) now passes, plus a strict-boundary test that reaching LLTV still throws (the `+ 1n` guard). All 49 tests pass; verified the new "up to LLTV" test fails when the buffer is reintroduced.

## Companion

Paired with a vvrm change in `morpho-org/morpho-apps` that keeps the MAX button at the safe offset but allows manual entry up to the SDK ceiling behind the existing acknowledgment checkbox. Until this SDK release is published and pinned there, vvrm's interim ceiling is `LLTV − buffer`; it moves to `LLTV` once this lands.

### DIFF
--- a/packages/morpho-sdk/src/helpers/constant.ts
+++ b/packages/morpho-sdk/src/helpers/constant.ts
@@ -4,7 +4,14 @@ import { type Address, maxUint96 } from "viem";
 /** Maximum slippage tolerance: 10% */
 export const MAX_SLIPPAGE_TOLERANCE = MathLib.WAD / 10n;
 
-/** Default LLTV buffer: 0.5% below LLTV. Prevents instant liquidation on new positions. */
+/**
+ * Recommended UI buffer: 0.5% below LLTV. A safe default for a "max borrow"
+ * button so fresh positions don't sit at the liquidation edge. This is NOT
+ * enforced by the transaction builders — `validatePositionHealth` /
+ * `validatePositionHealthAfterWithdraw` only reject positions at or above the
+ * true LLTV — so integrators may let users borrow past this buffer (up to just
+ * under LLTV) behind an explicit risk acknowledgment.
+ */
 export const DEFAULT_LLTV_BUFFER = MathLib.WAD / 200n;
 
 /** Maximum absolute share price cap (100 RAY). Prevents absurd maxSharePrice values in repay. */
--- a/.changeset/borrow-withdraw-up-to-lltv.md
+++ b/.changeset/borrow-withdraw-up-to-lltv.md
@@ -1,0 +1,9 @@
+---
+"@morpho-org/morpho-sdk": minor
+---
+
+Allow borrowing and withdrawing collateral up to (just under) the market LLTV.
+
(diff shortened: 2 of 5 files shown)


Foulks-Plb:
@codex review

packages/morpho-sdk/src/helpers/constant.ts
```diff
@@ -4,7 +4,14 @@ import { type Address, maxUint96 } from "viem";
...
-/** Default LLTV buffer: 0.5% below LLTV. Prevents instant liquidation on new positions. */
+/**
+ * Recommended UI buffer: 0.5% below LLTV. A safe default for a "max borrow"
+ * button so fresh positions don't sit at the liquidation edge. This is NOT
+ * enforced by the transaction builders — `validatePositionHealth` /
+ * `validatePositionHealthAfterWithdraw` only reject positions at or above the
+ * true LLTV — so integrators may let users borrow past this buffer (up to just
+ * under LLTV) behind an explicit risk acknowledgment.
+ */
 export const DEFAULT_LLTV_BUFFER = MathLib.WAD / 200n;
```
Rubilmax, line 15: i would deprecate it because it can bring confusion on whether it is actually implemented or not

PR #886, opened 2026-07-17
fix(morpho-sdk): forward-accrue assets-mode repay maxSharePrice

Foulks-Plb:
## Summary

Fixes an assets-mode repay reverting on quiet markets (VAU-1206): a user could not repay on app.morpho.org (`execution reverted`) on the sUSDai/USDC Arbitrum market, while the fallback app succeeded.

**Root cause.** The bundler path (`Bundler3`/`GeneralAdapter1.morphoRepay`) enforces a `maxSharePrice` slippage bound; core `repay` (used by the fallback app) has none — hence "fallback OK, app KO". In `MorphoBlue.repay` / `repayWithdrawCollateral`, the **assets** path computed `maxSharePrice` from the un-accrued `positionData.market` snapshot (as of the on-chain `lastUpdate`), while the **shares** path already forward-accrued the market by 2h. On-chain, `morphoRepay` accrues interest `lastUpdate → execution` *before* enforcing the bound, so on a market not accrued recently the borrow share price rose past the default `0.03%` `slippageTolerance` ceiling and the bundle reverted. At ~8% APR, 0.03% is only ~1.3 days of accrual → a quiet market reverts systematically.

**Fix.** Forward-accrue the market for **both** repay modes before deriving `maxSharePrice`, mirroring the shares path. Scope is deliberately limited to the repay `max` bounds — the `min` bounds (borrow/withdraw) are unaffected because accrual moves the price *away* from their floor.

## Changes

- `entities/blue/blue.ts` — hoist the 2h forward-accrual out of the shares-only branch in `repay` and `repayWithdrawCollateral` so assets mode uses the accrued market too. `erc20Amount` / `transferAmount` and the post-repay health check are unchanged (the latter already used `accrualTimestamp`); only `maxSharePrice` widens.
- `entities/blue/blue.test.ts` — 2 hermetic regression tests (deterministic clock, market with non-zero `rateAtTarget` + stale `lastUpdate`) asserting the built `maxSharePrice` equals the forward-accrued bound and is strictly looser than the pre-fix (un-accrued) bound.
- changeset: `patch`.

## Test plan

- [x] `tsc --noEmit` (morpho-sdk): clean
- [x] `biome check` on changed files: clean
- [x] Regression tests pass; confirmed non-vacuous (reverting the fix fails them: stale `1000300…` ≠ accrued `1001102…`, ~0.08% > 0.03% ceiling)
- [ ] Full morpho-sdk fork suite not run locally — needs an archive RPC at the fork block (public endpoints 403 on archive). New tests are fork-independent.

## Follow-up

- After release, bump `@morpho-org/morpho-sdk` in `morpho-org/morpho-apps` `apps/vvrm-app` (currently pinned `5.2.0`) to pick up the fix.

### DIFF
--- a/.changeset/repay-assets-forward-accrual.md
+++ b/.changeset/repay-assets-forward-accrual.md
@@ -1,0 +1,17 @@
+---
+"@morpho-org/morpho-sdk": patch
+---
+
+Fix assets-mode repay reverting on quiet markets. `MorphoBlue.repay` and
+`MorphoBlue.repayWithdrawCollateral` now derive `maxSharePrice` from the
+2h-forward-accrued market in **both** repay modes, not just shares mode.
+
+Previously the assets path computed the slippage bound from the un-accrued
+`positionData.market` snapshot (as of the market's on-chain `lastUpdate`),
+while on-chain `morphoRepay` accrues interest `lastUpdate → execution` before
+enforcing the bound. On a market that hasn't accrued recently, the accrued
+borrow share price rose past the default 0.03% `slippageTolerance` ceiling and
+the bundle reverted — e.g. at ~8% APR, 0.03% is only ~1.3 days of accrual, so a
+quiet market reverted systematically. Direct core `repay` (no bundler, no
+bound) was unaffected, which is why the fallback path succeeded while
+app.morpho.org did not.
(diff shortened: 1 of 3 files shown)


Foulks-Plb:
@codex review

0xbulma (approved):

PR #893, opened 2026-07-21
fix(morpho-sdk): forward-accrue supply maxSharePrice

Foulks-Plb:
## Motivation

`MorphoBlue.supply` reverts on-chain on the `morphoSupply` `maxSharePrice` guard for quiet markets. It derived the bound from the un-accrued `marketData` snapshot (as of the market's `lastUpdate`), while on-chain `morphoSupply` accrues interest `lastUpdate → execution` first — which raises the supply share price. Once accrued interest exceeds the default 0.03% `slippageTolerance` (only a few days at ~10% APR), the accrued price exceeds the stale bound and the bundle reverts. This is the supply counterpart of the repay fix (VAU-1206), which had been missed; the native-ETH-only path just surfaced it first.

## Solution

`supply()` now forward-accrues `marketData` to `max(now, lastUpdate) + 2h` before computing `maxSharePrice`, mirroring `repay`/`repayWithdrawCollateral`. Added a hermetic regression test (5-day-stale WETH market, native supply) asserting the bound is derived from the accrued market and is `>=` the on-chain accrued share price, plus a patch changeset. `withdraw`/`borrow` use a lower `minSharePrice` bound that accrual only relaxes, so they were unaffected and are left unchanged.

### DIFF
--- a/.changeset/supply-forward-accrual.md
+++ b/.changeset/supply-forward-accrual.md
@@ -1,0 +1,18 @@
+---
+"@morpho-org/morpho-sdk": patch
+---
+
+Fix loan-asset supply reverting on quiet markets. `MorphoBlue.supply` now
+derives `maxSharePrice` from the 2h-forward-accrued market instead of the raw
+`marketData` snapshot.
+
+Previously the supply path computed the slippage bound from the un-accrued
+`marketData` snapshot (as of the market's on-chain `lastUpdate`), while on-chain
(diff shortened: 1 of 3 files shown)


Foulks-Plb:
@codex review

remiroyc (approved):

0xbulma (approved):

PR #909, opened 2026-07-27
docs: add TIB for VaultExitBundlesV1 in-kind redemption

Foulks-Plb:
## Motivation

A depositor in an illiquid Morpho vault is stuck: `withdraw` and `redeem` revert for want of liquidity, and `MorphoVaultV2.forceWithdraw` / `forceRedeem` still need someone to free real assets — which a fully-utilized market cannot do. Getting out depends on the curator reallocating. [`VaultExitBundlesV1`](https://github.com/morpho-org/bundles/blob/main/src/vault-exit/VaultExitBundlesV1.sol) removes that dependency by transferring the vault's Blue exposure to the user **in-kind** instead of liquidating it, so no new liquidity has to exist for the exit to succeed. `morpho-sdk` has no support for `morpho-org/bundles` today.

## Solution

A TIB (`Proposed`) freezing the design for the two **in-kind redemption** entry points — V1 and V2. `vaultExitBundlesV1ForceWithdrawVaultV2` is explicitly out of scope and gets its own TIB; `forceWithdraw` / `forceRedeem` are untouched and undeprecated.

### Vault V1 IKR

Capture d’écran 2026-07-22 à 19 47 09

### Vault V2 IKR

Capture d’écran 2026-07-22 à 19 46 51

The design is a thin 1:1 binding (assets-only, caller-supplied market list) wrapped in exhaustive pre-flight validation. The load-bearing section is a **validation matrix** derived by walking `VaultExitBundlesV1.sol`, `TokenLib.sol`, and `VaultV2.sol` line by line — every `require`, every unchecked array index, and every nested call in `onMorphoSupply` / `onMorphoFlashLoan` appears in exactly one row, split into synchronous checks (pure, from already-fetched vault data, zero extra RPC) and asynchronous ones batched into a single multicall.

Four findings drove it:

- **Vault V2's EIP-712 domain is non-standard** — `EIP712Domain(uint256 chainId,address verifyingContract)`, no `name`, no `version`. Our `getPermitTypedData` emits four fields and hashes the wrong domain, and Vault V2 has no `eip712Domain()` for `fetchToken` to fall back on. Needs a dedicated `getVaultV2PermitTypedData`. The same bug is latent today in `getGeneralAdapterRequirements({ address: <a Vault V2>, useSimplePermit: true })`.
- **The permit must be `maxUint256`** — allowance is consumed twice per market (main withdraw plus the penalty withdraw inside `forceDeallocate`), each with its own rounding and an accrual in between, so no caller can size it. `exit()` skips the decrement at max, making the approval permanent; documented in Security as non-exploitable, since every entry point spends only `allowance[msg.sender][bundler]`.
- **The redemption loop is unbounded** — `for (uint256 i; assetsToDeallocate > 0; i++)` with no bound on `i`, so an under-supplied list dies with a raw `panic 0x32`. Coverage must be summed over **id-deduplicated** markets, since `expectedSupplyAssets` is re-read each iteration and a repeat contributes ~0; after dedup, sum is equivalent to the greedy loop, which is what makes leaving ordering to the caller safe.
- **Blue's token balance is a live constraint** — Blue credits the supply before the callback but pulls tokens after, so it must physically hold the chunk mid-callback. A 100%-utilized market is exitable in-kind only if Blue holds that loan token elsewhere.

### DIFF
--- a/docs/tibs/TIB-2026-07-27-vault-exit-in-kind-redemption.md
+++ b/docs/tibs/TIB-2026-07-27-vault-exit-in-kind-redemption.md
@@ -55,9 +55,11 @@ the pre-flight validation that turns those opaque failures into named SDK errors
 
 **Goals**
 
-- Add `vaultExitBundlesV1InKindRedemptionVaultV1` and `vaultExitBundlesV1InKindRedemptionVaultV2` to
-  `morpho-sdk` as pure, synchronous actions plus lazy entity handles, following the existing
-  Client → Entity → Action layering.
+- Add `vaultV1InKindRedemption` and `vaultV2InKindRedemption` to `morpho-sdk` as pure, synchronous
+  actions plus lazy entity handles — thin wrappers over the contract's
+  `vaultExitBundlesV1InKindRedemptionVaultV1` / `...VaultV2` entry points — following the existing
+  Client → Entity → Action layering. These short `vaultV{1,2}InKindRedemption` names are the frozen
+  **SDK** surface used everywhere below; the long names are the **contract** functions they encode.
 - **Validate exhaustively before building.** Every revert reachable from a well-formed call must
   surface as a named, exported error class at build time — in particular the unbounded-loop panic,
   the Blue token-balance shortfall, the Vault V2 gates, and allowance sufficiency. Share sufficiency
@@ -119,7 +121,11 @@ The amount of Blue exposure the user receives differs by vault version: for **Va
 full `exitAssets`; for **Vault V2** it is `exitAssets` *net of the force-deallocation penalty* —
 `floor(exitAssets·WAD / (WAD + penalty))` — because the penalty is charged in the same step that
 burns the vault shares. Quoting the V2 output as the full `exitAssets` overstates the proceeds by the
-penalty and would propagate into UI quotes and accounting assertions; the SDK reports the net figure.
+penalty. The SDK does **not** surface this net figure in the action's args or output: `amount` stays
+penalty-*inclusive* (it maps one-to-one to the contract's `exitAssets`), and the net proceeds are the
+caller's to derive from the `penalty` already on the `getData()` snapshot via the formula above. A
+first-class net-proceeds surface belongs to the deferred preview helper (see Future Considerations),
+not to this action's args — adding it here would widen the very surface this TIB is keeping minimal.
 They get there differently.
 
 **Vault V2** — per market in the caller's list, the bundler supplies into Blue *on behalf of the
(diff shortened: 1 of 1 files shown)


Foulks-Plb:
@codex review

Foulks-Plb:
@codex review

Foulks-Plb:
**Integrated:**

- **Exercise the V2 permit path on a fork (P1).** Right, and it was the most valuable catch: the two-field EIP-712 domain is the whole risk of this PR, and the plan let it slip through. The fork section now *requires* at least one `supportSignature: true` V2 case that signs the returned `PermitRequirementSignature`, feeds the `v/r/s` into the bundle, and executes — so `getVaultV2PermitTypedData`'s domain is checked against the live `VaultV2.DOMAIN_SEPARATOR()` by the contract's own `ecrecover`. Added a matching security invariant, and spelled out that a unit round-trip (self-consistent) and an approve-based exit (empty-permit sentinel) both leave the domain unexercised.

- **Do not deduplicate Vault V1 coverage (P2).** Correct, and it contradicted the TIB's own reasoning. V2 drains the adapter position mid-loop (`forceDeallocate`+`withdraw` in `onMorphoSupply`), so a duplicate contributes ≈0 the second time → dedup is right. V1 runs the single `withdraw` *after* the loop, so `expectedSupplyAssets(vault, id)` reads the same undrained position every iteration → a duplicate legitimately re-contributes. V1 coverage now sums the **raw ordered list**; dedup is scoped to V2, the subtlety note is split V1/V2, and the invariant now asserts the mirror V1 list is *accepted* where the V2 one is rejected.

- **Chain-ID guard in the validation matrix (P1).** Confirmed the existing Vault V1/V2 entities all guard `client.chain?.id !== this.chainId` at entry. Added `ChainIdMismatchError` as the **first** synchronous row (ahead of the vault-address row) plus its invariant test.

- **Read the permit spender from action metadata (P2).** Confirmed `PermitArgs` is `{ owner, nonce, asset, signature, amount, deadline }` — no `spender`. The check is now `action.args.spender === vaultExitBundlesV1` (owner/asset/amount/deadline stay on `args`), with a note that `args.spender` would compare `undefined` and reject every valid ERC-2612 signature.

- **One action name throughout (P2).** Unified: `vaultV1InKindRedemption` / `vaultV2InKindRedemption` are the frozen SDK names used everywhere; the long `vaultExitBundlesV1InKindRedemptionVaultV{1,2}` are now labelled as the contract entry points they encode.

- **Express share sufficiency in asset terms (P2).** The delegated caller rule was dimensionally wrong. It now reads `amount ≤ vault.previewRedeem(sharesHeld)` (asset terms), notes the over-cap after appreciation / under-cap after bad debt, and that the V2 penalty leg burns extra shares the caller's budget must cover.

**Resolved differently — Expose the promised net V2 proceeds (P2):** real contradiction, resolved by *removing the promise* rather than adding the surface. `amount` stays penalty-inclusive; the TIB no longer claims the SDK "reports the net figure" — the net is caller-derived from the snapshot `penalty`, and a first-class net-proceeds surface remains the deferred preview helper. Keeping the frozen action args minimal is deliberate here.

**Acknowledged, deferred to implementation (not folded into this TIB pass):**

- **Exact V2 max-exit amount (P3).** Agreed — `mulDivUp(covered + 1, WAD + penalty, WAD) - 1` is the exact bound; the current `mulDivDown(covered, …)` under-reports by up to 1 wei in the error payload. Trivial, will fix in the coverage-error implementation.
- **Canonical adapter key before reading the penalty (P2).** Agreed — resolve the snapshot's canonical adapter address and key `forceDeallocatePenalties` / position lookups off it, so checksum-casing can't turn a valid `adapter` into an `undefined` penalty. Implementation-detail; will handle in the entity.
- **Verify deployed runtime bytecode before enabling addresses (P1).** Agreed the selectors+immutables gate is insufficient; a normalized runtime-bytecode match (or a fork re-run against the deployed contract) is the right bar before flipping an address slot on. This is a deployment-time process step captured for the rollout, not a TIB-frozen decision.

docs/tibs/TIB-2026-07-27-vault-exit-in-kind-redemption.md
```diff
@@ -0,0 +1,755 @@
...
+guessable from the revert data.
+
+This TIB freezes the decision for integrating the two **in-kind redemption** entry points, and for
+the pre-flight validation that turns those opaque failures into named SDK errors.
+
+## Goals / Non-Goals
+
+**Goals**
+
+- Add `vaultV1InKindRedemption` and `vaultV2InKindRedemption` to `morpho-sdk` as pure, synchronous
```
Rubilmax, line 58:
```suggestion
- Add `vaultV1InKindRedeem` and `vaultV2InKindRedeem` to `morpho-sdk` as pure, synchronous
```

Rubilmax (approved):

PR #912, opened 2026-07-29
docs: add TIB for Vault V2 public-allocator shared liquidity

Foulks-Plb:
## Motivation

The SDK can answer *"how much more can be borrowed on this Blue market, counting liquidity a public allocator could pull in?"* for **MetaMorpho Vault V1 only** — `ReallocationData` → `computeReallocations` → `VaultReallocation[]`. `morpho-org/vault-v2` has since shipped [`BluePublicAllocator`](https://github.com/morpho-org/vault-v2/blob/main/src/periphery/blue-public-allocator/BluePublicAllocator.sol) (last touched `b417825`, 2026-07-29), for which the SDK has no ABI, address key, entity, or compute — so as Vault V2 TVL grows, the V1-only engine silently under-reports a market's reallocatable liquidity and quotes borrows as impossible when they would succeed.

## Solution

This PR is **documentation only** — one new TIB, no package source touched, no changeset (root `AGENTS.md` §7 excludes non-API documentation-only changes). It freezes the design of a Vault V2 mirror that is strictly additive: the V1 surface is untouched (no renames, no deprecations, no breaking change), the new symbols mirror V1 names with a `VaultV2` **suffix** (`computeReallocationsVaultV2`, `ReallocationDataVaultV2`, …), and the engine is deliberately **not** wired into any transaction flow — no action, no bundler encoder, and `MorphoBlue.getReallocationData()` keeps its exact current behavior, with the v1/v2 dispatch deferred to the blue bundles migration.

The load-bearing part is a contract-level finding verified line-by-line against `VaultV2.sol` and `MorphoMarketV1AdapterV2.sol`: the vault **rebases** `allocation[id]` to `expectedSupplyAssets` rather than incrementing it by `assets`, which means shared allocation ids (`adapterId`, `collateralId`) do **not** net to zero across a reallocation and `allocation[id]` never bounds the deallocate leg. The intuitive port of the V1 algorithm is therefore wrong in the unsafe direction, and the TIB records it as rejected Alternative 4 with an invariant test per trap.

## Review notes

Six locked decisions worth a second opinion, all argued in **Considered Alternatives**: the sibling-class-over-parameterized-`ReallocationData` split, the vault-grouped result shape, full-parity cap simulation, the shared-id correction above, deferring the encoder, and dropping V1's four deprecated utilization options. Two **Open Questions** are left for reviewers: whether `BluePublicAllocator` gets a deterministic cross-chain address, and whether the liquidity metric should count vault idle by default.

### DIFF
--- a/docs/tibs/TIB-2026-07-29-vault-v2-public-allocator-shared-liquidity.md
+++ b/docs/tibs/TIB-2026-07-29-vault-v2-public-allocator-shared-liquidity.md
@@ -554,6 +554,15 @@ parameter is the end state V1 is heading toward; V2 starts there.
 - Should `getPublicReallocationLiquidityVaultV2` count idle by default? It is included here, on the
   grounds that idle is genuinely reallocatable and excluding it under-reports — but a dashboard
   showing "shared liquidity" may prefer to attribute idle separately.
+- Should reallocation candidates be favored by penalty cost? Unlike PAV1's single per-vault fee,
+  PAV2 charges a penalty per market moved, so a plan spread across many markets is proportionally
+  more expensive for the borrower. This raises a product question: keep ranking candidates by size
+  (as V1 does today), or rank by liquidity obtained per unit of penalty. @Foulks-Plb has made a
+  first algorithm proposal to answer this and related product needs — the phase structure is
+  unchanged (Phase 1 Friendly, then Phase 2 Aggressive); the only difference is that each
+  reallocation is assigned a cost and the least-expensive reallocations are prioritized within each
+  phase. The idea is currently under discussion with the product team — see the
+  [Slack thread](https://morpholabs.slack.com/archives/C0AJMKR8VB9/p1785422085768909?thread_ts=1785398315.187139&cid=C0AJMKR8VB9).
 
 ## References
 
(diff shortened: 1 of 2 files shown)


docs/tibs/TIB-2026-07-29-vault-v2-public-allocator-shared-liquidity.md
```diff
@@ -554,6 +554,15 @@ parameter is the end state V1 is heading toward; V2 starts there.
 - Should `getPublicReallocationLiquidityVaultV2` count idle by default? It is included here, on the
   grounds that idle is genuinely reallocatable and excluding it under-reports — but a dashboard
   showing "shared liquidity" may prefer to attribute idle separately.
+- Should reallocation candidates be favored by penalty cost? Unlike PAV1's single per-vault fee,
```
Foulks-Plb, line 551: Open question before start (Should reallocation candidates be favored by penalty cost?)
(https://morpholabs.slack.com/archives/C0AJMKR8VB9/p1785398315187139)

Rubilmax (approved):

jinmel (approved):

PR #841, opened 2026-07-02
feat(morpho-ts): add Robinhood Chain (4663) support

Foulks-Plb:
## Summary

Adds **Robinhood Chain** (chain id `4663`) to the shared chain and address registries in `@morpho-org/morpho-ts`, so all maintained SDK packages resolve Morpho on Robinhood natively.

This ports the integration we shipped in an ad-hoc/runtime-injected form in `morpho-apps` (vvrm) into the SDK proper. Addresses are taken verbatim from the canonical source of truth — `morpho-org/deployments` `address-book.json → chains["4663"]` (era 2) — and cross-checked against the published protocol docs.

## Changes

- **`chain.ts`** — add `ChainId.RobinhoodMainnet = 4663` and its `CHAIN_METADATA` entry (name `Robinhood Chain`, native `ETH`, explorer `https://robinhoodchain.blockscout.com`, identifier `robinhood`).
- **`addresses.ts`** — register the era-2 address set and deployment blocks, plus the wrapped-native → native unwrap mapping.
- **Tests** — extend the era-2 `test.each` cases in `chain.test.ts` and `addresses.test.ts`.
- **Changeset** — `morpho-ts` minor (new `ChainId` member + registry entries); patch the maintained direct dependents (`blue-sdk-viem`, `evm-simulation`, `morpho-sdk`) so their latest releases resolve the new entry.

## Addresses (source: `morpho-org/deployments` address-book, era 2)

| Contract | Address |
|---|---|
| Morpho (Blue) | `0x9D53d5E3bd5E8d4Cbfa6DB1ca238AEA02E651010` |
| AdaptiveCurveIrm | `0x2BD3d5965B26B51814AC95127B2b80dD6CcC0fa1` |
| Bundler3 | `0x6478e9393d4C5bB4d53ee881d1DE78786A0344a6` |
| GeneralAdapter1 | `0xc5E188541D107e8B79e43478bDE365F1406665D6` |
| VaultV2Factory | `0x0FBad98595b0186dA120E41f77C102beb49f803c` |
| MorphoMarketV1AdapterV2Factory | `0x79370Ed003CE325C088E530d5e8655c99c2993e1` |
| MorphoVaultV1AdapterFactory | `0x7a91222F3f7B927bB8fb624593Ca86e111C2F85e` |
| RegistryList | `0xe785a2eFD384BA7B95BaEd3851BC76aeD67C676f` |
[...]

PR #915, opened 2026-08-04
feat(morpho-sdk): add vault in-kind redemption

Rubilmax:
## Summary

- add pure `vaultV1InKindRedeem` and `vaultV2InKindRedeem` transaction actions backed by `VaultExitBundlesV1`
- add lazy Vault V1/V2 entity flows with typed preflight validation, max-share permit/approval requirements, and Vault V2's two-field permit domain
- register the standalone bundle address slot and publish its ABI
- add unit, property, mock-client, and mainnet-fork coverage, plus SDK docs and a changeset

## Why

Vault exits can require withdrawing illiquid Morpho positions in kind. This exposes the bundle flow through the SDK while detecting coverage, gate, balance, chain, vault, permit, and adapter failures before calldata is built.

This PR is stacked on #909.

## Validation

- `pnpm lint`
- `pnpm test` — 220 files, 2,744 tests passed
- `pnpm --filter @morpho-org/morpho-sdk exec tsc --noEmit`
- focused permit encoder tests
- Vault V2 mainnet-fork permit execution against the contract's `DOMAIN_SEPARATOR()`


packages/morpho-sdk/src/actions/signatures/getVaultExitBundlesV1PermitStruct.ts
```diff
@@ -0,0 +1,138 @@
...
+      actual: requirementSignature.args.asset,
+    });
+  }
+  if (requirementSignature.args.amount !== maxUint256) {
+    throw new InKindRedeemPermitMismatchError({
+      field: "amount",
+      expected: String(maxUint256),
+      actual: String(requirementSignature.args.amount),
+    });
+  }
```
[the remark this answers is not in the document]
Rubilmax: Thanks — this is the same behavior as the existing permit-based action flows, rather than something specific to VaultExitBundlesV1. Their buildTx paths also select a permit by kind, and getTokenRequirementActions validates the signed asset and amount but not the owner or the requirement action spender.\n\nThe intended contract is that signatures passed to buildTx come from the same ActionOutput.getRequirements() flow. Requirement construction and signing bind the owner and spender, while the token verifies that binding onchain. Mixing a RequirementSignature from another flow can therefore cause a revert, but cannot authorize a different execution.\n\nAdding owner and spender provenance checks only here would make this flow inconsistent. If we want buildTx to defend against cross-flow signature mixing, that should be addressed centrally across all permit consumers, so I suggest leaving this unchanged in this PR.

Rubilmax (commented):
almost lgtm, pending deployment

packages/morpho-sdk/src/actions/vaultV2/inKindRedeem.ts
```diff
@@ -0,0 +1,124 @@
...
+  Transaction<VaultV2InKindRedeemAction>
+> => {
+  if (args.amount <= 0n) throw new NonPositiveInputError("amount", args.amount);
+  if (args.deadline <= 0n)
+    throw new NonPositiveInputError("deadline", args.deadline);
+  if (args.marketParamsList.length === 0)
+    throw new EmptyMarketParamsListError();
+
+  const to = getChainAddress(vault.chainId, "bundles.vaultExitBundlesV1");
+  const marketParamsList = args.marketParamsList.map((marketParams) => ({
```
jinmel, line 76: is this redundant? args.marketParamsList could it be just a copy of args.marketParmasList?

packages/morpho-sdk/src/actions/vaultV1/inKindRedeem.ts
```diff
@@ -0,0 +1,120 @@
...
+  Transaction<VaultV1InKindRedeemAction>
+> => {
+  if (args.amount <= 0n) throw new NonPositiveInputError("amount", args.amount);
+  if (args.deadline <= 0n)
+    throw new NonPositiveInputError("deadline", args.deadline);
+  if (args.marketParamsList.length === 0)
+    throw new EmptyMarketParamsListError();
+
+  const to = getChainAddress(vault.chainId, "bundles.vaultExitBundlesV1");
+  const marketParamsList = args.marketParamsList.map((marketParams) => ({
```
jinmel, line 78: is this redundant? this is practically just a copy of args.marketParamsList

packages/morpho-sdk/src/actions/requirements/encode/encodeVaultSharesPermit.ts
```diff
@@ -0,0 +1,168 @@
+import { type Address, Eip5267Domain, Token } from "@morpho-org/blue-sdk";
```
jinmel, line 1: i thought shares are just erc20, why do we need another Permit? can we reuse erc20 permit?

jinmel (dismissed):

packages/morpho-sdk/src/actions/vaultV2/inKindRedeem.ts
jinmel: Resolving — not redundant, but the reason is runtime-level, not type-level. `InputMarketParams` is structural, so callers routinely pass full `MarketParams` class instances (which are globally cached in `MarketParams._CACHE`). This action embeds the list in its deep-frozen `action.args`, so without the field-pick copy, `deepFreeze` would freeze caller-owned, process-wide-shared cached instances. The copy gives the freeze a fresh SDK-owned POJO with exactly the 5 tuple fields. Existing blue actions never hit this because they embed only `marketParams.id`, not the params objects.

packages/morpho-sdk/src/actions/vaultV1/inKindRedeem.ts
jinmel: Resolving — same as the V2 thread: the field-pick copy exists so `deepFreeze` never propagates into caller-owned (and globally cached) `MarketParams` class instances that structurally satisfy `InputMarketParams`. See the discussion on `vaultV2/inKindRedeem.ts` for the full reasoning.

packages/morpho-sdk/src/actions/requirements/encode/encodeVaultSharesPermit.ts
jinmel: Resolving — shares are ERC-20 and the ERC-2612 permit *struct* is reused (both encoders delegate to `getPermitTypedData`). What cannot be reused is the EIP-712 *domain* path: Vault V2 hashes a two-field domain (`chainId` + `verifyingContract`, no name/version) and does not implement `eip712Domain()`, so `encodeErc20Permit` → `fetchToken` would silently fall back to a 4-field domain with `version: "1"` and the signature would revert on-chain with `InvalidSigner()`. Plus: different spender allowlist (`vaultExitBundlesV1`), deadline shared with the bundle calldata (vs independent now+2h), pinned `maxUint256` amount, and signer-must-equal-owner binding. Documented in the TIB, Considered Alternatives §3.

Rubilmax: Unfortunately not because of VaultV2 that dont use the typical permit fields for the domain (only chain id and contract, no version for example)

Rubilmax:
@codex

packages/morpho-sdk/src/entities/vaultV1/vaultV1.ts
```diff
@@ -313,6 +401,173 @@ export class MorphoVaultV1 implements VaultV1Actions {
...
+    const deadline = deadlineOverride ?? now + Time.s.from.h(2n);
+    if (deadline <= now) throw new ExpiredDeadlineError(deadline, now);
+
+    const vaultExitBundlesV1 = getChainAddress(
+      this.chainId,
+      "bundles.vaultExitBundlesV1",
+    );
+    const addresses = getChainAddresses(this.chainId);
+    const blue = addresses.blue ?? addresses.morpho;
+    const covered = marketParamsListSnapshot.reduce((total, marketParams) => {
```
tarikbellamine: This over-counts a repeated market id and accepts exits that cannot execute.

The raw ordered sum is the correct bound for what it's checking, loop termination. The contract really does re-read the undrained `expectedSupplyAssets(BLUE, marketParams, vault)` each iteration, so a duplicate genuinely subtracts twice and `panic 0x32` is avoided. That part of the TIB is accurate.

But the loop supplies flash-loaned assets on behalf of `sender`, not the vault, so the vault's own position never grows. `MetaMorpho._withdrawable` (MetaMorpho.sol:862-874) returns `min(supplyAssets, availableLiquidity)` where `supplyAssets` is the vault's own position. Supplying 2P into market A cannot let the vault return more than P from A, so the trailing `IMetaMorpho(vault).withdraw(exitAssets, ...)` reverts `NotEnoughLiquidity()`.

Concretely, using this PR's own V1 fixture (position 1000, `amount: 1_500n`, list `[m, m]`): the loop supplies 1000 then 500 into the same market, `Σ assets_i` does equal `exitAssets`, and the vault still returns only 1000 against a 1500 withdraw.

Duplicates never enable a V1 exit. They only burn flash-loaned assets and guarantee failure, after the user has authorized a `maxUint256` share permit.

Plain dedup is not the fix, it would break the panic bound and reject lists that don't panic. Suggest keeping the raw ordered greedy and additionally tracking assets assigned per market id, rejecting with `InKindRedemptionCoverageError` when a market is drawn from twice with a non-zero chunk.

packages/morpho-sdk/src/entities/vaultV1/vaultV1.inKindRedeem.test.ts
```diff
@@ -0,0 +1,319 @@
...
+    encodeReadResult({
+      abi: blueAbi,
+      functionName: "feeRecipient",
+      result: params.blueFeeRecipient ?? IN_KIND_USER,
+    }),
+  ]);
+};
+
+describe("MorphoVaultV1.inKindRedeem", () => {
+  test("default: accepts raw duplicate coverage and builds the V1 action", () => {
```
tarikbellamine: This is the primary happy-path test and it asserts the behaviour flagged at `vaultV1.ts:453`.

`inKindVaultV1Data()` gives the vault a single market position of 1000 assets (`supplyShares: 1_000_000_000n` of `totalSupplyShares: 1_000_000_000n` against `totalSupplyAssets: 1_000n`), and `totalSupply` / `lastTotalAssets` are both `1_000n`. The test then accepts `amount: 1_500n`. That exit exceeds the vault's entire asset base, so it reverts on-chain regardless of how coverage is computed.

Suggest inverting this to `test("error: duplicate markets do not double-count coverage")` expecting `InKindRedemptionCoverageError`, mirroring the V2 case at `vaultV2.inKindRedeem.test.ts:98`, and moving the happy path to a distinct-market list.

docs/tibs/TIB-2026-07-27-vault-exit-in-kind-redemption.md
```diff
@@ -510,45 +509,36 @@ which puts `VaultExitBundlesV1` on the pinned mainnet fork ahead of any live dep
 - **Security invariants as tests** (§5): the permitted value is always `maxUint256`; the empty permit
   is exactly `v=0, r=0, s=0`; a **V2** duplicate-heavy list that only covers the amount when
   double-counted is *rejected* (deduped coverage), while the mirror **V1** list — where duplicates
-  legitimately re-contribute the undrained vault position — is *accepted* (raw-ordered coverage);
-  an under-covering list is rejected rather than left to panic; the **default**
+  legitimately re-contribute the undrained vault position — is *accepted* (raw-ordered coverage); a
```
tarikbellamine: This promotes the V1 duplicate rule to a security invariant, which is the part most likely to be trusted later without re-derivation.

The underlying claim in the "Deduplication is load-bearing on V2, and forbidden on V1" bullet (lines 340-366, which predate this PR) says a duplicated V1 entry "re-contributes its full position, exactly as the contract will", and that deduplicating "would reject exits the contract can execute". The first half is true of the loop. The second is not true of the withdraw, because `MetaMorpho._withdrawable` caps each market at the vault's own position.

Suggest rewriting the V1 half to say the raw ordered list is required only because it is what the unbounded loop indexes, adding a matrix row for per-market over-assignment reverting `NotEnoughLiquidity()`, and flipping this invariant so a duplicate-heavy V1 list is rejected.

packages/morpho-sdk/src/actions/signatures/getVaultExitBundlesV1PermitStruct.ts
```diff
@@ -0,0 +1,138 @@
...
+  if (!isAddressEqual(requirementSignature.args.asset, params.vault)) {
+    throw new InKindRedeemPermitMismatchError({
+      field: "asset",
+      expected: params.vault,
+      actual: requirementSignature.args.asset,
+    });
+  }
+  if (requirementSignature.args.amount !== maxUint256) {
```
tarikbellamine: The binding checks cover `action.type`, `args.asset` and `args.amount`, but not `args.owner` or `action.args.spender`, both of which are on the signature.

`selectRequirementSignatures` buckets `permit` and `permit2` together, so a signature produced for a different spender or by a different account, on the same vault-share token with `maxUint256`, passes every guard here, gets embedded, and reverts in `TokenLib.submitPermit` as a bare `InvalidSigner()` once the user has signed and paid gas. That's the failure class this flow exists to remove.

Worth noting the base TIB described owner and spender as an authorization-binding invariant with its own tests, and this PR rewrites that to "left to the vault's on-chain ERC-2612 verification". On-chain verification does catch it, just in the worst place.

`owner` is available as `args.userAddress` and the spender is already resolved as `to` in both actions. Suggest threading both in and throwing `InKindRedeemPermitMismatchError` with `field: "owner"` / `field: "spender"`. Note spender lives on `action.args`, not `args`.

packages/morpho-sdk/src/entities/vaultV2/vaultV2.ts
```diff
@@ -348,6 +438,215 @@ export class MorphoVaultV2 implements VaultV2Actions {
...
+
+    return {
+      getRequirements: async (): Promise<readonly ActionRequirement[]> => {
+        // Vault gates are intentionally left to the final transaction. In particular, a receive
+        // gate may inspect VaultExitBundlesV1's transient `initiator`, which is populated only
+        // during the actual periphery call. Reading gates directly (inside or outside multicall)
+        // would therefore use a different execution context and could reject a valid exit or
+        // approve one that later reverts. Simulate the finalized transaction after authorization
+        // when gate compatibility must be checked before submission.
+        const [allowance, nonce, blueBalance] = await multicall(
```
tarikbellamine, line 605: The multicall reads allowance, nonce and Blue balance, but the PR also drops the `canSendShares` async row the base TIB had.

The stated reason for dropping gate preflight is that a receive-assets gate may inspect the periphery's transient `initiator`, which is unset during a standalone `eth_call`. That's sound, and it's specific to `canReceiveAssets`. `VaultV2.canSendShares(address)` (VaultV2.sol:926) is a plain account-keyed view with no transient dependency, and the V2 exit hits it on all three legs (VaultV2.sol:806 and 851).

Declining is still defensible, since a custom `sendSharesGate` is an arbitrary contract that could itself read `initiator`. But the TIB's justification only covers half of what was removed. Either add `canSendShares(userAddress)` here, or narrow the TIB paragraph to `canReceiveAssets` and say why the send gate is also skipped.

Adjacent: this multicall re-reads Blue's balance to narrow the drift window but not the vault's idle `assetBalance`, which moves on every deposit or withdraw and feeds straight into `assetsToDeallocate`. It would be free here. The TIB declines re-reads because they "only narrow the window, never close it", which applies equally to the Blue balance read you do perform.

packages/morpho-sdk/src/entities/vaultV2/vaultV2.ts
```diff
@@ -348,6 +438,215 @@ export class MorphoVaultV2 implements VaultV2Actions {
...
+
+    const vaultExitBundlesV1 = getChainAddress(
+      this.chainId,
+      "bundles.vaultExitBundlesV1",
+    );
+    const addresses = getChainAddresses(this.chainId);
+    const blue = addresses.blue ?? addresses.morpho;
+
+    return {
+      getRequirements: async (): Promise<readonly ActionRequirement[]> => {
```
tarikbellamine, line 594: `deadline` is validated once at handle construction (`vaultV2.ts:486`, `vaultV1.ts:445`) and never re-checked here, even though `getRequirements()` re-reads allowance, nonce and balances fresh.

A handle held past the 2h default (a modal left open, a queued job) hands back a permit requirement whose deadline has already passed. The user signs a dead signature and `buildTx()` emits calldata that reverts with `DeadlinePassed()` / `PermitDeadlineExpired()` without any SDK error firing.

Suggest re-running the `ExpiredDeadlineError` check at the top of `getRequirements()` in both entities.

packages/morpho-sdk/src/entities/vaultV2/vaultV2.ts
```diff
@@ -348,6 +438,215 @@ export class MorphoVaultV2 implements VaultV2Actions {
...
+
+    const penalty =
+      vaultData.forceDeallocatePenalties[soleAdapter.address] ?? 0n;
+    const idleAssets = MathLib.min(vaultData.assetBalance, amount);
+    const assetsToDeallocate = MathLib.wDivDown(
+      amount - idleAssets,
+      MathLib.WAD + penalty,
+    );
+    if (idleAssets === 0n && assetsToDeallocate === 0n) {
+      throw new NonPositiveInputError("assetsToDeallocate", assetsToDeallocate);
```
tarikbellamine: Minor: this reports `assetsToDeallocate` as the offending field, but that is a derived internal quantity the integrator never supplied. The message renders as `Input "assetsToDeallocate" must be positive, got "0"`, and `NonPositiveInputError`'s own JSDoc documents `field` as "Public input field whose value is invalid".

Suggest naming the actual public input, or adding a dedicated error saying the requested amount rounds to no deallocation once the penalty is applied and the vault holds no idle assets.

packages/morpho-sdk/src/entities/vaultV2/vaultV2.ts
```diff
@@ -117,6 +134,79 @@ export interface VaultV2Actions {
...
     buildTx: () => Readonly<Transaction<VaultV2RedeemAction>>;
   };
+  /**
+   * Prepares an illiquid Vault V2 exit into idle assets and Morpho Blue supply positions.
+   *
+   * The vault must have exactly one `MorphoMarketV1AdapterV2`. `amount` is penalty-inclusive and
+   * the caller controls market order. Call `getRequirements()` before `buildTx()` so Blue balance,
+   * allowance, and nonce are checked on-chain. Vault gates are enforced by the final transaction
+   * and are not preflighted because receive gates may depend on VaultExitBundlesV1's transient
+   * initiator. The SDK intentionally does not validate the user's share balance; it must cover both
```
tarikbellamine: The TIB's Security section requires the caller's sizing rule on both entity methods, specifically that it is asset-denominated against `amount <= vault.previewRedeem(sharesHeld)` and not a raw share comparison. `vaultV1.ts:136` has that sentence. This one only says "it must cover both the main and penalty legs".

V2 is the case where it matters more, since the penalty leg burns shares on top of `previewWithdraw(amount)`. Suggest adding the `previewRedeem(sharesHeld)` sentence here and keeping the existing penalty note.

packages/morpho-sdk/src/helpers/previewVaultV2InKindRedeem.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/** Inputs used to preview Vault V2 in-kind redemption market choices. */
+export interface PreviewVaultV2InKindRedeemParams {
+  /** Positive penalty-inclusive amount the user wants to exit. */
+  readonly requestedExitAssets: bigint;
+  /** Timestamp used to accrue every Morpho Blue market before converting adapter shares. */
+  readonly timestamp: bigint;
+}
+
+interface VaultV2InKindRedeemMarketPreview {
```
tarikbellamine: `previewVaultV2InKindRedeem` is exported through `helpers/index.ts` and returns `readonly VaultV2InKindRedeemMarketPreview[]`, but this interface has no `export` and isn't in the barrel, while its sibling `PreviewVaultV2InKindRedeemParams` is.

The stated audience is a Vault V2 frontend holding these choices in state, and it can't name the type without `ReturnType<typeof previewVaultV2InKindRedeem>[number]`. Suggest exporting the interface, adding the barrel entry, and giving the seven fields JSDoc so TypeDoc has something to render.

.changeset/register-vault-exit-bundles-address.md
```diff
@@ -0,0 +1,21 @@
+---
+"@morpho-org/morpho-ts": minor
+"@morpho-org/morpho-sdk": minor
```
tarikbellamine, line 4: `@morpho-org/evm-simulation` is the only other package declaring `@morpho-org/morpho-ts` as a direct runtime dependency, and it isn't listed, so it gets an auto-generated "Updated dependencies" line rather than an explicit release.

Prior address-registry PRs handled this explicitly. `packages/evm-simulation/CHANGELOG.md` carries "Patch maintained packages with direct runtime dependencies on `@morpho-org/morpho-ts` so their latest releases resolve the new registry entry" from both #828 and #841, and #841 added it via the follow-up commit "align Robinhood changeset release scope".

Suggest adding `"@morpho-org/evm-simulation": patch` and that sentence to the body.

packages/morpho-sdk/src/entities/vaultV2/vaultV2.inKindRedeem.test.ts
```diff
@@ -0,0 +1,443 @@
...
+      .getRequirements();
+
+    await expect(requirements).rejects.toBeInstanceOf(
+      InsufficientBlueBalanceForInKindRedeemError,
+    );
+  });
+
+  test("behavior: exact max allowance returns no authorization", async () => {
+    const handle = createMockClient(mainnet);
+    mockV2Requirements(handle, { allowance: maxUint256 });
```
tarikbellamine: The allowance branch is only exercised at `0n` and `maxUint256`, so weakening `allowance === maxUint256` to `allowance >= amount` keeps this suite green. The TIB calls out that short-circuiting on a merely large allowance is unsafe, since the exact burn is unknowable, so it is worth a test that fails if the operator changes. Same gap at `vaultV1.inKindRedeem.test.ts:238`.

Separately, every Blue-balance case here and on the fork uses a single market, so `peak` always equals the full `assetsToDeallocate` and nothing distinguishes the largest-chunk bound from a plain sum. A two-market case with `blueBalance` below the sum but at or above the largest chunk would pin that down.

packages/morpho-sdk/src/actions/vaultV1/inKindRedeem.ts
tarikbellamine, line 78: Minor, AGENTS.md §1: the five-field `MarketParams` to `InputMarketParams` projection is byte-identical at four sites (here, `actions/vaultV2/inKindRedeem.ts:76`, `entities/vaultV1/vaultV1.ts:433`, `entities/vaultV2/vaultV2.ts:471`), which is over the three-call-site extraction threshold. The entity/action pair also applies it twice to the same list in one call chain.

Relatedly, `78eeda20` inlined the accrual expression `market.accrueInterest(MathLib.max(now, market.lastUpdate)).toSupplyAssets(shares)` as two-use, but it has three sites (`vaultV2.ts:523`, `vaultV1.ts:459`, `previewVaultV2InKindRedeem.ts:70`), so by the same rule it qualifies to stay extracted and keeps the accrual-timestamp convention in one place.

packages/morpho-sdk/src/actions/signatures/getVaultExitBundlesV1PermitStruct.ts
```diff
@@ -0,0 +1,138 @@
...
+  /** Vault-share allowance authorized by the permit. */
+  readonly value: bigint;
+  /** Vault permit nonce signed by the owner. */
+  readonly nonce: bigint;
+  /** Timestamp after which the permit is invalid. */
+  readonly deadline: bigint;
+  /** ECDSA recovery identifier, or zero for the empty-permit sentinel. */
+  readonly v: number;
+  /** ECDSA signature `r`, or zero for the empty-permit sentinel. */
+  readonly r: `0x${string}`;
```
tarikbellamine: Minor: `r` and `s` inline `` `0x${string}` `` rather than viem's `Hex`. These are the only two occurrences of that inlined literal in `packages/morpho-sdk/src`; every other module imports `type Hex`.

packages/morpho-sdk/src/types/error.ts
```diff
@@ -55,6 +55,208 @@ export class NonPositiveInputError extends Error {
...
+  ) {
+    super(
+      `Vault "${vault}" is Morpho Blue "${blue}"'s fee recipient. Use another exit path because VaultExitBundlesV1 cannot safely account for the vault's accrued fee shares.`,
+    );
+    this.name = "VaultIsBlueFeeRecipientError";
+  }
+}
+
+/** Thrown when a vault-shares requirement cannot be safely encoded as an ERC-2612 permit. */
+export class InKindRedeemPermitMismatchError extends Error {
```
tarikbellamine: Two naming points, both cheap now and major-version breaks after publish (§7).

The new classes split across two prefixes for one feature: `InKindRedemptionRequiresSingleAdapterError` / `InKindRedemptionCoverageError` against `InsufficientBlueBalanceForInKindRedeemError` / `InKindRedeemPermitMismatchError`, while the TIB freezes `InKindRedeem` as the SDK verb.

And `InKindRedeemPermitMismatchError` is thrown from `getVaultExitBundlesV1PermitStruct`, which the deferred `vaultExitBundlesV1ForceWithdrawVaultV2` entry point will reuse as-is. A force-withdraw failing with an error named "InKindRedeem" will be wrong the day that lands. Something entry-point-neutral like `VaultExitBundlesV1PermitMismatchError` would survive it.

Also `field`, `expected` and `actual` are typed as bare `string`, so integrators can't exhaustively switch on the failing field. Narrowing `field` to a literal union would help.

packages/morpho-sdk/src/types/error.ts
```diff
@@ -55,6 +55,208 @@ export class NonPositiveInputError extends Error {
...
+  public constructor(public readonly adapter: Address) {
+    super(
+      `Adapter "${adapter}" does not support Vault V2 in-kind redemption. Use a MorphoMarketV1AdapterV2-backed vault.`,
+    );
+    this.name = "UnsupportedInKindAdapterError";
+  }
+}
+
+/** Thrown when the ordered market list cannot cover the requested in-kind redemption. */
+export class InKindRedemptionCoverageError extends Error {
```
tarikbellamine: Minor: `required` carries different semantics per version with no discriminator. On V1 it is the caller's `amount`; on V2 it is `assetsToDeallocate`, after idle and penalty. The shared message renders as `In-kind redemption requires "${required}" assets`, so a V2 user asking to exit 1000 with 900 idle sees a number they never typed and can't map back to their input.

Suggest a `version` discriminant, or carrying the penalty-inclusive equivalent alongside `required`.

docs/tibs/TIB-2026-07-27-vault-exit-in-kind-redemption.md
```diff
@@ -55,24 +55,25 @@ the pre-flight validation that turns those opaque failures into named SDK errors
...
+  Client → Entity → Action layering. These short `vaultV{1,2}InKindRedeem` names are the frozen
   **SDK** surface used everywhere below; the long names are the **contract** functions they encode.
-- **Validate exhaustively before building.** Every revert reachable from a well-formed call must
-  surface as a named, exported error class at build time — in particular the unbounded-loop panic,
-  the Blue token-balance shortfall, the Vault V2 gates, and allowance sufficiency. Share sufficiency
-  is the one deliberate exception: `amount` sizing against the user's balance is the caller's job
-  (see Non-Goals).
+- **Surface supported pre-flight failures before submission.** Handle construction synchronously
+  validates snapshot-dependent inputs and market coverage. When the caller awaits
+  `getRequirements()`, it performs the RPC-backed Blue balance, Vault V2 gate, allowance, nonce,
```
tarikbellamine: This says `getRequirements()` performs "the RPC-backed Blue balance, Vault V2 gate, allowance, nonce, and Morpho-deployment checks", but the same diff removes the gate rows from the async matrix and adds "Vault V2 gates are intentionally not preflighted" further down. The multicall at `vaultV2.ts:573` reads allowance, nonces and balanceOf only.

Suggest dropping "Vault V2 gate" from this list and pointing at the not-preflighted paragraph.

docs/tibs/TIB-2026-07-27-vault-exit-in-kind-redemption.md
```diff
@@ -88,14 +89,13 @@ the pre-flight validation that turns those opaque failures into named SDK errors
...
   `balanceOf` read it would only duplicate.
-- No composition with a penalty-free `withdraw` leg. In-kind redemption never touches the vault's
-  pre-existing idle assets; splitting an exit across both paths is deferred (see Future
-  Considerations).
+- No separate composition with a penalty-free `withdraw` leg. VaultExitBundlesV1 atomically
+  withdraws available idle assets first and applies in-kind redemption only to the remainder.
 - No new runtime dependencies. `viem` stays the only peer dep of `morpho-sdk`.
 
-## Current Solution
+## Pre-change State
```
tarikbellamine: Minor: this renames the template's `## Current Solution` section to `## Pre-change State`. `docs/tibs/TEMPLATE.md` and the five other TIBs in `docs/tibs/` all use `## Current Solution`, so this breaks greppability across the set.

Rubilmax: Not important, ignored

packages/morpho-sdk/src/types/error.ts
Rubilmax: The error is not for a user anyway it is for the developer

packages/morpho-sdk/src/actions/vaultV1/inKindRedeem.ts
Rubilmax: Yes but each bit of code is so small that extracting it actually complexifies the review for not much benefit considering it can always be one-line inlined

packages/morpho-sdk/src/entities/vaultV2/vaultV2.ts
Rubilmax: only if done similarly in other actions of the morpho-sdk

packages/morpho-sdk/src/actions/signatures/getVaultExitBundlesV1PermitStruct.ts
Rubilmax: we dont do these checks in other actions of the morpho-sdk

jinmel (dismissed):

packages/morpho-sdk/src/actions/requirements/encode/encodeVaultSharesPermit.ts
```diff
@@ -0,0 +1,168 @@
...
+    chainId,
+    spender,
+    allowed: ["vaultExitBundlesV1"],
+  });
+
+  const action: PermitAction = {
+    type: "permit",
+    args: {
+      spender,
+      amount: maxUint256,
```
tarikbellamine: Forcing `maxUint256` has two consequences worth a deliberate answer rather than a default. Both trace back to this line and to the matching rejection in `getVaultExitBundlesV1PermitStruct` (any signed amount other than max throws).

**The residual allowance is reachable through the contract's third entry point.** The TIB argues the standing allowance is not third-party exploitable because every entry point spends `allowance[msg.sender][bundler]`. I verified that and it is true. It rules out the wrong threat model. `vaultExitBundlesV1ForceWithdrawVaultV2` (VaultExitBundlesV1.sol:184) takes a caller-supplied `referralFeePct` bounded only by `require(referralFeePct < WAD)`, a caller-supplied `referralFeeRecipient`, and a caller-supplied `minSharePriceE27`, and finishes with `safeTransfer(asset, referralFeeRecipient, referralFeeAssets)`. Its preconditions (`adaptersLength() == 1` on a `MorphoMarketV1AdapterV2`) are exactly the vault shape this V2 flow targets. So a user who completes an in-kind exit holds a permanent unbounded allowance to a contract that will, on one transaction they sign, burn their shares and forward up to 99.99% of the proceeds to an arbitrary address with slippage disabled. That is phishing amplification rather than a direct exploit, but the standing allowance is precisely what removes the approval step and its unlimited-approval warning, leaving a single confirmation on a call to an audited Morpho contract. The TIB mentions "the general exposure of a standing approval" but does not name the entry point that makes this particular one unusual.

**It also discards the flow's only slippage control.** The contract explicitly does not check share price (natspec L59, L118) and twice designates the allowance as the substitute: *"The allowance/permit of shares can also be used to bound the max burned shares"* (L51, L106). The TIB's reason for declining is that the allowance is consumed twice per market with rounding and interest between, so no caller can size it. That holds on V2, but it is false on V1, which has exactly one `IMetaMorpho.withdraw` (L100) spending one computable `previewWithdraw(amount)`. And "not exactly sizeable" does not imply "must be unbounded" — a padded bound is computable on both, and `TokenLib.submitPermit` passes `permit.value` straight through, so the contract imposes no max requirement. Every other vault action in this SDK carries a `slippageTolerance`; this one, most likely to be used during a liquidation cascade, carries none.

Suggest either taking an explicit bounded value (the owner's share balance, or `toShares(amount)` grossed up by `(WAD + penalty)` with headroom) and relaxing the max-equality check to accept anything at or above the computed bound, or, if max stays the default, documenting the ForceWithdraw referral-fee exposure on both entity methods and emitting a revocation `Transaction` alongside the exit so integrators can zero the allowance afterwards.

packages/morpho-sdk/src/entities/vaultV2/vaultV2.ts
```diff
@@ -348,6 +446,223 @@ export class MorphoVaultV2 implements VaultV2Actions {
...
+        if (deadline <= requirementsTimestamp) {
+          throw new ExpiredDeadlineError(deadline, requirementsTimestamp);
+        }
+        // Vault gates are intentionally left to the final transaction. A receive gate may inspect
+        // VaultExitBundlesV1's transient `initiator`, which is populated only during the actual
+        // periphery call. A send-share gate is arbitrary external code and is evaluated repeatedly
+        // after intermediate penalty share burns, so one standalone read is not execution-equivalent
+        // either. Simulate the finalized transaction after authorization when gate compatibility
+        // must be checked before submission.
+        const [allowance, nonce, blueBalance] = await multicall(
```
tarikbellamine, line 605: This multicall re-reads allowance, nonce and Blue's balance, but takes `forceDeallocatePenalty(adapter)` and the vault's idle `assetBalance` from the caller's snapshot. The contract reads both live (VaultExitBundlesV1.sol:138 and :146).

The reason this needs its own fix rather than being covered by an allowance bound: shares burned is approximately `previewWithdraw(exitAssets)` **regardless of the penalty**, while the value received is `idle + (E - idle)/(1 + p)`. Raising `p` shrinks what the user gets without changing what they burn, so no allowance cap detects it.

And the timelock is not the protection it looks like. `submit` sets `executableAt = block.timestamp + timelock[selector]` and `timelocked()` accepts `block.timestamp >= executableAt`, while VaultV2's own natspec says "At creation, all settings are set to their default values. Notably, timelocks are zero which is useful to set up the vault quickly." So on a default-configured vault a curator can submit and execute a penalty raise atomically in the victim's block, up to `MAX_FORCE_DEALLOCATE_PENALTY = 0.02e18`. `forceDeallocate` then does `withdraw(penaltyAssets, address(this), onBehalf)`, burning the exiting user's shares and sending the assets to the vault, so it is a real transfer to the remaining holders, up to about 1.96% of the in-kind portion, silently. An allocator can reach the same outcome with no timelock at all by moving idle into the adapter, forcing the whole exit through the penalized leg.

Both reads are free here since the multicall already exists. Suggest adding `forceDeallocatePenalty(adapter)` and `balanceOf(vault)` and throwing a typed error when the live values are worse than the snapshot promised (live penalty above snapshot penalty, or live idle below snapshot idle). Worth saying plainly in the JSDoc that this closes the stale-snapshot and careless-curator cases but not a same-block hostile curator, for which this entry point has no on-chain bound, unlike `vaultExitBundlesV1ForceWithdrawVaultV2`'s `minSharePriceE27`.

packages/morpho-sdk/src/entities/vaultV2/vaultV2.ts
```diff
@@ -117,6 +135,86 @@ export interface VaultV2Actions {
...
+  /**
+   * Prepares an illiquid Vault V2 exit into idle assets and Morpho Blue supply positions.
+   *
+   * The vault must have exactly one `MorphoMarketV1AdapterV2`. `amount` is penalty-inclusive and
+   * the caller controls market order. Call `getRequirements()` before `buildTx()` so Blue balance,
+   * allowance, and nonce are checked on-chain. Vault gates are enforced by the final transaction
+   * and are not preflighted: receive gates may depend on VaultExitBundlesV1's transient initiator,
+   * while arbitrary send-share gates may depend on intermediate state changed by the exit's
+   * multiple share burns. The SDK intentionally does not validate the user's share balance; size
+   * `amount` in asset terms so `amount <= vault.previewRedeem(sharesHeld)` instead of comparing it
```
tarikbellamine: This sizing rule is off by the contract's own rounding allowance, and it fails at exactly the most common call.

`VaultExitBundlesV1` V2 natspec:

```
L110: The assets value of the sum of the shares withdrawn can be greater than exitAssets, but no greater that exitAssets+N.
L112: Requires the sender to have enough shares to withdraw exitAssets+N.
```

N is the iteration count, one wei of rounding each. The V1 entry point's equivalent (L54) says only `exitAssets`, with no `+N`, so this is a V2-specific requirement.

A user doing a full exit and following this JSDoc literally sets `amount = previewRedeem(sharesHeld)`, which leaves no headroom for the `+N`, and the exit reverts. Given the SDK deliberately does not validate share sufficiency and pushes that onto the caller, the guidance it hands them needs to match the contract.

Suggest stating the rule with the buffer, along the lines of `amount + marketParamsList.length <= vault.previewRedeem(sharesHeld)`, and noting that V1 does not need the extra term.

tarikbellamine (dismissed):
Approving. Did a second pass scoped purely to security this time, reading the deployed `VaultExitBundlesV1` and `TokenLib` at `9994e6ab`, plus `Morpho.sol`, `VaultV2.sol`, `IGate.sol` and `MetaMorpho.sol`.

The hardening commit landed the important one: the rewritten V1 per-market assignment cap is sound in the dangerous direction. I tried to construct an SDK-accepts / chain-reverts input against it and could not. When it accepts, `assigned_m <= available_m` for every unique market, so each listed market's post-supply liquidity covers what was assigned to it and the trailing `withdraw` is always satisfiable. The `balanceOf(MORPHO)` term in `_withdrawable` never binds either, because the flash loan restores Blue's balance before the withdraw.

Also confirmed clean, so nobody needs to re-derive these:

- **Callback reachability is closed.** Blue dispatches both callbacks to its own caller (`Morpho.sol` L192, L428), so `onMorphoSupply` / `onMorphoFlashLoan` are only reachable when the periphery itself called Blue, with `data` it encoded one statement earlier. An attacker calling `Blue.flashLoan` gets the callback on their own contract. `require(msg.sender == BLUE)` is complete.
- **Hostile gates cannot steal.** All four `IGate` functions are `external view`, so Solidity emits STATICCALL. No mutation, no reentrancy into the standing allowance. The whole hostile-gate surface is denial of exit.
- **Signature handling.** The empty-permit sentinel is unforgeable (viem range-checks r/s and only yields `v` in {27,28}), the 64-byte EIP-2098 path correctly depends on the `yParity + 27` fallback, and high-s malleability recovers the same signer so it only burns a nonce, which `submitPermit`'s skip branch tolerates.
- **Nonce front-running is harmless** — a third party submitting the lifted permit just does the user's work.
- Fee-on-transfer and rebasing tokens revert rather than stranding value; a curator cannot disable a market with a live position; V2 penalty rounding favours the user.

One correction to my earlier review: I rated the missing permit `owner`/`spender` binding as high. That was wrong. `TokenLib.submitPermit` hardcodes `permit(msg.sender, address(this), ...)` with no try/catch, and the `Permit` struct carries neither field, so a mismatched signature just fails `ecrecover` and reverts. It is a robustness gap, not a vulnerability. Not worth blocking on.

Three medium items left inline. None blocks the merge in my view, but the penalty re-read closes real value extraction and the `+N` one is a one-line doc fix for a guaranteed revert on the most common call, so both are worth doing before this ships. I also left the low-severity items out of the comments: the V1 loop is over-strict in the safe direction (it rejects some exits the chain would execute, since `_withdrawMorpho` walks the whole withdrawQueue and the SDK never models queue markets outside the list), `vaultData.asset` is trusted where `MORPHO()` is verified on-chain in the same multicall, `marketParamsList` length is unbounded, and the parse-failure branch interpolates raw signature bytes into the error message. Happy to file any of those separately if you want them tracked.

packages/morpho-sdk/src/entities/vaultV2/vaultV2.ts
Rubilmax: In all other action flows, we dont care about aging input vault data, hence why we 100% rely on it here and wont be re-fetching onchain data instead of reading it and trusting the input data.

packages/morpho-sdk/src/entities/vaultV1/vaultV1.ts
```diff
@@ -313,6 +406,192 @@ export class MorphoVaultV1 implements VaultV1Actions {
...
+    if (covered < amount) {
+      throw new InKindRedeemCoverageError({
+        required: amount,
+        covered,
+        maxExitAssets: covered,
+      });
+    }
+
+    // V1 interest cannot lower share price, so the current preview upper-bounds execution.
+    const requiredShareAllowance = vaultData.toShares(amount);
```
tarikbellamine: This under-sizes the allowance on any vault with a non-zero fee, so the V1 exit reverts after the user has already approved or signed.

The comment's premise is right and the conclusion does not follow. V1 interest cannot lower the share price, but the un-accrued preview is a price the vault never quotes, because `MetaMorpho.withdraw` mints the fee shares before computing the burn:

```solidity
uint256 newTotalAssets = _accrueFee();   // _mint(feeRecipient, feeShares)
shares = _convertToSharesWithTotals(assets, totalSupply(), newTotalAssets, Math.Rounding.Ceil);
```

`totalSupply()` is read after the mint, so:

- authorized here: `ceil(amount * (S + virtualShares) / (TA + 1))`
- burnt on chain: `ceil(amount * (S + feeShares + virtualShares) / (TA + 1))`

The shortfall is about `amount * fee * pendingInterest / totalAssets`. It is non-zero whenever `fee > 0` and interest is pending, which is the steady state of every production vault, since `lastTotalAssets` only moves on deposit/withdraw while the underlying markets accrue continuously. Now that the allowance is bounded rather than `maxUint256`, `_spendAllowance` reverts `ERC20InsufficientAllowance`.

Against live state for `0xBEEF01735c132Ada46AA9aA4c54623cAA92A64CB` (the vault `vaultV1.inKindRedeem.integration.test.ts` pins: fee 5%, `totalAssets` 75044731072464, `lastTotalAssets` 75043751946943, `totalSupply` 6.6055e25), for that test's own `amount` of 1 USDC:

```
authorized : 880,204,295,696,434,754
burnt      : 880,204,869,908,000,735
short by   :         574,211,565,981
```

The bound only holds if `newInterest * (1 - fee) >= fee * pendingInterest`, which here needs roughly 120h of accrual inside a 2h deadline.

The V2 path in this same commit already does the right thing — it converts against `vaultData.accrueInterest(allowanceTimestamp)` and takes the max — so this reads as an asymmetry rather than a deliberate choice.

Suggest `const requiredShareAllowance = vaultData.accrueInterest(now).toShares(amount);`. `AccrualVault.accrueInterest` already reproduces `_accruedFeeShares` exactly (`feeAssets = wMulDown(totalInterest, fee)`, `feeShares = toShares(feeAssets, "Down")` over `totalAssets - feeAssets`), and it reproduces the contract's burn with zero slack on the numbers above. No second endpoint is needed on V1: `d/dt(totalAssets - feeAssets) = (1 - fee) * totalAssets' >= 0`, so accruing at `now` upper-bounds every later inclusion time.

Worth noting the unit fixtures cannot catch this as written — `test/fixtures/inKindRedeem.ts` sets `fee: 0n` and `lastTotalAssets: totalAssets`, which zeroes both terms that produce the shortfall. A fixture with `fee > 0n` and `lastTotalAssets < totalAssets` would regression-cover it.

tarikbellamine (approved):

PR #929, opened 2026-08-13
test: stabilize fork CI and standardize test layout

Rubilmax:
## Summary

- Standardize tests across packages: colocate unit tests as `*.test.ts`, keep integration and fork tests under each package's singular `test/` directory as `*.integration.test.ts`, and route both sets through explicit Vitest projects.
- Shard CI by package, running each package's unit and fork projects on one runner while preserving the stable `test / vitest` gate.
- Let Vitest scale fork files across workers while limiting concurrent fork fixtures within each worker; remove the bespoke per-RPC semaphore and process caps.
- Harden `@morpho-org/test` around Anvil startup, cancellation, retries, unexpected exits, shutdown, and cleanup. Vitest and Playwright fixtures now preserve both test and teardown failures.
- Redact fork-provider credentials from Anvil diagnostics at the source and sanitize failed Vitest reports before upload. Successful reports are no longer retained, and failed artifacts expire after one day.
- Move the WDK integration suite onto the shared pinned-block Anvil fixture and align its package scripts and guards with the standardized layout.
- Restore and save Foundry caches per package and run attempt after any non-cancelled test attempt, including failures.

## Why

Fork tests previously combined broad in-process concurrency with fragile Anvil startup and teardown. Failures could leak child processes, obscure the original error during cleanup, or leave CI with unsafe diagnostic artifacts. Test placement and project routing also varied between packages, while one oversized CI job made failures slow to isolate and retry.

The final design keeps one isolated Anvil process per fork case, uses Vitest's native worker scheduling instead of cross-process coordination, and makes lifecycle, routing, sanitization, and failure reporting explicit.

## Validation

- `pnpm run build:ci`
- `pnpm lint`
- Full `morpho-sdk` unit and fork shard with coverage
- Shared `test` unit and fork projects
- Scripts project, including report-redaction tests
- `git diff --check`

## Release

Includes a minor changeset for `@morpho-org/test`, covering the hardened lifecycle behavior and readonly public process/client fields.


Rubilmax:
LGTM even with codex comments

remiroyc (dismissed):

Rubilmax:
@codex

Rubilmax (commented):
LGTM

.changeset/calm-anvils-share.md
```diff
@@ -0,0 +1,7 @@
+---
+"@morpho-org/test": minor
+---
+
+Add opt-in per-RPC Anvil fork limits for shared runners, make unexpected exits fail teardown, clean up child processes when startup fails, and redact fork-provider credentials from recorded diagnostics.
+
+This minor release intentionally makes the existing `spawnAnvil()` result fields (`rpcUrl` and `stop`) and `ViemTestContext.client` readonly. This is a source-level breaking change for consumers that assign those fields; keep mutable wrapper objects locally if reassignment is required.
```
0xbulma, line 7: **SemVer: this must be `major`, not `minor`.** _(re-reviewed — 3 findings on this file folded here)_

**1. The level is wrong** — `release-integrity`, confidence 92%.

The changeset says it outright: *"This is a source-level breaking change for consumers that assign those fields."* AGENTS.md §7: **Major: removed/renamed/retyped public symbols.** `@morpho-org/test` is published (2.8.4, not `private`), so `minor` ships 2.8.4 → 2.9.0 and lands a compile-breaking change **inside every consumer's `^2.8.0` caret range** with no major-version signal.

Retyped public symbols:
- `spawnAnvil()` → `SpawnedAnvil` with `readonly rpcUrl` / `readonly stop`, plus a **required** `stopAndWait`
- `ViemTestContext.client` → `readonly`
- `stop()` behaviour: returns `false` on repeat calls and no longer owns cleanup or slot release (moved to the `close` handler)

**2. Correcting it to `major` breaks a peer dependent — and the audit is missing** _(new since my first pass)_ — confidence 76%.

```
packages/morpho-test/package.json:34   peerDependencies  @morpho-org/test: ^2.8.0
.changeset/config.json                 bumpVersionsWithWorkspaceProtocolOnly: true
```

`morpho-test` is the **only non-dev dependent** (the other 8 are `devDependencies`, which don't matter). Because Changesets only auto-bumps workspace-protocol ranges, that explicit `^2.8.0` is maintained by hand — and §7 says *"Do not rely on Changesets to infer peer-dependent releases."* At `major`, `^2.8.0` will not admit `3.0.0` and `@morpho-org/morpho-test` becomes unresolvable.

→ Widen to `"^2.8.0 || ^3.0.0"` and add a `morpho-test: patch` changeset **in this PR**.

**3. The body omits most of the new surface** — confidence 86%.

This text becomes the published CHANGELOG. Missing: `AnvilStartupError` / `AnvilProcessError` / `AnvilCleanupError` (new in `src/errors.ts`, re-exported from `src/index.ts:3`), `stopAndWait()`, the `SpawnedAnvil` / `SpawnAnvilOptions` exports, the `MORPHO_TEST_MAX_ANVIL_PROCESSES_PER_RPC` / `MORPHO_TEST_ANVIL_RUN_ID` env-var contract, and the `binary` fix (it was previously ignored as the executable *and* passed to anvil as a spurious `--binary` flag). Integrators pattern-matching on typed errors (§3) can't discover the new classes from these notes.

---
**Resolution:** either bump to `major` **and** do the peer audit in (2), or drop the `readonly` retypes so the release is genuinely additive.

.github/workflows/test.yml
```diff
@@ -30,28 +72,97 @@ jobs:
           cache: pnpm
 
       - name: Restore Foundry cache
+        if: matrix.uses_foundry
         uses: actions/cache/restore@2c8a9bd7457de244a408f35966fab2fb45fda9c8 # v6
         with:
           path: ~/.foundry/cache/rpc
-          key: foundry-${{ github.job }}-${{ github.sha }}
-          restore-keys: foundry-${{ github.job }}-
+          key: foundry-${{ github.job }}-${{ matrix.name }}-${{ github.sha }}-${{ github.run_attempt }}
```
0xbulma, line 82: **This fragments the Foundry RPC cache 9 ways and kills cross-package reuse.** _(re-reviewed — confidence 70%)_

The key went from one shared `foundry-vitest-<sha>` to per-package `foundry-package-<name>-<sha>-<attempt>`:

- Blocks fetched by `blue-sdk-viem` no longer warm `morpho-sdk`, `liquidity-sdk-viem`, etc. Every fork shard cold-fetches the same pinned blocks independently → **more RPC pressure, which is the root cause of the flakes this PR fixes.**
- ~9 cache entries per commit instead of 1, against a 10 GB per-repo LRU budget. Entries from prior commits evict faster, so the `restore-keys` fallback hits less often than it looks.
- **Additionally** _(new since my first pass)_: renaming the job `vitest` → `package` orphans **every existing `foundry-vitest-*` entry**, so the first run after this merges is fully cold across all 9 shards.

There is also no package-agnostic fallback tier left in `restore-keys`, so a newly added or renamed shard can never warm-start from another shard's cache.

→ Append `foundry-${{ github.job }}-` as a final `restore-keys` tier, and consider a shared key for the RPC cache — the blocks are pinned and package-agnostic.

.github/workflows/test.yml
```diff
@@ -30,28 +72,97 @@ jobs:
...
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: vitest-${{ matrix.name }}
+          path: vitest-reports
+          retention-days: 1
 
       - name: Save Foundry cache
-        if: always()
+        # Attempts need unique keys because GitHub caches are immutable.
+        if: success() && matrix.uses_foundry
```
0xbulma: **`always()` → `success()` means a failing shard discards the warm cache it just built.** _(re-reviewed — confidence 88%, highest-confidence CI finding in the re-review)_

This is the exact path that matters for flake recovery: shard fails → its newly-fetched RPC blocks are thrown away → because the key embeds `github.run_attempt`, the failed attempt's key is never written → the rerun falls through `restore-keys` to an **older commit's** cache and cold-fetches the same pinned blocks again from the metered provider → higher chance it flakes again.

That works directly against the PR's stated goal.

I take the point in the comment above (unique keys per attempt, so `always()` would accumulate entries) — but that argues for bounding the *number* of entries, not for discarding the cache precisely when it is most valuable. The Foundry RPC cache is append-only per-block JSON, so a failed run leaves it **valid, not corrupt**.

→ Restore `if: always() && matrix.uses_foundry`, keeping the `run_attempt` suffix to satisfy the immutable-key constraint.

.github/workflows/test.yml
```diff
@@ -14,8 +14,50 @@ permissions:
   contents: read
 
 jobs:
-  vitest:
+  package:
+    name: Package (${{ matrix.name }})
     runs-on: ubuntu-latest
+    strategy:
```
0xbulma, line 21: **The matrix is a hand-maintained duplicate of the project list, with no guard.** _(re-reviewed — raised independently by two agents, `ci-security` + `release-integrity`, confidence 85%)_

I verified the matrix covers all 20 projects in `vitest.config.ts` **today**. Nothing keeps it that way, and the failure mode is worse than I first described:

```
.github/workflows/push.yml   publish + version-pr  →  needs: test
```

Add a project to `vitest.config.ts` without editing this matrix and it **silently never executes**, while the aggregate `vitest` job (which only asserts `needs.package.result == success`) still reports green — and `publish` ships unverified package source. That contradicts §7's *"`main` is always releasable."* The old `pnpm run test:ci` ran every project by construction, so this diff introduces the false-green path. (The reverse direction fails loudly: `vitest --project <unknown>` errors.)

→ Add a parity step to the `vitest` job that reads the configured project names (e.g. `pnpm exec vitest list --json`) and fails when they differ from the union of the matrix `projects:` values — wiring the expected list through an `env:` binding rather than duplicating it a third time. Or generate the matrix with `fromJSON` from a preceding setup job.

**Separately (low, confidence 85%):** no `timeout-minutes` on either job, so both inherit GitHub's 360-minute default. With `fail-fast: false` and the retry math in `packages/test/src/vitest.ts` (see my comment there), a wedged fork can pin a runner for hours while the required check stays pending. Suggest ~30 min on `package`, ~10 on `vitest`.

packages/test/src/anvil.ts
```diff
@@ -282,42 +308,395 @@ function toArgs(obj: AnvilArgs) {
...
+      });
+
+      subprocess.stderr.on("data", (data) => {
+        // Startup warnings are diagnostic; only timeout, error, or early close is fatal.
+        if (args.forkUrl) {
+          forkStderrObserved = true;
+          if (settled && !stopRequested && !forkStderrWarningEmitted) {
+            forkStderrWarningEmitted = true;
+            console.warn(
+              `[port ${port || "??"}] Anvil emitted stderr output. Details were redacted because a fork URL is configured.`,
```
0xbulma: **All Anvil stderr is discarded for every CI fork test.** _(re-reviewed — `web3` confirms the mechanism at 90% confidence, but rates the severity **low**; recording that honestly rather than leaving my original framing)_

`createViemTest` always sets `forkUrl` (defaulting to `chain.rpcUrls.default.http[0]`), so this early-return fires on every fork test and the `stderr` accumulator stays empty. The diagnostics further down then degrade to `"Anvil stderr was redacted because a fork URL is configured"` with no cause — every fork startup/exit failure in CI becomes undiagnosable, in a PR whose purpose is diagnosing fork flakes. It also over-redacts the common case where the fork URL carries no credential at all.

It's the third redaction layer over the same secret:
1. GitHub Actions already masks registered `secrets.*` values in logs.
2. `scripts/redact-vitest-reports.mjs` already scrubs the uploaded artifact.
3. This.

→ Keep buffering stderr and redact only the secret — reuse the `redactSecrets` substitution set (raw, `new URL(v).href`, JSON-escaped, URL-encoded) over the accumulated text before it reaches the error messages and `console.warn`. Same applies to the `close` handler's buffer and the `error` path below.

scripts/redact-vitest-reports.mjs
```diff
@@ -0,0 +1,197 @@
...
+    (left, right) => right.length - left.length,
+  )) {
+    const occurrences = redacted.split(representation).length - 1;
+    if (occurrences === 0) continue;
+    replacements += occurrences;
+    redacted = redacted.replaceAll(representation, REDACTION);
+  }
+
+  for (const representation of representations) {
+    if (!redacted.includes(representation)) continue;
```
0xbulma: **This verification pass is tautological.** _(re-reviewed — confirmed, but rated **low** at 58% confidence; correcting my original weighting)_

It re-tests the same `representations` that `replaceAll` just removed, so it can only fire if a replacement *re-introduces* an earlier representation — which requires that representation to share an affix with the constant `REDACTION` (`<redacted-rpc-url>`). That's impossible for the URL-shaped secrets this script handles, so this second full pass over every report file is unreachable in practice.

The consequence is what matters: it cannot detect a **transformed** representation, which is the case that actually bites — ANSI colour codes injected mid-string by Vitest's error diffs, line-wrapping, truncation of long strings, or a partial URL without scheme. Any of those defeats the substring pass *and* this check, silently. `morpho-org/sdks` is public, so the artifact is world-readable.

→ Either fold the assertion into the replacement loop (re-check only the representation just replaced), or verify against a **decomposed** signal — the API-key path segment and the host — so transformations still trip it. Best of all: keep the credentialed URL out of the report at source (see my comment on `anvil.ts`).

**Note:** two stronger issues in this same file are flagged separately below — a fail-open main-guard, and the CLI's fail-closed branches being untested.

packages/test/src/anvilProcessSlot.ts
```diff
@@ -0,0 +1,456 @@
...
+      `Abandoned Anvil process "${pid}" could not be inspected. Check process permissions before retrying.`,
+      { cause: error },
+    );
+  }
+
+  let processDescription: string;
+  try {
+    // `eww` includes the full child environment on both BSD and GNU ps. The
+    // random marker survives exec and distinguishes this child from PID reuse.
+    processDescription = execFileSync(
```
0xbulma: **`ps eww` dumps the child's argv and full environment — and on failure that text is attached verbatim as an error `cause`.** _(re-reviewed — `web3` sharpened this into a concrete leak path, confidence 58%)_

I originally flagged this as "not logged today, but one careless `cause:` away." It's closer than that:

`ps eww -p <pid> -o command=` captures the Anvil child's command line — which contains `--fork-url https://<provider>/<API_KEY>`, since `toArgs` puts `forkUrl` on argv — **and** its entire inherited environment (`MAINNET_RPC_URL`, `BASE_RPC_URL`, `ARBITRUM_RPC_URL`). When `execFileSync` throws, the raw error object carries that captured text on its `stdout` / `stderr` / `output` properties, and it is attached verbatim as `cause` on the `AnvilCleanupError` just below. `spawnAnvil` re-throws `AnvilCleanupError` **unchanged**, so this path bypasses every `args.forkUrl ? "…redacted…" : error` guard `anvil.ts` applies to its spawn / stderr / close failures — and reaches the Vitest reporter and the uploaded blob report.

→ Never attach the raw child-process error. Build the cause the way `anvil.ts` does — a plain error stating the `ps` invocation failed, carrying `status` / `signal` only, explicitly dropping `stdout`, `stderr`, `output`. Better still, avoid reading the environment at all: `/proc/<pid>/environ` on Linux, or have the child touch an identity file keyed on the UUID.

**Scope note, now with evidence.** I flagged that this 456-line file is opt-in and unused here. That's confirmed from the other direction: `.github/workflows/test.yml` never sets `MORPHO_TEST_MAX_ANVIL_PROCESSES_PER_RPC`, so `acquireAnvilProcessSlot` short-circuits to the no-op slot and **the entire semaphore is inert in CI** — while fork concurrency is simultaneously unbounded. See my separate comment on the workflow's `Test` step.

0xbulma (dismissed):
## Review — verified, with findings

**Ran locally:** `pnpm lint` · `pnpm -w run build:ci` · all 12 non-fork projects (**185 files / 2693 tests passed**). Fork projects not run (no RPC).

### Verified safe — I checked these first because the diff looks alarming

- **No orphaned tests.** `vitest list --filesOnly` collects **238/238** test files tracked in git. Zero files missed by the rewritten include globs.
- **No lost cases.** Per-package test-case counts vs `main`: every package flat or up, **+54 net**. The `blue.test.ts` −781 and the WDK `tests/` deletion are moves, not drops.
- **Matrix is complete** — all 20 configured projects have a row.
- **Dropping `sequence: { concurrent: false }` on `evm-simulation` is fine.** All four `vi.mock`-sharing files use `describe.sequential`; the one plain-`describe` file (`bundler-retention.test.ts`) is 20/20 synchronous tests, which cannot interleave. 8/8 repeat runs green.
- `binary` was previously broken (ignored as the executable *and* passed as a bogus `--binary` flag) — `anvil.ts:397` fixes it.

### Findings

Seven inline. The changeset one is the only rules violation (AGENTS.md §7) and should be settled before merge; the rest are trade-offs worth a second look, several of which cut against this PR's own stability goal.

One more without a diff anchor: **`test:ci` (`package.json:36`) is now orphaned.** CI no longer runs it, so the command developers run locally and the command CI runs have drifted apart. Either delete it or redefine it as the matrix entrypoint.

Approving so this isn't blocked on me — the findings below are yours to triage.

.changeset/calm-anvils-share.md
Rubilmax: yes, acknowledged in the changeset. does not deserve a major for me because NOBODY should be overwriting those field

.github/workflows/test.yml
Rubilmax: more RPC pressure for the first 9 runs, then not anymore
the first 9 runs have already completed

10G lru budget is gigantic compared to the size of cache. everything's fine

0xbulma (commented):
Some open items to consider

.github/workflows/test.yml
Rubilmax: pointless to check the job matrix: both have distinct purposes and if the ci is missing a shard, the issue may be on the dev who forgot to add it (or maybe it is intentional)

0xbulma (commented):
## Re-review — `/facets:pr-review-local` (11-agent panel)

Re-ran the full review engine against `eb27628b..4ddc5d19`: **11 agents, 0 failed, 40 raw findings → 39 after dedup, 0 dropped by the scope filter.**

**Posting medium and above only** (1 high + 23 medium). The 15 low findings are held back as noise — happy to post them if you want the full set.

**My 7 earlier comments have been updated in place** rather than duplicated. Four were confirmed and sharpened, and two I've corrected downward — `web3` and `simplification` rate the Anvil-stderr and tautological-verification points **low**, not medium, so those comments now say so. The changeset comment absorbed two additional findings on the same file.

### What the panel found that my first pass missed

1. **The peer-range break** — `packages/morpho-test/package.json:34` pins `@morpho-org/test: ^2.8.0` in `peerDependencies`, and `bumpVersionsWithWorkspaceProtocolOnly: true` means Changesets will never auto-bump it. Correcting the changeset to `major` makes `morpho-test` unresolvable. Folded into the changeset comment.
2. **`registerChildProcess` is entirely untested** — the test stub has no `pid`, so `if (childPid !== undefined)` is always false and the child-record write that makes slot reclamation possible never runs under unit test.
3. **The semaphore is inert in CI** while fork concurrency is unbounded — neither slot env var is set, and the project-wide cap was removed rather than raised.
4. **A truncated-port parse** — the new stdout buffering fixes the split-banner case but not the split-port case it implies.
5. **A half-propagated rule edit** in `.agents/commands/pr-fix.md` — line 410 was missed while 403 and 796 were updated.

### What the panel cleared

- **No lost test coverage.** `vitest list` collects 238/238 tracked test files; per-package case counts vs `main` are flat or up (+54 net). The ~90-file rename cascade left no stale doc pointers (`docs` returned only two low findings after 44 tool calls).
- **Removing `sequence: { concurrent: false }` from `evm-simulation` is safe** — every `vi.mock`-sharing file uses `describe.sequential`, and the one plain-`describe` file is 20/20 synchronous tests. `tests` independently declined to flag it. 8/8 repeat runs green locally.
- **`dependencies` returned clean.** The lockfile delta matches the one added devDependency exactly.

Locally: `pnpm lint`, `pnpm -w run build:ci`, and all 12 non-fork projects (185 files / 2693 tests) pass.

.agents/commands/pr-fix.md
```diff
@@ -400,7 +400,7 @@ For each file with findings, build a complete understanding before touching anyt
    - Type definitions, interfaces, or schemas referenced at the flagged location
    - Helper modules used inside the changed function
 
-4. **Find callers** — use the **Grep tool** (preferred over raw grep — it handles regex metacharacters in symbol names safely). Search across `packages/**` (this monorepo has no top-level `test/` directory; tests live under `packages/*/test/` and colocated next to source in `morpho-sdk` / `evm-simulation`):
+4. **Find callers** — use the **Grep tool** (preferred over raw grep — it handles regex metacharacters in symbol names safely). Search across `packages/**` (this monorepo has no top-level `test/` directory; unit tests are colocated beside source and integration tests live under `packages/*/test/`):
```
0xbulma, line 403: **The test-layout rule change is only half-propagated in this file.** _[skill-authoring] · confidence 92%_

Line 403 (here) and line 796 were both updated to the new §5 split. But **line 410** — the main-flow counterpart of the very clause rewritten at 796, seven lines below this one in the same context-gathering block — still reads:

> read the corresponding `packages/<pkg>/src/index.ts` re-export entry point AND any test file under `packages/<pkg>/test/` that exercises the symbol.

Under the new layout `packages/<pkg>/test/` holds **only** `*.integration.test.ts`. So the main flow now sends the fixer straight past the colocated `src/**/*.test.ts` unit tests that are the primary coverage surface for a public export, while the watcher flow at 796 correctly points at both.

→ Rewrite line 410 to mirror 796: name the colocated unit tests (`src/**/*.test.ts`) alongside `packages/<pkg>/test/**/*.integration.test.ts`.

.github/workflows/test.yml
```diff
@@ -30,28 +73,103 @@ jobs:
...
         with:
           cache: false # Disable built-in cache to always save it.
 
       - run: pnpm install --frozen-lockfile
 
-      - run: pnpm run test:ci
+      - name: Test
+        id: test
+        env:
+          MAINNET_RPC_URL: ${{ matrix.uses_foundry && secrets.MAINNET_RPC_URL || '' }}
```
0xbulma, line 97: **The per-RPC semaphore this PR adds is inert in CI — and fork concurrency is unbounded.** _[performance] · confidence 65%_

This `env:` block sets neither `MORPHO_TEST_MAX_ANVIL_PROCESSES_PER_RPC` nor `MORPHO_TEST_ANVIL_RUN_ID`, so `acquireAnvilProcessSlot` (`packages/test/src/anvilProcessSlot.ts:178-183`) short-circuits to the no-op slot. All 456 lines of cross-process slot coordination — plus ~570 lines of tests — never execute on this repo's CI.

Meanwhile the guard that *was* there is gone: root `sequence.concurrent: true`, no `maxConcurrency` on any fork project, and Vitest defaults (workers = vCPUs−1, 5 concurrent tests per file) mean a fork shard on a 4-vCPU `ubuntu-latest` runner can hold **~15 concurrent Anvil mainnet forks**, each an independent fork-RPC consumer. CPU/memory thrash plus provider rate-limiting is precisely the flakiness this PR is chasing.

The old cap of `1` was rightly removed as too serial — but it was removed rather than *raised*.

→ Set a bounded-but-concurrent value here, e.g. `MORPHO_TEST_MAX_ANVIL_PROCESSES_PER_RPC: 4` with `MORPHO_TEST_ANVIL_RUN_ID: ${{ github.run_id }}-${{ github.run_attempt }}-${{ matrix.name }}`; or equivalently set `maxConcurrency` on the `*-fork` projects in `vitest.config.ts`. Either way the new semaphore should actually be exercised by the CI it was written for.

packages/test/src/vitest.ts
```diff
@@ -17,94 +22,178 @@ BigInt.prototype.toJSON = function () {
...
+    gasPrice: parameters.gasPrice ?? 0n,
+    blockBaseFeePerGas: parameters.blockBaseFeePerGas ?? 0n,
+  };
 
   return test.extend<ViemTestContext<chain>>({
-    // biome-ignore lint/correctness/noEmptyPattern: required by vitest at runtime
-    client: async ({}, use) => {
-      const { rpcUrl, stop } = await spawnAnvilWithRetry(parameters);
+    client: async ({ signal }, use) => {
+      const maxAttempts = process.env.CI ? 3 : 1;
```
0xbulma, line 72: **The two retry layers multiply: up to 9 Anvil spawns for one test.** _[performance] · confidence 80%_

`maxAttempts = process.env.CI ? 3 : 1` here nests inside the pre-existing root `retry: 2` in `vitest.config.ts:35`, so a single test can spawn **3 × 3 = 9** Anvil forks and cold-fetch fork state nine times.

Each fixture attempt can burn the full startup budget — `Math.max(args.timeout ?? 45_000, 45_000) + 15_000` = 60s in `anvil.ts` — plus the 1s/2s backoff below. So one unhealthy fork URL costs ~183s per fixture and roughly **9 minutes of runner time per test** before it reports failure, on a job with no `timeout-minutes`.

→ Bound the fixture by a wall-clock deadline rather than an attempt count (stop retrying once the elapsed time exceeds the fork `testTimeout` budget), and drop the root `retry` to 0–1 for the `*-fork` projects now that the fixture retries setup itself — so the two layers stop compounding.

packages/test/src/anvil.ts
```diff
@@ -282,42 +308,395 @@ function toArgs(obj: AnvilArgs) {
...
+          new AnvilStartupError(
+            "Anvil startup was cancelled before its RPC server began listening. Retry after the competing fork finishes.",
+            { cause: options.signal?.reason },
+          ),
+        );
+      }
+      subprocess.stdout.on("data", (data) => {
+        // Anvil can split its listening message across stdout chunks.
+        stdout = `${stdout}${data.toString()}`.slice(-1_024);
+        const listenMatch = stdout.match(/Listening on 127.0.0.1:(\d+)/);
```
0xbulma: **The buffering fix doesn't cover the split it implies: a chunk boundary inside the port digits still settles on a truncated port.** _[correctness] · confidence 75%_

The comment above says Anvil can split its listening message across stdout chunks — which is exactly why the accumulator was added. But `/Listening on 127.0.0.1:(\d+)/` has no terminator, so if the boundary falls **inside the digits** (`"…127.0.0.1:310"` then `"01\n"`) the regex matches on the first `data` event, sets `settled = true`, and takes the truncated port `310`.

`spawnAnvil` then returns `http://localhost:310`, where nothing is listening, and the fixture fails with an opaque connection error instead of the real port — a new flake source in a PR fixing flakes.

→ Only settle once the port is provably complete: match `/Listening on 127\.0\.0\.1:(\d+)\r?\n/` against the buffer (note the escaped host dots too — the current pattern's `.` are wildcards), so the trailing newline proves the digits aren't truncated.

packages/test/src/anvilProcessSlot.ts
```diff
@@ -0,0 +1,456 @@
...
+
+        try {
+          // Publishing a populated directory by rename prevents contenders from observing a
+          // partially written owner token. Existing non-empty slot directories are not replaced.
+          renameSync(candidatePath, lockPath);
+
+          let releasePromise: Promise<void> | undefined;
+          return {
+            registerChildProcess: (pid, identity) => {
+              if (!Number.isSafeInteger(pid) || pid <= 0) {
```
0xbulma: **`registerChildProcess` is never invoked by any unit test — the whole child-record path ships unexercised.** _[tests] · confidence 92%_

`createFakeAnvilProcess` in `anvil.test.ts` builds its stub as `Object.assign(new EventEmitter(), { stdout, stderr, exitCode, kill, unref })` — **no `pid`**. So `anvil.ts:646`'s `if (childPid !== undefined)` is always false in unit tests: the `child` record is never written, and the registration `fail(...)` wrapper at `anvil.ts:649-658` never runs either.

That leaves all three typed-error branches here untested — invalid PID (this line), empty identity (:249), and slot-ownership loss between reservation and startup (:259) — and, more importantly, the child record is the thing that makes abandoned-slot reclamation possible at all. Only the crash fork fixture touches the happy path.

→ Add `pid` to `createFakeAnvilProcess`, then add cases in `anvilProcessSlot.test.ts` that call `acquireAnvilProcessSlot` directly and assert `AnvilStartupError` for `registerChildProcess(0, id)`, `registerChildProcess(pid, "")`, and for an `owner` file rewritten by another worker before registration.

packages/test/src/anvilProcessSlot.ts
```diff
@@ -0,0 +1,456 @@
...
+          try {
+            lockedOwner = readFileSync(join(lockPath, "owner"), "utf8").trim();
+          } catch (readError) {
+            if (hasErrorCode(readError, "ENOENT")) continue;
+            throw readError;
+          }
+
+          const separatorIndex = lockedOwner.indexOf("-");
+          const ownerPid = Number(lockedOwner.slice(0, separatorIndex));
+          if (!Number.isSafeInteger(ownerPid) || ownerPid <= 0) continue;
```
0xbulma: **A corrupt owner token silently `continue`s instead of reporting the malformed lock.** _[error-handling] · confidence 85%_

When `<slot>.lock/owner` is corrupt or its PID prefix is unparseable — `indexOf("-")` returns −1, or the prefix isn't numeric — this skips to the next slot. Every sibling branch in this catch throws `AnvilCleanupError` telling the operator to remove the stale lock; this one stays quiet.

The slot then becomes permanently unreclaimable and the reservation loop spins until `assertReservationActive()` throws the generic *"could not reserve a process slot within 120000 ms — increase the shared RPC budget or reduce fork concurrency"*, which points the operator at the wrong cause entirely. With `MORPHO_TEST_MAX_ANVIL_PROCESSES_PER_RPC=1` it deadlocks every fork test in the run.

→ Throw an `AnvilCleanupError` naming the slot and the malformed owner token, matching the wording of the invalid-child-record error at :383.

packages/test/src/anvilProcessSlot.ts
```diff
@@ -0,0 +1,456 @@
...
+          if (error instanceof AnvilStartupError) throw error;
+          if (
+            !hasErrorCode(error, "EEXIST") &&
+            !hasErrorCode(error, "ENOTEMPTY")
+          )
+            throw error;
+
+          let lockedOwner: string;
+          try {
+            lockedOwner = readFileSync(join(lockPath, "owner"), "utf8").trim();
```
0xbulma: **The read-owner-or-skip block is written verbatim three times.** _[simplification] · confidence 90%_

Lines 334-340, 354-360, and 410-415 are identical:

```ts
let x: string;
try { x = readFileSync(join(lockPath, "owner"), "utf8").trim(); }
catch (readError) { if (hasErrorCode(readError, "ENOENT")) continue; throw readError; }
```

Three call sites clears AGENTS.md §1's bar for extracting a helper ("at least three call sites"), and the repetition is a real comprehension cost in the reclamation flow, which is the hardest logic in the file.

→ Add a module-local `readLockOwner(lockPath): string | undefined` returning `undefined` on ENOENT and rethrowing otherwise, reducing each site to `const owner = readLockOwner(lockPath); if (owner === undefined) continue;`.

packages/test/src/anvilProcessSlot.ts
```diff
@@ -0,0 +1,456 @@
...
+  } catch (error) {
+    if (hasErrorCode(error, "ESRCH")) return;
+    throw new AnvilCleanupError(
+      `Abandoned Anvil process "${pid}" could not be stopped gracefully. Check process permissions before retrying.`,
+      { cause: error },
+    );
+  }
+
+  const gracefulDeadline = Date.now() + ANVIL_ABANDONED_PROCESS_GRACE_PERIOD_MS;
+  while (Date.now() < gracefulDeadline) {
```
0xbulma: **The poll-until-ESRCH loop is duplicated, and the bare liveness probe appears twice more.** _[simplification] · confidence 82%_

`terminateAbandonedAnvilProcess` repeats the same loop at 115-126 (grace period) and 146-157 (post-SIGKILL) — identical `process.kill(pid, 0)` / ESRCH-return / `AnvilCleanupError`-throw / `await setTimeout(RETRY_DELAY)` bodies, differing only in the deadline constant and the message. The same probe-with-ESRCH-handling shape also appears at 45-53 and 69-79.

→ Extract one `waitForProcessExit(pid, timeoutMs, inspectionFailureMessage): Promise<boolean>` and use it for both waits (and reuse the probe for the bare checks).

packages/test/src/anvil.ts
```diff
@@ -282,42 +308,395 @@ function toArgs(obj: AnvilArgs) {
...
+    });
+    let stopInitiated = false;
+    let stopRequested = false;
+    let forceKillTimeout: ReturnType<typeof globalThis.setTimeout> | undefined;
+    let processCloseTimeout:
+      | ReturnType<typeof globalThis.setTimeout>
+      | undefined;
+    let processCloseObserved = false;
+    let cleanupAwaited = false;
+    let resolveProcessClosed = () => {};
```
0xbulma: **Hand-rolled deferred promise resolvers where the platform has the primitive.** _[simplification] · confidence 85%_

Lines 417-422 declare `resolveProcessClosed` / `rejectProcessClosed` with dummy initializers that exist only to satisfy definite-assignment, then reach into a `new Promise` executor to overwrite them. The repo targets ESNext on Node ≥26, where this is one line:

```ts
const { promise: processClosed, resolve: resolveProcessClosed, reject: rejectProcessClosed } =
  Promise.withResolvers<void>();
```

Worth doing here specifically because `spawnAnvil` is now a ~300-line function carrying five mutable lifecycle flags — every construct that can shrink is worth shrinking.

packages/test/src/anvil.test.ts
```diff
@@ -0,0 +1,962 @@
...
+      cause: expect.objectContaining({ code: "ENOTDIR" }),
+    });
+    expect(spawnMock).not.toHaveBeenCalled();
+  });
+
+  test("error: AnvilStartupError bounds startup and frees its slot", async () => {
+    process.env.MORPHO_TEST_MAX_ANVIL_PROCESSES_PER_RPC = "1";
+    process.env.MORPHO_TEST_ANVIL_RUN_ID = `unit-startup-timeout-${process.pid}`;
+    const forkUrl = "https://startup-timeout-rpc.example";
+    const rpcId = createHash("sha256")
```
0xbulma: **The private lock-directory layout is re-derived by hand ten times in this file.** _[simplification] · confidence 85%_

`anvilProcessSlot.ts:218-221` owns the derivation — `runId` sanitisation, `sha256(forkUrl).slice(0, 16)`, `join(tmpdir(), "morpho-test-anvil", runId, rpcId)`. It's recomputed at lines 311-320, 365-374, 526, 578, 625, 678, 726, 776, 876, 914 here, and once more in `test/fixtures/vitest-semaphore-crash/holder.fixture.ts`.

Twelve places to edit for one layout change — and the failure mode is silent: the tests keep passing against a path the implementation no longer uses.

→ Export an `@internal` `anvilSlotLockDirectory({ forkUrl, runId })` from `anvilProcessSlot.ts`, use it inside `acquireAnvilProcessSlot`, and call it from the tests and the fixture.

packages/test/test/fixtures/vitest-semaphore-crash/holder.fixture.ts
```diff
@@ -0,0 +1,35 @@
...
+const test = createViemTest(mainnet, {
+  forkBlockNumber: 0,
+  forkUrl,
+  stepsTracing: false,
+});
+
+test("holds Anvil until its worker is killed", async ({ client }) => {
+  const childRecord = JSON.parse(
+    await readFile(
+      join(tmpdir(), "morpho-test-anvil", safeRunId, rpcId, "0.lock", "child"),
```
0xbulma: **Same layout duplication as `anvil.test.ts`, but here it degrades the test into a false pass.** _[correctness] · confidence 80%_

This fixture reimplements the private derivation owned by `anvilProcessSlot.ts`: the `runId.replaceAll(/[^a-zA-Z0-9_-]/g, "_")` sanitisation, the `sha256(forkUrl).slice(0, 16)` rpcId, and the `<tmpdir>/morpho-test-anvil/<runId>/<rpcId>/0.lock/child` layout.

The consequence is worse than duplication: if the layout changes, this copy goes stale and the orphan-reclamation test in `test/vitest.integration.test.ts` degrades to the opaque *"The crash fixture did not acquire Anvil in time"* timeout rather than a clear failure — so the reclamation path silently stops being tested.

→ Export an `@internal` `getAnvilSlotChildRecordPath({ forkUrl, runId, slot })` and call it from both `acquireAnvilProcessSlot` and here.

packages/test/test/vitest.integration.test.ts
```diff
@@ -0,0 +1,243 @@
...
+      }
+      if (originalMaxProcesses === undefined)
+        delete process.env.MORPHO_TEST_MAX_ANVIL_PROCESSES_PER_RPC;
+      else
+        process.env.MORPHO_TEST_MAX_ANVIL_PROCESSES_PER_RPC =
+          originalMaxProcesses;
+      if (originalRunId === undefined)
+        delete process.env.MORPHO_TEST_ANVIL_RUN_ID;
+      else process.env.MORPHO_TEST_ANVIL_RUN_ID = originalRunId;
+      if (replacement !== undefined) await replacement.stopAndWait();
```
0xbulma: **Sequential `await`s in `finally` — and this PR makes the first one far likelier to reject.** _[error-handling] · confidence 85%_

The block awaits `replacement.stopAndWait()`, then `upstream.stopAndWait()`, then `rm(coordinationDirectory)`. This PR gives `stopAndWait()` two new rejection paths — `AnvilProcessError` on any unexpected Anvil exit, `AnvilCleanupError` on slot-release failure — so a rejection from the first call now skips the other two and **leaks the still-running `upstream` Anvil process, its reserved slot, and the temp directory** for the rest of the CI run.

Same pattern at lines 53-56 and 101-104.

→ Run them independently and surface every failure:

```ts
const results = await Promise.allSettled([...]);
const rejected = results.filter(r => r.status === "rejected");
if (rejected.length > 0) throw new AggregateError(rejected.map(r => r.reason), "Fork test cleanup failed.");
```

packages/test/src/playwright.ts
```diff
@@ -11,28 +12,57 @@ export const createViemTest = <chain extends Chain>(
...
-      await use(client);
+      try {
+        await stopAndWait();
+      } catch (error) {
+        cleanupFailed = true;
+        cleanupFailure = error;
+      }
 
-      await stop();
+      if (fixtureFailed && cleanupFailed)
```
0xbulma, line 76: **Fourth copy of the same failure-composition tail.** _[simplification] · confidence 72%_

The `fixtureFailed` / `cleanupFailed` tail added here (30-65) is a verbatim copy of the one added in `vitest.ts:166-196`, and both hand-build exactly what `createAnvilFailureCleanupError` in `anvil.ts:18-28` already builds — `new AnvilCleanupError(message, { cause: new AggregateError([failure, cleanupError], summary) })`. A fourth manual copy sits at `vitest.ts:127-135`.

→ Give `createAnvilFailureCleanupError` an optional `summary` parameter, export it `@internal` from `errors.js`, and call it from all three sites.

packages/test/test/fixtures/vitest-semaphore/worker-a.fixture.ts
```diff
@@ -0,0 +1,44 @@
...
+  `${process.pid}\n`,
+);
+
+const test = createViemTest(mainnet, {
+  forkBlockNumber: 0,
+  forkUrl: process.env.MORPHO_TEST_SEMAPHORE_RPC_URL,
+  stepsTracing: false,
+});
+
+test("worker A acquires the shared slot", async ({ client }) => {
```
0xbulma: **`worker-a` and `worker-b` are byte-identical except two string literals.** _[simplification] · confidence 85%_

All 44 lines match apart from the `worker-a`/`worker-b` literals and the test title — the ready-file write, `createViemTest` setup, `active.lock` exclusive open, ready-count poll loop and done-file write are all duplicated.

→ Move the body into a sibling helper (e.g. `worker.fixture-helper.ts` exporting `registerWorkerFixture(name: "worker-a" | "worker-b")`) and reduce each file to an import plus one call — keeping the two separate files Vitest needs to get parallel workers.

scripts/redact-vitest-reports.mjs
```diff
@@ -0,0 +1,197 @@
...
+  if (files === 0) {
+    throw new VitestReportSanitizationError(
+      `No Vitest reports were found in "${directory}". Do not upload an unverified artifact.`,
+    );
+  }
+
+  return { files, replacements };
+}
+
+if (
```
0xbulma, line 178: **The sanitizer is fail-open at its entry point — and the workflow never checks that it ran.** _[ci-security] · confidence 55%_

Every fail-closed guard in this file — the missing-env check, the `realpathSync` containment check, the zero-files check, the post-redaction verification — lives **inside** this `import.meta.url === pathToFileURL(process.argv[1]).href` guard.

If that equality ever fails, the module loads, does nothing, and exits **0**. Node realpath-resolves the ESM main module while `process.argv[1]` keeps the literal invocation path, so any symlinked component in the checkout or workspace path desynchronizes them. `.github/workflows/test.yml:117` then proceeds to `upload-artifact` the **unredacted** blob report — and `morpho-org/sdks` is public, so that artifact is world-readable.

The workflow's only signal is the exit code; it never asserts that sanitization actually happened. I've flagged the confidence at 55% because I could not demonstrate the desync on a realistic runner path — but the design is fail-open, and for the sole control on a public artifact it should be fail-closed by construction.

→ Compare against the resolved real path (`pathToFileURL(realpathSync(process.argv[1])).href`), or move the CLI body into a dedicated entry file with no guard at all. **And** have the workflow step assert the success line, e.g. `node scripts/redact-vitest-reports.mjs vitest-reports | tee /dev/stderr | grep -q '^Verified '`, so a silent no-op fails the job *before* upload.

scripts/redact-vitest-reports.mjs
```diff
@@ -0,0 +1,197 @@
...
+    if (directory === undefined) {
+      throw new VitestReportSanitizationError(
+        "Pass the Vitest report directory as the first argument.",
+      );
+    }
+
+    const missingSecrets = RPC_SECRET_ENV_NAMES.filter(
+      (name) => !process.env[name],
+    );
+    if (missingSecrets.length > 0) {
```
0xbulma, line 193: **The CLI's fail-closed branches — the ones CI actually depends on — have no test.** _[tests] · confidence 78%_

`redact-vitest-reports.test.mjs` imports and exercises only `redactSecrets` and `sanitizeVitestReports`. The CLI block (163-197) is untested, so none of these are verified: the missing-directory-argument guard (:169), this missing-RPC-credentials guard (:178), and `process.exitCode = 1` on failure (:195).

That matters more than usual because a regression here **fails open silently** — the workflow step would still exit 0 and upload the report.

→ Add a case that runs the script as a child process with one RPC env var unset — `spawnSync(process.execPath, [scriptPath, tempDir], { env })`, the same pattern the `agents-engine` script tests already use — asserting a non-zero `status` and that the report file on disk is unchanged.

scripts/redact-vitest-reports.mjs
```diff
@@ -0,0 +1,197 @@
...
+
+    const result = sanitizeVitestReports(
+      directory,
+      RPC_SECRET_ENV_NAMES.map((name) => process.env[name] ?? ""),
+    );
+    process.stdout.write(
+      `Verified ${result.files} Vitest report file(s); redacted ${result.replacements} credential occurrence(s).\n`,
+    );
+  } catch (error) {
+    const message =
```
0xbulma, line 207: **The CLI catch discards the `cause` chain the function deliberately attaches.** _[error-handling] · confidence 82%_

`sanitizeVitestReports` wraps the underlying fs/verification failure with `{ cause: error }` at 148-151 precisely so the caller can report it. This catch prints only `error.message`.

So when redaction fails in CI, the job is blocked with nothing but the generic *"Could not sanitize reports in 'vitest-reports'. Do not upload the Vitest report."* — no filename, no errno. On a step whose whole job is to fail closed before a public artifact upload, that's the one moment you need the detail.

→ Append the cause when present, e.g. the cause's `stack` (falling back to `message`, or `String(cause)` for non-Errors) before setting `process.exitCode = 1`.

packages/test/src/vitest.ts
Rubilmax: > on a job with no timeout-minutes.

does not happen. closed

0xbulma (commented):
## Re-review at `e08481ba` — 4 findings, and several withdrawn

I re-ran the full panel against the current head and then checked it against **the actual CI run** ([32237537252](https://github.com/morpho-org/sdks/actions/runs/32237537252)) rather than reasoning from config. That changed my conclusions, so this comment leads with what I got wrong.

### Corrections to my earlier review

**I was wrong about the fork shard not fitting its budget.** I claimed `morpho-sdk` would need ~15 minutes and couldn't finish. It runs in **3m17s** against `timeout-minutes: 20`. My arithmetic assumed `maxConcurrency: 1` serialized the whole project; it only bounds tests *within* a file, and files run in parallel — the log shows 403s of test time compressed into 172s wall clock, ~2.3×. Sorry for the noise on that one.

**Withdrawn entirely** after the run showed zero retries and zero Anvil errors across 107 files:
- removing `max-parallel: 4` leaving fork concurrency unbounded
- the two retry layers multiplying to 9 Anvil spawns
- `!cancelled()` skipping the Foundry cache save on job timeout

**Re-rated down.** My "world-readable artifact" framing was too broad: redaction and upload are gated `if: failure() && steps.test.outcome == 'failure'`, so no artifact exists on a green run. And because `forkUrl` is `process.env.MAINNET_RPC_URL` verbatim, the runner's own masking already covers the whole-value case — confirmed as `***` in the green log.

### On the resolved threads

All 24 of my earlier comments are marked resolved. Most genuinely are — deleting `anvilProcessSlot.ts` closed six at once, and that was the right call. Three were resolved but the underlying finding is still open, so I'm re-raising them here rather than reopening old threads:

| Thread | Status |
|---|---|
| `.changeset:7` — semver + dependent audit | still `minor`; the peer range moved the *opposite* way (see inline) |
| `test.yml:20` — matrix parity guard | still absent |
| `anvil.ts:561` — stderr redaction | stream no longer dropped ✅, but a fragment path survives (see inline) |

### The one decision I can't make for you

The changeset has been `minor` across three reviews. §7 says **"Major: removed/renamed/retyped public symbols,"** and the changeset body itself says *"This is a source-level breaking change for consumers that assign those fields."* `spawnAnvil()`'s return became `SpawnedAnvil` with `readonly rpcUrl`/`readonly stop`, and `ViemTestContext.client` **and** `PlaywrightTestContext.client` both became `readonly` (the last one isn't mentioned in the body).

Three rounds unactioned reads as a deliberate choice rather than an oversight, so rather than repeat it a fourth time: **major, or drop the `readonly` retypes and ship genuinely additive at minor?** Either is defensible — shipping a self-declared breaking change as a minor inside every consumer's `^2.8.0` caret isn't.

### Verified locally
`pnpm lint` clean · 186 files / 2705 tests pass · CI green at this commit.

Holding back 11 low-severity findings (mostly duplicated client setup in the newly-colocated tests) — happy to post them if useful, but they're noise while the two highs are open.

packages/test/src/vitest.ts
```diff
@@ -16,94 +22,200 @@ BigInt.prototype.toJSON = function () {
...
 
+      const { client, stopAndWait } = initialized;
+      const cleanupState: { failed: boolean; failure: unknown } = {
+        failed: false,
+        failure: undefined,
+      };
+
+      // Report teardown after Vitest removes its fixture cleanup callback so a
+      // retry cannot replay the first attempt's rejected cleanup promise.
+      onTestFinished(() => {
```
0xbulma, line 193: **HIGH — the teardown hook binds to the wrong test under concurrency.** _[error-handling] · confidence 90%_

`onTestFinished` imported from `"vitest"` resolves its target via `getCurrentTest()`, a **module-global** in `@vitest/runner`. This repo runs `sequence: { concurrent: true }`, so by the time the fixture resumes after `await spawnAnvil(...)`, that global points at whichever test most recently entered `runTest` — not the one that owns this fixture.

Verified empirically with a throwaway fixture in the `test` project: **every** fixture registered against a sibling, never its own task.

Two consequences:
1. An `AnvilCleanupError`/`AnvilProcessError` in `cleanupState` is reported against an unrelated test. If that test finishes first, the runner clears `onFinished` and the hook never fires — the cleanup failure, and the stray Anvil process it signals, are **silently swallowed**, because the `finally` no longer rethrows.
2. If every later-started test has already completed, `getCurrentTest()` is `undefined` and `createTestHook` throws *before* `use(client)` — so `stopAndWait()` never runs and the Anvil leaks.

`maxConcurrency: 1` only covers fork projects under CI. Local fork runs and the CI unit projects (which exercise `createViemTest` in `vitest.test.ts`) still run 5-wide. And `vitestCompatibility.test.ts` replaces `onTestFinished` with a synchronous mock, so no test covers the real binding.

→ Take the hook off the test context instead of the module import:

```ts
client: async ({ signal, onTestFinished }, use) => { ... }
```

The context-bound hook is tied to the owning task and is the documented form for concurrent tests.

scripts/redact-vitest-reports.mjs
```diff
@@ -0,0 +1,220 @@
...
+    const serializedValues = new Set([secret]);
+    try {
+      const url = new URL(secret);
+      serializedValues.add(url.href);
+      serializedValues.add(url.hostname);
+      if (url.host !== url.hostname) serializedValues.add(url.host);
+      if (url.username) serializedValues.add(url.username);
+      if (url.password) serializedValues.add(url.password);
+      const lastPathSegment = url.pathname.split("/").filter(Boolean).at(-1);
+      if (lastPathSegment) serializedValues.add(lastPathSegment);
```
0xbulma, line 60: **The redaction hardening corrupts the report it's meant to protect.** _[correctness] · confidence 85%_

`MIN_DERIVED_SECRET_LENGTH` is applied only to `url.searchParams` values (line 63). `url.username`, `url.password` and `lastPathSegment` go in **ungated** — even though the comment two lines below states exactly the hazard they share.

I reproduced it:

```
secret: https://user:s3cr3t@rpc.example/eth

IN : {"tests":[...,"eth_call method works"],"method":"eth_getBalance"}
OUT: {"tests":[...,"<redacted-rpc-url>_call m<redacted-rpc-url>od works"],
      "m<redacted-rpc-url>od":"<redacted-rpc-url>_getBalance"}
```

`user` and `eth` are short enough to appear as ordinary substrings, so they're replaced across test names and **inside a JSON key** — `"method"` became `"m<redacted-rpc-url>od"`.

The output is still *syntactically valid* JSON, so the `JSON.parse` guard at line 149 passes it, the corrupted blob uploads, and `vitest --merge-reports` then fails on the renamed key. That turns a failing fork shard into a failing aggregation job with an unrelated-looking error.

→ Hoist the length gate so every **derived** representation passes it — `url.username`, `url.password` and `lastPathSegment` too — keeping only `secret`, `url.href`, `url.host` and `url.hostname` unconditional.

packages/morpho-test/package.json
```diff
@@ -31,7 +31,7 @@
   "peerDependencies": {
     "@morpho-org/blue-sdk": "^6.0.0",
     "@morpho-org/morpho-ts": "^2.7.0",
-    "@morpho-org/test": "^2.8.0",
+    "@morpho-org/test": "^2.9.0",
```
0xbulma: **This moved the opposite way from what the audit needed.** _[release-integrity, dependencies] · confidence 90%_

My earlier comment asked for this range to be **widened** so a major bump of `@morpho-org/test` stays resolvable. It was **narrowed** instead — `^2.8.0` → `^2.9.0` — and shipped as `"@morpho-org/morpho-test": patch`.

Narrowing *drops* 2.8.x from the accepted set. Every consumer on `^4.0.0` picks that patch up automatically, and anyone pinned to `@morpho-org/test@2.8.x` then hits an unsatisfied peer — ERESOLVE on npm 7+, a hard failure under pnpm `strict-peer-dependencies`. §7 defines patch as "bug fixes and internal source maintenance"; silently removing accepted compatibility isn't that.

It also isn't needed. `morpho-test`'s `src/` is untouched by this PR, and its entire consumption of the package is one symbol:

```
packages/morpho-test/src/fixtures/markets.ts:7  import { randomAddress } from "@morpho-org/test";
packages/morpho-test/src/fixtures/vaults.ts:7   import { randomAddress } from "@morpho-org/test";
```

`randomAddress` comes from `packages/test/src/fixtures.ts`, which this PR doesn't modify. Nothing here needs 2.9.0.

It also hardcodes the assumption that `@morpho-org/test` publishes as 2.9.0 — if the changeset is corrected to major (see the review body), `^2.9.0` is unsatisfiable and this file needs editing again.

→ Restore `"^2.8.0"` and drop the `morpho-test: patch` entry — there's then no published-contract change left for it. If the narrowing is deliberate, it needs to be `minor`/`major`, with the dropped 2.8.x support stated in the body.

packages/test/src/anvil.ts
```diff
@@ -282,42 +313,305 @@ function toArgs(obj: AnvilArgs) {
...
+        resolve();
+      });
+
+      subprocess.stderr.on("data", (data) => {
+        // Startup warnings are diagnostic; only timeout, error, or early close is fatal.
+        const dataString = data.toString();
+        stderr = redactAnvilDiagnostic(`${stderr}${dataString}`, args).slice(
+          -4_096,
+        );
+        if (!settled || stopRequested) return;
```
0xbulma: **Residual of the stderr fix: a fragment path still reaches the public job log.** _[web3] · confidence 88%_

Credit where due — the stream is no longer dropped wholesale, which was the substance of my earlier comment. This is the leftover.

The line buffer is fed only *after* `settled`: this guard short-circuits before `pendingForkStderr` is written. But Anvil emits fork diagnostics **before** it starts listening — retry/backoff lines, and CI runs `--retries 10 --fork-retry-backoff 500`, so pre-settle stderr is the expected traffic here.

So a stderr line carrying the fork URL that straddles the settle instant loses its head, and the tail is then treated as a whole line. `redactAnvilDiagnostic` matches only whole values (`secret`, `new URL(secret).href`), so a fragment matches nothing. Reproduced against this exact code using the file's own fake-subprocess harness:

```
chunk 1 (pre-settle): "retrying fork request to https://eth-mainnet.example/v2/SUPERSE"
chunk 2:              "Listening on 127.0.0.1:31099"
chunk 3:              "CRETAPIKEY failed\n"
  → console.warn("[port 31099] /SUPERSECRETAPIKEY failed")
```

GitHub's masking can't help — it matches the exact registered secret, and only a fragment is present. **Scoping this honestly:** it needs Anvil to actually emit fork stderr, so it doesn't fire on a green run, and I found no fragments in the current green log. It's conditional on a fork retry — which is precisely the flaky-CI scenario this PR exists to handle.

One more thing that makes this worse than it looks: `IS_LOGGING_DISABLED: true` (`.github/workflows/test.yml:99`) **is read by nothing**. I grepped the working tree, `node_modules`, and the base commit — only the workflow lines that *set* it. So there is no console suppression in CI; the default reporter forwards these warnings verbatim into the world-readable log on every run.

→ Two parts:
1. Move the line-splitting above this guard so pre-settle bytes seed the buffer, keeping only the `console.warn` loop behind it, and flush on `close`.
2. Widen `redactAnvilDiagnostic` to the representation set `redactSecrets` already builds — ideally by lifting that builder into one shared helper both call, so the two redactors can't drift again.

scripts/redact-vitest-reports.mjs
Rubilmax: it's not a full url redaction on purpose, because we are not designing a fully-featured redaction software. Let's limit the redaction to what actually can leak in our test reports: the rpc url. Closed.

packages/test/src/anvil.ts
Rubilmax: never happens, but ill fix just in case

Rubilmax: actually not fixed because it should only be redacted from test reports to avoid ci leaks, not from tests directly at the anvil level

jinmel (dismissed):

.changeset/calm-anvils-share.md
```diff
@@ -0,0 +1,7 @@
+---
+"@morpho-org/test": minor
```
jinmel, line 2: **[P1] Preserve API mutability or publish a major release**

For TypeScript consumers that reassign `spawnAnvil().rpcUrl`, `.stop`, or a fixture `client`, this change makes existing code stop compiling, as the changeset itself acknowledges. Publishing it as `minor` violates the repository rule that retyped public symbols require a major bump ([AGENTS.md §7](https://github.com/morpho-org/sdks/blob/bcdbc735609fc70c7785bb93d50fc60f83c816c6/AGENTS.md#L167-L169)). Either keep these fields mutable for a 2.x minor release or mark this as major and perform the dependent peer-range audit.

Rubilmax: acknowledged in the changeset already!

PR #939, opened 2026-08-25
fix(morpho-sdk): bound in-kind redeem allowance

Rubilmax:
## Motivation

Vault V2 in-kind redemption can burn one more base-unit-equivalent of shares than its bounded permit or approval when market accrual shifts assets between separately rounded penalty legs. The resulting allowance underflow surfaces as EVM panic 0x11. Fork CI also sends excessive RPC traffic when tests within every file run concurrently.

## Solution

- Include the documented per-market V2 rounding buffer in bounded share permits and approvals.
- Cover the adjusted allowance bounds in unit and fork-backed integration tests.
- Serialize tests within each file only on CI while retaining file-level parallelism.

## Validation

- `pnpm lint`
- `pnpm --filter @morpho-org/morpho-sdk build`
- Full CI-mode morpho-sdk suite: 1,163 tests passed
- In-kind redemption integration test passed three times with concurrency forced on

### DIFF
--- a/.changeset/calm-vaults-redeem.md
+++ b/.changeset/calm-vaults-redeem.md
@@ -1,0 +1,5 @@
+---
+"@morpho-org/morpho-sdk": patch
+---
+
+Include Vault V2 per-market rounding headroom in bounded in-kind redeem share permits and approvals.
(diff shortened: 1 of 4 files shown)


Rubilmax (commented):
LGTM

Foulks-Plb (approved):

PR #949, opened 2026-08-26
feat(morpho-sdk): complete protocol facades and shared-cap search

Rubilmax:
## Motivation
Recent Vault V2 in-kind redemption and public allocator work left facade gaps in morpho-sdk, while non-monotonic shared-cap rounding could select reallocations far above requested liquidity.

## Solution
This adds raw Blue and Midnight category subpaths, protocol-qualified shared-facade aliases with working legacy deprecations, forwards the missing allocator and in-kind surfaces, updates docs and reviewer rules, and bounds shared-cap searches with regression coverage.

Validation passed for the morpho-sdk build, 15 facade tests, 124 review-engine tests, Biome, and package-map checks.


packages/morpho-sdk/src/entities/vaultV2BlueReallocationData.ts
```diff
@@ -1423,25 +1420,20 @@ export class VaultV2BlueReallocationData
...
-                if (probeReallocation(midpoint)?.withinAllCaps === true) {
-                  feasible = midpoint;
-                } else {
-                  infeasible = midpoint;
-                }
+            if (remainingAssets != null && probe?.withinAllCaps !== true) {
+              // ponytail: bounded base-unit scan; expose a caller ceiling if wider searches become necessary.
+              const searchUpper = MathLib.min(
+                lower,
+                selectedAssets + MAX_SHARED_CAP_SEARCH_STEPS,
```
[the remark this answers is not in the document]
Rubilmax: 1024 is fine for most tokens, even low-decimal, high-value loan tokens, discarded

Rubilmax:
@codex

packages/morpho-sdk/src/entities/vaultV2BlueReallocationData.ts
```diff
@@ -1364,21 +1366,15 @@ export class VaultV2BlueReallocationData
...
-                  ? undefined
-                  : postTransition.context.firstTotalAssets[
-                      firstTotalAssetsKey
-                    ]) ?? postVault._totalAssets;
+                postTransition.context.firstTotalAssets[
+                  findAddressKey(
+                    postTransition.context.firstTotalAssets,
+                    reallocation.vault,
+                  )!
+                ]!;
```
Foulks-Plb: These non-null assertions are safe *today* — `applyPublicReallocation` unconditionally populates `firstTotalAssets` under a key address-equal to `reallocation.vault` (see the `firstTotalAssets[firstTotalAssetsKey] == null` block below), so the lookup and value are always present post-transition. But the `!` gives no runtime guard: if that invariant is ever broken by a future refactor, `firstTotalAssets` becomes `undefined` and flows straight into `VaultV2Utils.allocationHeadroom(...)` bigint math, throwing a bare `TypeError` instead of a typed SDK error (against the spirit of the §2.2 / silent-failure rules). The removed `?? postVault._totalAssets` fallback was dead code, but consider keeping an explicit guard (or restoring a fallback) so the failure mode stays typed if the invariant regresses.

packages/morpho-sdk/src/entities/vaultV2BlueReallocationData.ts
```diff
@@ -1423,25 +1418,20 @@ export class VaultV2BlueReallocationData
...
-              let feasible = lower;
-              while (infeasible + 1n < feasible) {
-                const midpoint = (infeasible + feasible) / 2n;
-                if (probeReallocation(midpoint)?.withinAllCaps === true) {
-                  feasible = midpoint;
-                } else {
-                  infeasible = midpoint;
-                }
+            if (remainingAssets != null && probe?.withinAllCaps !== true) {
+              // ponytail: bounded base-unit scan; expose a caller ceiling if wider searches become necessary.
```
Foulks-Plb: `// ponytail:` looks like an internal codename/marker left in shipped source. Please remove or reword it to a plain explanatory comment. (The rest of the note — bounded base-unit scan, caller ceiling for wider searches — is good to keep.)

jinmel (approved):

PR #950, opened 2026-08-26
ci: refresh GitHub Actions and fix Foundry setup

Rubilmax:
## Motivation
CI started failing after Foundry changed foundryup from a shell script to a native binary, while other action pins had eligible updates.

## Solution
Pin foundry-toolchain v1.9.1 and refresh checkout, setup-node, cache, and pnpm setup to the latest releases at least 72 hours old.
All actions remain pinned to immutable SHAs with exact version comments.
Tests not run because changes are workflow YAML only; YAML parsing, SHA/tag/age checks, and git diff --check pass.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,13 +12,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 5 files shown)


Rubilmax (commented):
LGTM

florian-adm-morpho (approved):

PR #984, opened 2026-09-01
blue bundles for review

Foulks-Plb:
Fork of #971 (the `next` branch) rebased onto `main` for standalone review.

Contains the same modifications as #971 — the full **BlueBundlesV1** SDK actions migration — but based on `main` instead of the `next` integration branch, so the diff can be reviewed on its own.

Source: `next` → PR #971. Diff is byte-identical to `main...next` (113 files).

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/984" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


Foulks-Plb:
@codex review

Foulks-Plb (commented):
## Parallel PR Review (Claude — local:pr-review-gh)

**Reviewed commit:** `6ffe1e2`
**Scope:** 113 files, BlueBundlesV1 v5→v6 migration. 9 review agents ran (6 baseline + web3, release-integrity, dependencies); 0 failed.

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 1 |
| Medium | 2 |
| Low | 1 |

**4 new findings** since the last review of this PR.

**Came back clean** (no findings in lane): `web3` (Permit2 spender/uint256, chainId gating, LLTV buffer, calldata vs pinned ABI, penalty bounds, Action-layer purity all verified), `correctness`, `error-handling`, `tests` (fork + invariant coverage present), `docs` (JSDoc/TIB/pointers), `performance`, `dependencies` (peer-range widen is sound). The findings below are release-sequencing / DRY — no fund-safety or correctness defects surfaced.

_Automated parallel review — findings posted as inline comments with `event=COMMENT` (no approve / request-changes). Human verdict left to reviewers._

.changeset/blue-v2-only-reallocations.md
```diff
@@ -0,0 +1,14 @@
+---
+"@morpho-org/morpho-sdk": major
+"@morpho-org/wdk-protocol-lending-morpho-evm": major
+"@morpho-org/liquidity-sdk-viem": patch
+---
+
+Accept only Vault V2 BluePublicAllocator reallocations in high-level Morpho Blue write inputs.
```
Foulks-Plb, line 7: **[HIGH]** (confidence: 80%) [release-integrity] WHAT: The Vault V1 reallocation deprecation minor (deprecate-v1-blue-reallocations.md → morpho-sdk minor + wdk minor) is unconsumed in the SAME changeset set as this major (blue-v2-only-reallocations.md → morpho-sdk major + wdk major) that makes high-level Blue writes accept only Vault V2, so `changeset version` collapses morpho-sdk 5.8.0 straight to 6.0.0 (and wdk 1.1.0 to 2.0.0), the deprecation never publishes as a standalone minor, and V1-reallocation consumers get no deprecated-but-functional release window — violating root §7 ("Vault V1 reallocations ... must be deprecated in a published minor before the new majors accept only Vault V2") and TIB-2026-08-28 lines 21/81-88 ("The deprecation minor must publish before the BlueBundlesV1 major"; it explicitly rejects deleting the V1 surface in the major without a prior minor). FIX: Land deprecate-v1-blue-reallocations.md on its own and publish morpho-sdk 5.9.0 / wdk 1.2.0 to `latest` before consuming the BlueBundlesV1 major changesets, so the V1 high-level reallocation input ships deprecated-but-accepted in a published minor first.

packages/morpho-sdk/src/actions/blue/supplyCollateral.ts
```diff
@@ -1,126 +1,108 @@
...
+  params: BlueSupplyCollateralParams,
+): Readonly<Transaction<BlueSupplyCollateralAction>> => {
+  const transaction = blueSupplyCollateralBorrow({
+    ...params,
+    args: {
+      ...params.args,
+      borrowAssets: 0n,
+      maxLtv: maxUint256,
+    },
   });
```
Foulks-Plb, line 103: **[MEDIUM]** (confidence: 70%) [simplification] WHAT: The four simple pure builders — blueSupplyCollateral (supplyCollateral.ts:104), blueBorrow (borrow.ts:106), blueRepay (repay.ts:114) and blueWithdrawCollateral (withdrawCollateral.ts:99) — each end with an identical deepFreeze({ ...transaction, action: { ...transaction.action, type } }) re-type/re-freeze block that differs only by the literal action-type string (4 call sites, meeting root §1's >=3-call-site extraction threshold). FIX: Extract one retypeBlueBundlesV1Transaction(transaction, type) helper into actions/blue/common.ts and call it from all four builders.

packages/morpho-sdk/src/actions/blue/supplyCollateralBorrow.ts
```diff
@@ -1,230 +1,251 @@
...
-    throw new NegativeInputError("amount", amount);
+  } = params.args;
+  // Reject > uint256 so a direct caller gets the SDK's typed `InputExceedsMaxError`, not viem's
+  // `IntegerOutOfRangeError` at encode time. `maxUint256` stays valid as the `maxLtv` sentinel.
+  for (const [field, value] of [
+    ["collateralAssets", collateralAssets],
+    ["borrowAssets", borrowAssets],
+    ["maxLtv", maxLtv],
+  ] as const) {
+    if (value < 0n) throw new NegativeInputError(field, value);
```
Foulks-Plb: **[MEDIUM]** (confidence: 55%) [simplification] WHAT: The same uint256 field guard (`if (value < 0n) throw new NegativeInputError(field, value); if (value > maxUint256) throw new InputExceedsMaxError({ field, value, max: maxUint256 })`) is duplicated as the loop body in supplyCollateralBorrow.ts:149-151 and repayWithdrawCollateral.ts:154-156, and again inline in refinance.ts:145-151 (>=3 call sites). FIX: Add a validateUint256Field(field, value) helper in actions/blue/common.ts and call it from the field loops and refinance.

.changeset/validate-next-pipeline.md
```diff
@@ -0,0 +1,5 @@
+---
+"@morpho-org/morpho-sdk": patch
+---
+
+Validate the `next` prerelease pipeline end-to-end after making pending-changeset detection prerelease-aware. No runtime behavior change.
```
Foulks-Plb: **[LOW]** (confidence: 55%) [release-integrity] WHAT: validate-next-pipeline.md is a morpho-sdk patch changeset that self-describes as "No runtime behavior change" with no backing published-source change in this PR, so it will emit a phantom "Validate the next prerelease pipeline..." line in morpho-sdk's public 6.0.0 changelog — contrary to root §7 ("Do not generate a changeset for repo metadata, non-API ... changes"). FIX: If it was only a `next`-pipeline validation artifact, remove it before the stable major release so it does not leak into the public changelog; keep it only if it represents an actual release-tooling change consumers should see.

.changeset/blue-bundles-position-actions.md
```diff
@@ -0,0 +1,23 @@
...
+and inconsistent withdrawal-only funding inputs in the combined builders with the SDK's typed
+`InputExceedsMaxError`. Forward a caller-supplied reusable `approvalAmount` from
+`getRequirements(...)` on the Blue collateral-supply and repay prerequisite paths (previously
+dropped), while keeping the saturated-repay token cap. Mark the new Blue action argument shapes, the
+combined-builder parameter interfaces, the `BlueActions` entity write-method parameter shapes (and
+the shared `AssetsOrSharesArgs`), the `BlueTokenRequirementsParams` prerequisite options, and the
+WDK Blue-write option types `readonly`.
+
+Patch `@morpho-org/liquidity-sdk-viem` as a maintained direct dependent of the `morpho-sdk` major
+(its `morpho-sdk` peer range already accepts `^6.0.0`); this is the explicit dependent bump the
```
0xbulma: **[LOW]** (confidence: 88%) [release-integrity] WHAT: The changeset body asserts liquidity-sdk-viem's '`morpho-sdk` peer range already accepts `^6.0.0`', which is false - on `main` the range is `^5.4.0`, and the widening to `^5.4.0 || ^6.0.0` is performed by this same PR via `.changeset/blue-v2-only-reallocations.md`. Changeset bodies are published verbatim into `CHANGELOG.md`, so the v6.0.0 release notes will contain a wrong statement about the dependent-bump audit that section 7 requires reviewers to be able to verify from the changelog.

Suggestion: Reword to state that the peer range is widened to `^5.4.0 || ^6.0.0` in this release (cross-referencing the `blue-v2-only-reallocations` changeset) rather than claiming it already accepted `^6.0.0`.

.changeset/deprecate-v1-blue-reallocations.md
```diff
@@ -0,0 +1,9 @@
+---
+"@morpho-org/morpho-sdk": minor
```
0xbulma, line 2: **[HIGH]** (confidence: 92%) [docs, release-integrity] WHAT: The Vault V1 deprecation minor and the V2-only major ship in the same release. `deprecate-v1-blue-reallocations.md` declares `@morpho-org/morpho-sdk: minor` + `wdk-protocol-lending-morpho-evm: minor`, while `blue-v2-only-reallocations.md` declares `major` for the same two packages; with no `.changeset/pre.json` in the repo, a single `changeset version` run collapses both into one bump, so morpho-sdk goes 5.x -> 6.0.0 and WDK 1.x -> 2.0.0 with no published minor ever carrying the `@deprecated` Vault V1 surfaces. Root AGENTS.md section 7 explicitly excludes Vault V1 reallocations from the BlueBundlesV1 route exception: 'their high-level Blue and WDK flows must be deprecated in a published minor before the new majors accept only Vault V2 reallocations.' Both TIBs restate the constraint verbatim (TIB-2026-08-28 Assumptions: 'The deprecation minor must publish before the BlueBundlesV1 major').

Suggestion: Split the release - land `deprecate-v1-blue-reallocations.md` (and the source changes it describes) on its own PR, publish the morpho-sdk / WDK deprecation minor, then rebase this PR so only the major changesets remain pending; alternatively remove `blue-v2-only-reallocations.md`'s V2-only input narrowing from this PR and defer it to the release after the deprecation minor is on npm.

0xbulma (commented):
## Parallel PR Review (Claude — local:pr-review-gh)

**Reviewed commit:** `6ffe1e21`
**Scope:** 113 file(s), `26014581..6ffe1e21` (recomputed merge-base; 0 merge commits in range)
**Panel:** 8 agents — 6 baseline + `web3` (HAS_WEB3) + `release-integrity` (HAS_RELEASE). 0 failed.

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 2 |
| Medium | 10 |
| Low | 9 |

All 21 findings are new since the last review of this PR (first review). 0 findings were dropped by the engine's scope filter.

### Themes

- **Release sequencing (high).** The Vault V1 deprecation minor and the V2-only major are in the same pending changeset batch, so the minor never publishes on its own — the one case root `AGENTS.md` §7 carves *out* of the BlueBundlesV1 route exception.
- **`encodeErc20Permit2TransferFrom` guards (high/medium).** The new public encoder is missing both the Permit2-address check and the expired-deadline check its sibling encoders and the resolver already have.
- **Double-encode in `blue.ts` (medium/low).** `repay()`/`withdrawCollateral()` build a combined transaction, discard it, and rebuild — three agents converged on this independently.

_Automated parallel review, `event=COMMENT` only — no approval or change request. Re-runs on new commits if `--watch` is active._

.changeset/validate-next-pipeline.md
0xbulma: **[LOW]** (confidence: 85%) [release-integrity] WHAT: `.changeset/validate-next-pipeline.md` declares a `@morpho-org/morpho-sdk: patch` bump for release-pipeline validation, and its own body says 'No runtime behavior change' - it corresponds to no published package source change (the prerelease-detection fix lives in `scripts/release/detect-pending-changesets.mjs`, repo tooling). Root AGENTS.md section 7 forbids changesets for repo-metadata / tooling-only changes; this one will emit a meaningless entry in the morpho-sdk 6.0.0 CHANGELOG. It appears to be a leftover artifact carried over from validating the `next` branch pipeline in PR #971.

Suggestion: Delete `.changeset/validate-next-pipeline.md`.

packages/liquidity-sdk-viem/package.json
```diff
@@ -31,7 +31,7 @@
   "peerDependencies": {
     "@morpho-org/blue-sdk": "^6.0.0",
     "@morpho-org/blue-sdk-viem": "^5.0.0",
-    "@morpho-org/morpho-sdk": "^5.4.0",
+    "@morpho-org/morpho-sdk": "^5.4.0 || ^6.0.0",
```
0xbulma, line 34: **[MEDIUM]** (confidence: 75%) [release-integrity] WHAT: `@morpho-org/liquidity-sdk-viem` widens its `@morpho-org/morpho-sdk` peer range from `^5.4.0` to `^5.4.0 || ^6.0.0` but is only bumped `patch` in `.changeset/blue-bundles-position-actions.md` and `.changeset/blue-v2-only-reallocations.md`. Accepting a new peer major is additive compatibility surface, which root AGENTS.md section 7 classifies as `minor` ('Patch: bug fixes and internal source maintenance changes. Minor: additive surface'), and the package's own precedent agrees - CHANGELOG 4.1.0 shipped the narrower `morpho-sdk` peer move to `^5.4.0` as a Minor Change. Section 7's dependent-bump clause also requires the peer dependent be included 'with the appropriate bump', not a default patch.

Suggestion: Change the `@morpho-org/liquidity-sdk-viem` entry to `minor` in `.changeset/blue-v2-only-reallocations.md` (the changeset that actually performs the widening) so the published release signals the broadened peer contract.

packages/morpho-sdk/AGENTS.md
```diff
@@ -23,34 +34,50 @@ Protocol terms used across this package's docs and JSDoc:
...
-- **PublicAllocator V1** — MetaMorpho allocator that moves liquidity from one or more sorted source markets into a target via `reallocateTo(...)`; each call pays one `fee`.
-- **BluePublicAllocator** — the single canonical Vault V2 allocator registered per chain, which moves one source market or vault idle liquidity into the enclosing Blue action's target market via `reallocate(...)` or `allocateFromIdle(...)`. The caller supplies adapter addresses; the SDK resolves the allocator from the chain registry. Each call passes the vault's configured WAD-scaled `uint64 penalty`; the allocator pulls `ceil(assets × penalty / WAD)` of the target loan token from Bundler3 and donates it directly to the vault. Its canonical ABI export is `vaultV2BluePublicAllocatorAbi`.
+- **GeneralAdapter1** — the bundler-side adapter that holds approvals and executes composed calls.
+  Vault deposits and public low-level Bundler3 primitives use it; the high-level Blue write
+  methods do not.
+- **BlueBundlesV1** — the protocol-owned periphery called directly by the high-level Blue
```
0xbulma, line 44: **[LOW]** (confidence: 75%) [docs] WHAT: `BlueBundlesV1` is introduced here as a canonical glossary term and becomes the primary Blue routing term across every package doc, but root `AGENTS.md` section 6 still enumerates the glossary inventory as (`LLTV buffer`, `wNative`, `GeneralAdapter1`, `bundler3`, `PublicAllocator V1`, `BluePublicAllocator`, `MetaMorpho`, `Permit2`, `WAD`) with no `BlueBundlesV1`, leaving the spec's term list out of sync with the glossary it points at.

Suggestion: add `BlueBundlesV1` to the protocol-specific term list in root `AGENTS.md` section 6.

packages/morpho-sdk/BUNDLER3.md
```diff
@@ -1,128 +1,126 @@
...
+exposing intermediate state.
+
+### One spender for Bundler3 flows
+
+Vault deposits and advanced GeneralAdapter1 compositions share the same requirement resolver,
+[`getGeneralAdapterRequirements`](src/actions/requirements/generalAdapter/getGeneralAdapterRequirements.ts).
+An allowance to GeneralAdapter1 can be reused by those flows. It is not reused by direct
+BlueBundlesV1 writes because they have a different spender.
+
+## Pitfalls
```
0xbulma, line 104: **[MEDIUM]** (confidence: 75%) [docs] WHAT: The BUNDLER3.md rewrite deletes the former 'Withdraw / redeem do NOT go through Bundler3' caveat - that vault `withdraw`/`redeem` get no on-chain `minSharePrice`/`maxSharePrice` check and stay exposed to share-price manipulation between transaction construction and inclusion - even though those flows are unchanged in v6 (the routes table at line 45 still lists them as direct vault calls) and `ARCHITECTURE.md` keeps only the 'no inflation attack surface' rationale, not the trade-off, so the warning now exists nowhere in the repo.

Suggestion: re-add a Pitfalls bullet stating that direct vault `withdraw`/`redeem` carry no on-chain share-price bound and that callers must weigh share-price movement between build and inclusion.

packages/morpho-sdk/src/actions/blue/common.ts
```diff
@@ -0,0 +1,480 @@
...
+}
+
+/** @internal */
+export interface NormalizedBlueBundlesV1CommonParams {
+  referralFeePct: bigint;
+  referralFeeRecipient: Address;
+}
+
+interface BlueBundlesV1TokenPermit {
+  kind: 0 | 1 | 2;
```
0xbulma: **[MEDIUM]** (confidence: 70%) [correctness] WHAT: The BlueBundlesV1 token-permit discriminator is encoded as bare magic numbers - `kind: 0 | 1 | 2` on the interface, `kind: 0` in `EMPTY_TOKEN_PERMIT` (line 79), `kind: 2` for Permit2 SignatureTransfer (line 211) and `kind: 1` for ERC-2612 (line 262) - with no named constant, so transposing 1 and 2 produces perfectly valid calldata that routes the permit to the wrong contract branch and is invisible in review; AGENTS.md section 3 asks for `as const` + `satisfies` on protocol literals.

Suggestion: declare `const BLUE_BUNDLES_V1_PERMIT_KIND = { none: 0, erc2612: 1, permit2TransferFrom: 2 } as const` and derive the interface's `kind` type from it, then use the named members at the three construction sites.

packages/morpho-sdk/src/actions/blue/common.ts
```diff
@@ -0,0 +1,480 @@
...
+      expected: "permit or permit2TransferFrom",
+      actual: requirementSignature.action.type,
+    });
+  }
+
+  const serializedSignature = requirementSignature.args.signature;
+  let parsed: Signature;
+  try {
+    parsed =
+      size(serializedSignature) === 64
```
0xbulma, line 251: **[LOW]** (confidence: 90%) [tests] WHAT: `getBlueBundlesV1TokenPermit` (line 238) and `getBlueBundlesV1SignedAuthorization` (line 336) both branch on `size(serializedSignature) === 64` to accept an ERC-2098 compact signature, but no test ever exercises that branch - every blue action unit test builds its fixture with `serializeSignature({ r, s, yParity })` (65 bytes, e.g. supply.test.ts:177, withdraw.test.ts:184, refinance.test.ts:66) and the fork tests sign through `signAndVerifyTypedData` -> viem `signTypedData`, which also returns 65 bytes. A regression in the `parseCompactSignature`/`compactSignatureToSignature` -> `v` derivation would silently encode a wrong `v` into the BlueBundlesV1 permit/authorization struct with no failing test.

Suggestion: add a case to `supply.test.ts` (token permit) and `withdraw.test.ts` (signed authorization) that passes a 64-byte compact signature built with viem's `signatureToCompactSignature` + `serializeCompactSignature` and asserts the decoded `v`/`r`/`s` match the 65-byte equivalent.

packages/morpho-sdk/src/actions/blue/common.ts
```diff
@@ -0,0 +1,480 @@
...
+    throw new BlueBundlesV1RequirementSignatureMismatchError({
+      field: "deadline",
+      expected: String(authorizationSignature.args.deadline),
+      actual: String(authorizationSignature.action.args.deadline),
+    });
+  }
+
+  const serializedSignature = authorizationSignature.args.signature;
+  let parsed: Signature;
+  try {
```
0xbulma, line 347: **[MEDIUM]** (confidence: 94%) [simplification] WHAT: `getBlueBundlesV1TokenPermit` (lines 234-260) and `getBlueBundlesV1SignedAuthorization` (lines 332-358) contain a verbatim copy of the same ~27-line block - compact-vs-serialized signature parse in a try/catch, then `v ?? (yParity == null ? undefined : BigInt(yParity + 27))` with a null guard - and `actions/signatures/getVaultExitBundlesV1PermitStruct.ts:96-121` already holds a third near-identical copy, so root AGENTS.md section 1's three-call-site bar for extracting a helper is met.

Suggestion: extract one local helper in `common.ts` (e.g. `parseSignatureWithV(serialized)` returning `{ r, s, v }`) that takes the error factory or throws a shared typed error, and call it from both functions instead of duplicating the parse-and-normalize block.

packages/morpho-sdk/src/actions/blue/common.ts
```diff
@@ -0,0 +1,480 @@
...
+      marketParams: targetMarketParams,
+      fromIdle: reallocation.from.type === "idle",
+      sourceAdapter:
+        reallocation.from.type === "market"
+          ? reallocation.from.adapter
+          : zeroAddress,
+      sourceMarketParams:
+        reallocation.from.type === "market"
+          ? reallocation.from.marketParams
+          : MarketParams.idle(targetMarketParams.loanToken),
```
0xbulma: **[LOW]** (confidence: 85%) [performance] WHAT: `getBlueBundlesV1PublicAllocations` calls `MarketParams.idle(targetMarketParams.loanToken)` inside the per-reallocation `.map`, so every idle-source entry constructs a new `MarketParams`, which eagerly recomputes the same market id (`encodeAbiParameters` + `keccak256`) and liquidation-incentive factor; the helper is also invoked twice per operation (entity validation and again at `buildTx` encode time).

Suggestion: hoist `const idleMarketParams = MarketParams.idle(targetMarketParams.loanToken)` above the `.map` and reference it inside, so the idle params are hashed once per call rather than once per reallocation.

packages/morpho-sdk/src/actions/blue/common.ts
```diff
@@ -0,0 +1,480 @@
...
+      authorization: accepts.authorization,
+    });
+  if (permit?.action.type === "permit2") {
+    throw new BlueBundlesV1RequirementSignatureMismatchError({
+      field: "type",
+      expected: "permit or permit2TransferFrom",
+      actual: permit.action.type,
+    });
+  }
+  if (permit != null && permit2TransferFrom != null) {
```
0xbulma, line 467: **[LOW]** (confidence: 88%) [tests] WHAT: `selectBlueBundlesV1RequirementSignatures` adds a BlueBundlesV1-specific guard (`permit != null && permit2TransferFrom != null` -> `AmbiguousRequirementSignaturesError`) that has no test; the generic `selectRequirementSignatures` only rejects duplicates of the *same* kind (action.test.ts covers `[permitSignature, permit2Signature]`), so this new branch - which prevents one of two competing token signatures being silently dropped from the single BlueBundlesV1 permit slot - is the only thing standing between a caller and a mis-funded bundle, and nothing asserts it.

Suggestion: add an `error: AmbiguousRequirementSignaturesError` case in `supply.test.ts` that calls `blueSupply` with both an ERC-2612 `permit` and a `permit2TransferFrom` signature in the same array and asserts `toThrow(AmbiguousRequirementSignaturesError)`.

packages/morpho-sdk/src/actions/blue/common.ts
```diff
@@ -0,0 +1,480 @@
...
+    throw new AmbiguousRequirementSignaturesError("permit", 2);
+  }
+  return {
+    token:
+      (permit as Erc2612RequirementSignature | undefined) ??
```
0xbulma, line 472: **[MEDIUM]** (confidence: 85%) [correctness] WHAT: `selectBlueBundlesV1RequirementSignatures` reaches for a type assertion - `(permit as Erc2612RequirementSignature | undefined) ?? permit2TransferFrom` - because TypeScript will not narrow `permit` through the nested-discriminant guard `if (permit?.action.type === "permit2")` two lines above; this is exactly the escape hatch AGENTS.md section 3 ('Hard-to-type APIs are the wrong shape - redesign before reaching for an escape hatch') tells reviewers to reject, and it silently keeps compiling if `PermitRequirementSignature` later gains a third member.

Suggestion: replace the inline guard + cast with a real type predicate next to the existing `isPermitSignature` helper (e.g. `const isErc2612Signature = (s: PermitRequirementSignature): s is Erc2612RequirementSignature => s.action.type === "permit"`), then narrow with `if (permit != null && !isErc2612Signature(permit)) throw ...` so the assignment needs no `as`.

packages/morpho-sdk/src/actions/blue/supplyCollateralBorrow.ts
```diff
@@ -1,230 +1,251 @@
...
+  const reallocations = [...(params.args.reallocations ?? [])];
+  if (borrowAssets === 0n && reallocations.length > 0) {
+    throw new ReallocationsRequireBorrowError();
   }
-
-  if (minSharePrice < 0n) {
-    throw new NegativeInputError("minSharePrice", minSharePrice);
+  if (collateralAssets === 0n && requirementSignature != null) {
+    throw new UnexpectedRequirementSignatureError(
+      requirementSignature.action.type === "permit2TransferFrom"
```
0xbulma: **[LOW]** (confidence: 93%) [simplification] WHAT: `requirementSignature.action.type === "permit2TransferFrom" ? "permit2TransferFrom" : "permit"` is an identity mapping - `requirementSignature` is a `BlueBundlesV1TokenRequirementSignature`, whose `action.type` is exactly `"permit" | "permit2TransferFrom"` (types/action.ts:691-693) and is already assignable to `RequirementSignatureKind` (types/error.ts:681-686) - and the same dead ternary is repeated at `repayWithdrawCollateral.ts:184` and `common.ts:148`.

Suggestion: pass `requirementSignature.action.type` (resp. `params.requirementSignature.action.type`) directly to `new UnexpectedRequirementSignatureError(...)` at all three sites.

packages/morpho-sdk/src/actions/requirements/blue/getBlueAuthorizationRequirement.ts
```diff
@@ -100,17 +155,18 @@ export const getBlueAuthorizationRequirement = async (params: {
...
+      deadline: params.deadline,
     });
   }
 
   const isAuthorized = await pc.readContract({
     address: morpho,
     abi: blueAbi,
     functionName: "isAuthorized",
-    args: [userAddress, generalAdapter1],
+    args: [userAddress, authorized],
```
0xbulma, line 169: **[LOW]** (confidence: 72%) [tests] WHAT: the deleted `test/actions/blue/authorization.integration.test.ts` was the only fork test asserting the surviving `isAuthorized -> return null` branch (lines 153-154 and 172-173) against real Morpho state - it granted the authorization on-chain and re-called `getRequirements()` expecting length 0. The behaviour survives the BlueBundlesV1 migration (only the operator changed from GeneralAdapter1 to BlueBundlesV1), but the replacement coverage is mock-only (`getBlueAuthorizationRequirement.test.ts:33`); the new fork tests satisfy the requirement once and never re-query, so an operator-address or read-argument regression would pass every fork test.

Suggestion: in `test/entities/blue/blueBundlesV1.integration.test.ts`, after `satisfyBlueBundlesV1Requirements` sends the authorization transaction, re-call `action.getRequirements()` and assert no `isRequirementBlueAuthorization` entry remains.

packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Permit2TransferFrom.ts
```diff
@@ -0,0 +1,154 @@
...
+ */
+export const encodeErc20Permit2TransferFrom = (
+  params: EncodeErc20Permit2TransferFromParams,
+): Requirement<Permit2TransferFromRequirementSignature> => {
+  const { token, spender, amount, chainId, nonce, deadline } = params;
+  // Bound every uint256 field before signing: getPermit2TransferFromTypedData silently clamps an
+  // oversized allowance to maxUint256, which would desync the signed value from the requirement
+  // metadata that blueSupply verifies. Direct callers bypass the token-requirement resolver, so
+  // this public encoder must reject the same out-of-range inputs itself.
+  if (amount < 0n) {
```
0xbulma: **[MEDIUM]** (confidence: 90%) [simplification] WHAT: the `if (x < 0n) throw new NegativeInputError(...)` / `if (x > maxUint256) throw new InputExceedsMaxError({ field, value, max: maxUint256 })` pair is copy-pasted 15 times across 10 files in this diff (three times back-to-back here for `amount`/`nonce`/`deadline`, plus `common.ts:91-100`, `supply.ts:108-117`, `withdraw.ts:132-145`, `refinance.ts:147-156`, `getBlueBundlesV1TokenRequirements.ts:117-138,183-192`, `getBlueAuthorizationRequirement.ts:113-131`, `encodeErc20Permit.ts:95-105`), and two files already hand-roll a local `for (const [field, value] of [...] as const)` loop over it (`supplyCollateralBorrow.ts:144-153`, `repayWithdrawCollateral.ts:149-158`) - well past root AGENTS.md section 1's three-call-site bar.

Suggestion: add `validateUint256(field, value)` (and a `validateDeadline(deadline, now)` wrapping the non-positive / over-max / expired triple that `getBlueAuthorizationRequirement`, `encodeErc20Permit`, and `getBlueBundlesV1TokenRequirements` each repeat) to the existing `src/helpers/validate.ts` and call it from these sites.

packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Permit2TransferFrom.ts
```diff
@@ -0,0 +1,154 @@
...
+    throw new NegativeInputError("nonce", nonce);
+  }
+  if (nonce > maxUint256) {
+    throw new InputExceedsMaxError({
+      field: "nonce",
+      value: nonce,
+      max: maxUint256,
+    });
+  }
+  if (deadline <= 0n) {
```
0xbulma, line 86: **[MEDIUM]** (confidence: 85%) [web3] WHAT: `encodeErc20Permit2TransferFrom` bounds `deadline` only with `deadline <= 0n` (NonPositiveInputError) and `deadline > maxUint256` (InputExceedsMaxError), but never rejects a deadline that is already in the past - so a direct caller can be walked through a wallet EIP-712 prompt for a Permit2 SignatureTransfer that Permit2 will reject with `SignatureExpired`. Its sibling encoder `encodeErc20Permit` gained exactly this guard in the same PR (`if (params.deadline <= now) throw new ExpiredDeadlineError(params.deadline, now)`), and `getBlueBundlesV1TokenRequirements` also enforces it, so the SignatureTransfer encoder is the only signing entry point in this route without it - contradicting its own in-file comment that direct callers bypass the resolver and must be rejected here.

Suggestion: Add `const timestamp = Time.timestamp(); if (deadline <= timestamp) throw new ExpiredDeadlineError(deadline, timestamp);` alongside the existing deadline bounds checks (importing `Time` from `@morpho-org/morpho-ts` and `ExpiredDeadlineError` from `../../../types/index.js`) and document `@throws {ExpiredDeadlineError}` in the JSDoc.

packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Permit2TransferFrom.ts
```diff
@@ -0,0 +1,154 @@
...
+    throw new NonPositiveInputError("deadline", deadline);
+  }
+  if (deadline > maxUint256) {
+    throw new InputExceedsMaxError({
+      field: "deadline",
+      value: deadline,
+      max: maxUint256,
+    });
+  }
+  validateRequirementSpender({
```
0xbulma, line 103: **[HIGH]** (confidence: 85%) [web3, error-handling] WHAT: The encoder validates the spender against `bundles.blueBundlesV1` but never checks that the chain has a registered `permit2` address, while `getPermit2TransferFromTypedData` builds its EIP-712 domain as `verifyingContract: getChainAddresses(chainId).permit2`. Five registered chains (WorldChainMainnet, KatanaMainnet, HyperliquidMainnet, MonadMainnet, StableMainnet in packages/morpho-ts/src/addresses.ts) declare `bundles.blueBundlesV1` with no `permit2` entry, so on those chains `verifyingContract` is `undefined`; viem's `getTypesForEIP712Domain` filters undefined domain fields out entirely, so the wallet signs `EIP712Domain(string name,uint256 chainId)` instead of the Permit2 domain. `signAndVerifyTypedData` cannot catch this because its `verifyTypedData` check reuses the same malformed domain and recovers the signer correctly, so the caller receives a `Permit2TransferFromRequirementSignature` that is silently bound to the wrong domain separator and is unusable by Permit2 on-chain. `getBlueBundlesV1TokenRequirements` gates on `permit2 != null`, but this exported encoder is public API (`src/actions/requirements/encode/index.ts`) and its own comment states direct callers must be rejected here.

Suggestion: In `encodeErc20Permit2TransferFrom`, resolve `getChainAddresses(chainId).permit2` before signing and throw a typed error (e.g. via `getChainAddress(chainId, "permit2")` so an unsupported chain raises `UnknownAddressError`, or a dedicated guard) when it is null, so a chain without a registered Permit2 deployment can never produce a domain-less signature; document it with a `@throws` tag.

packages/morpho-sdk/src/entities/blue/blue.ts
```diff
@@ -1105,799 +1461,567 @@ export class MorphoBlue implements BlueActions {
...
-          useSimplePermit: reqParams?.useSimplePermit,
-          args: { amount: erc20Amount, from: userAddress },
-        });
-      },
-
+      getRequirements: combinedAction.getRequirements,
       buildTx: (signatures?: readonly RequirementSignature[]) => {
-        const { permit } = selectRequirementSignatures(signatures, {
-          permit: true,
+        const combinedTransaction = combinedAction.buildTx(signatures);
```
0xbulma, line 1489: **[MEDIUM]** (confidence: 92%) [correctness, simplification, performance] WHAT: `repay().buildTx()` calls `combinedAction.buildTx(signatures)` and throws the result away, using it only to read `combinedTransaction.value` and `combinedTransaction.action.args`; because `blueRepay` internally re-invokes `blueRepayWithdrawCollateral` (repay.ts:105-116 is literally `deepFreeze({...transaction, action: {...transaction.action, type: "blueRepay"}})`), every `repay().buildTx()` runs the full `encodeFunctionData` + `deepFreeze` + signature-parse pipeline twice for the same calldata, and couples the wrapper to the combined action's runtime arg shape rather than a typed value.

Suggestion: have `repayWithdrawCollateral` expose the already-computed normalized values (`nativeValue`, `maxRepayAssets`, `repayAssets`, `repayShares`, `referralFeePct`, `referralFeeRecipient`, `deadline`) on its returned object so `repay().buildTx()` reads them directly and calls `blueRepay` once, instead of building and discarding a combined transaction.

packages/morpho-sdk/src/entities/blue/blue.ts
```diff
@@ -1105,799 +1461,567 @@ export class MorphoBlue implements BlueActions {
...
     return {
-      buildTx: () =>
-        blueWithdrawCollateral({
-          market: {
-            chainId: this.chainId,
-            marketParams: this.marketParams,
-          },
+      getRequirements: () => combinedAction.getRequirements(),
+      buildTx: (signatures?: readonly RequirementSignature[]) => {
+        const combinedTransaction = combinedAction.buildTx(signatures);
```
0xbulma, line 1554: **[LOW]** (confidence: 92%) [performance] WHAT: `withdrawCollateral().buildTx()` builds a full combined transaction via `combinedAction.buildTx(signatures)` purely to read `combinedTransaction.action.args`, then calls `blueWithdrawCollateral`, which delegates to `blueRepayWithdrawCollateral` again - doubling the `encodeFunctionData` + `deepFreeze` + authorization-signature parse work per call.

Suggestion: surface the normalized `collateralAssets`/`maxLtv`/`deadline`/`referralFee*` values from the `repayWithdrawCollateral` closure on the returned action object and pass them straight to `blueWithdrawCollateral`, dropping the discarded `combinedAction.buildTx(...)` call.

packages/wdk-protocol-lending-morpho-evm/src/morpho-protocol-evm.ts
```diff
@@ -685,43 +773,47 @@ export default class MorphoProtocolEvm extends LendingProtocol {
...
-  public async getBorrowRequirements(
-    options: MorphoBorrowInput,
-  ): Promise<
-    (
-      | RequirementApproval
-      | RequirementAuthorization
-      | RequirementSignatureRequest
-    )[]
-  > {
+  getBorrowRequirements(
```
0xbulma, line 813: **[MEDIUM]** (confidence: 92%) [simplification] WHAT: `getBorrowRequirements` now declares a single overload signature immediately followed by its implementation - vestigial scaffolding left over from the two-overload Vault V1/V2 form this PR deleted; a one-signature overload adds nothing but hides the implementation's return type from the public surface. The same single-overload shape was copied to `getRepayRequirements` (1000), `getSupplyCollateralRequirements` (1204), and `getWithdrawCollateralRequirements` (1415).

Suggestion: collapse each pair into one `async` method carrying the explicit `Promise<readonly ...Requirement[]>` return annotation and delete the standalone overload declaration.

packages/wdk-protocol-lending-morpho-evm/src/morpho-protocol-evm.ts
```diff
@@ -1007,43 +1320,166 @@ export default class MorphoProtocolEvm extends LendingProtocol {
...
-    options: WithdrawOptions,
+    options: MorphoWithdrawCollateralOptions,
     config?: Erc4337TransactionConfig,
   ): Promise<WithdrawResult> {
     this._assertWritable("withdrawCollateral(options)");
 
-    const tx = await this._getWithdrawCollateralTransaction(options);
+    const action = await this._getWithdrawCollateralAction(options);
+    const tx = toWdkTransaction(
+      action.buildTx(
```
0xbulma, line 1364: **[MEDIUM]** (confidence: 85%) [correctness] WHAT: This PR deletes the private `_getWithdrawCollateralTransaction` helper and inlines its body - `toWdkTransaction(action.buildTx(options.requirementSignature ? [options.requirementSignature] : undefined))` - verbatim into both `withdrawCollateral` (line 1364) and `quoteWithdrawCollateral` (line 1467), breaking the `_getXAction` + `_getXTransaction` pairing every other flow in this file still follows (`_getSupplyTransaction` line 644, `_getBorrowTransaction` line 900, `_getRepayTransaction` line 1092, `_getSupplyCollateralTransaction` line 1312) and taking the identical signature-to-array expression from four to six copies, well past the repo's three-call-site extraction threshold (AGENTS.md section 1 Modularity).

Suggestion: restore a `private async _getWithdrawCollateralTransaction(options: MorphoWithdrawCollateralOptions): Promise<WdkTransaction>` that wraps `_getWithdrawCollateralAction` plus the `buildTx` call, and have both `withdrawCollateral` and `quoteWithdrawCollateral` call it.

packages/morpho-sdk/src/entities/blue/blue.ts
```diff
@@ -1105,799 +1461,567 @@ export class MorphoBlue implements BlueActions {
...
+        value: deadline,
+        max: quoteHorizon,
       });
     }
-
-    const { position: positionAfterRepay } = positionData.repay(
-      repayAssets,
-      repayShares,
-      accrualTimestamp,
+    const accrualTimestamp = repayShares > 0n ? deadline : quoteHorizon;
```
jinmel, line 1663: `maxRepayAssets` is derived by forward-accruing the snapshot to `deadline` under the snapshot utilization and adaptive-IRM state. That is a forecast, not an upper bound: intervening utilization or rate-state changes can make the live saturated-share repayment plus referral fee exceed this encoded cap, so BlueBundlesV1 reverts even before the deadline. A reusable ERC-20 allowance does not help because the call cap remains unchanged, and exact ERC-2612 or Permit2 signatures cannot be enlarged. Please expose or derive funding headroom, size requirements against that cap, and add a test that changes market state between quote and execution. The migration guide should not claim the two-hour horizon alone guarantees sufficiency.

.changeset/blue-v2-only-reallocations.md
```diff
@@ -0,0 +1,14 @@
+---
+"@morpho-org/morpho-sdk": major
```
jinmel, line 2: This major removal and the `deprecate-v1-blue-reallocations` minor changeset are in the same release batch. Changesets therefore resolves these packages directly to a major release, so consumers never receive the required published deprecation minor. That contradicts the TIB prerequisite and the repository rule excluding Vault V1 reallocations from the route-migration exception. Please publish the prerequisite minor first, then land this removal in the subsequent major.

packages/morpho-sdk/src/entities/blue/blue.ts
```diff
@@ -1105,799 +1461,567 @@ export class MorphoBlue implements BlueActions {
...
-        available: positionData.borrowShares,
-      });
-    }
-
-    if (requestedAssets > positionData.borrowAssets) {
-      throw new RefinanceExceedsBorrowAssetsError({
-        market: this.marketParams.id,
-        requested: requestedAssets,
-        available: positionData.borrowAssets,
+    const reallocations = [...(params.reallocations ?? [])];
```
jinmel, line 1959: This copies only the iterable container; every nested reallocation object remains a caller-owned live reference. The entity computes penalty and destination health from the current values, but `buildTx()` later remaps the same references, so a mutation between action creation and building can change calldata without rerunning the corresponding entity validation. The withdraw and supply-collateral/borrow handles use the same pattern. Please capture normalized plain-object snapshots here and close over that immutable value throughout the lazy action.

PR #990, opened 2026-09-01
feat(morpho-sdk)!: BlueBundlesV1 — refinance (3/3)

Foulks-Plb:
**PR 3 of 3** — split of #984. **Stacked on PR 2**; review/merge after it.

## This PR
Migrates the last Blue write — `refinance` — to route directly through BlueBundlesV1, retires the shared v5 reallocation-encoding helper `buildReallocationActions`, removes the now-dead v5 refinance validators, and finalizes the migration docs and error surface.

## End state
After this PR the stack reproduces #984's reviewed migration (`main…blue-bundles-for-review`) **plus the review-driven follow-ups made since that review** — it is **no longer byte-identical** to `blue-bundles-for-review`. The delta on top of the reviewed snapshot (`git diff blue-bundles-for-review blue-bundles-refinance`; at time of writing `4defaca5 → 190e0ad0`, ~14 files / +854 −486) is **post-review hardening only**: the signature-`deadline` guard, Permit2 hardening, added unit + fork tests, and a TIB cleanup. Those lines need review — they were not part of #984's review.

## Chain
1. PR 1 (position + foundation) → `main` (#988)
2. PR 2 (supply/withdraw) → PR 1
3. **PR 3 (this) → PR 2**

> ⚠️ Diff shown vs PR 2's branch. Merge PR 1 then PR 2 first (or retarget as they land).

## Verification
`tsc --noEmit` (morpho-sdk + wdk), Biome, and the morpho-sdk unit suite green (grown beyond #984's 933 with the post-review tests noted above). Fork tests need RPC — not run here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/990" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

PR #997, opened 2026-09-02
refactor(morpho-sdk)!: share VaultBundlesV1 prerequisites

jinmel:
## Summary

Implements the shared prerequisites from #951 on top of `next`, which includes the merged BlueBundlesV1 stack (#990):

- rename Blue-specific approval/signature APIs to fixed-bundle names, retaining deprecated compatibility aliases where applicable
- add the shared immutable bundle funding, referral-fee, requirement-selection, and resolution primitives used by Blue and Vault actions
- add the VaultBundlesV1 ABI surface and reusable share-price/slippage helpers
- keep Blue actions on the same common implementation so Vault support does not create a second source of truth

## Stack

1. This PR — shared prerequisites
2. #998 — Vault V1/V2 deposits
3. #999 — Vault V1/V2 withdrawals
4. #1000 — Vault V1/V2 redemptions and Vault V1-to-V2 migration

## Verification

Revalidated while rebasing the stack onto `next`:

- SDK and WDK TypeScript checks passed on all four PR heads.
- This PR's SDK/WDK unit suites: 87 files / 1,041 tests passed.
- Combined stack: 39 tests passed across 8 pinned Anvil fork suites covering BlueBundlesV1 and vault deposits, withdrawals, redemptions, and V1-to-V2 migration.
- Repository Biome, JSDoc coverage self-check, address checks, and `git diff --check` passed.

## Release sequencing

This stack implements TIB #951, but must not merge into a releasable major until the required published-minor deprecation runway is resolved. The earlier runway PR #991 was closed without merge.

### DIFF
--- a/packages/morpho-sdk/src/abis.ts
+++ b/packages/morpho-sdk/src/abis.ts
@@ -1250,7 +1250,7 @@ export const generalAdapter1Abi = [
   },
   {
     type: "function",
-    name: "permit2SignatureTransfer",
+    name: "permit2TransferFrom",
     inputs: [
       {
         name: "token",
--- a/packages/morpho-sdk/src/abis.ts
+++ b/packages/morpho-sdk/src/abis.ts
@@ -218,6 +218,147 @@ export const vaultExitBundlesV1Abi = [
   },
 ] as const satisfies Abi;
 
+/**
+ * Pinned ABI for the VaultBundlesV1 entrypoints used by Vault V1 and Vault V2 actions.
+ *
+ * Sourced from `morpho-org/bundles` commit
+ * `f27e7bcf744310303e24faa522b71d702e696686`.
+ */
+export const vaultBundlesV1Abi = [
+  { type: "error", name: "AlreadyInitiated", inputs: [] },
+  { type: "error", name: "DeadlinePassed", inputs: [] },
+  { type: "error", name: "InconsistentAssets", inputs: [] },
+  { type: "error", name: "NotExactlyOneZero", inputs: [] },
+  { type: "error", name: "PctExceeded", inputs: [] },
+  { type: "error", name: "SlippageExceeded", inputs: [] },
+  {
+    type: "function",
+    name: "initiator",
+    inputs: [],
+    outputs: [{ name: "", type: "address", internalType: "address" }],
+    stateMutability: "view",
+  },
+  {
+    type: "function",
+    name: "vaultBundlesV1Deposit",
+    inputs: [
+      { name: "vault", type: "address", internalType: "address" },
+      { name: "assets", type: "uint256", internalType: "uint256" },
+      {
+        name: "maxSharePriceE27",
+        type: "uint256",
+        internalType: "uint256",
+      },
+      {
+        name: "assetPermit",
+        type: "tuple",
+        internalType: "struct TokenPermit",
+        components: [
+          { name: "kind", type: "uint8", internalType: "enum PermitKind" },
+          { name: "data", type: "bytes", internalType: "bytes" },
+        ],
+      },
+      {
+        name: "referralFeePct",
+        type: "uint256",
+        internalType: "uint256",
+      },
+      {
+        name: "referralFeeRecipient",
+        type: "address",
+        internalType: "address",
+      },
+      { name: "deadline", type: "uint256", internalType: "uint256" },
+    ],
+    outputs: [],
+    stateMutability: "payable",
+  },
+  {
+    type: "function",
+    name: "vaultBundlesV1Withdraw",
+    inputs: [
+      { name: "vault", type: "address", internalType: "address" },
+      { name: "assets", type: "uint256", internalType: "uint256" },
+      { name: "shares", type: "uint256", internalType: "uint256" },
+      {
+        name: "sharesPermit",
+        type: "tuple",
+        internalType: "struct Permit",
+        components: [
+          { name: "value", type: "uint256", internalType: "uint256" },
+          { name: "nonce", type: "uint256", internalType: "uint256" },
+          { name: "deadline", type: "uint256", internalType: "uint256" },
+          { name: "v", type: "uint8", internalType: "uint8" },
+          { name: "r", type: "bytes32", internalType: "bytes32" },
+          { name: "s", type: "bytes32", internalType: "bytes32" },
+        ],
+      },
+      {
+        name: "referralFeePct",
+        type: "uint256",
+        internalType: "uint256",
+      },
+      {
+        name: "referralFeeRecipient",
+        type: "address",
+        internalType: "address",
+      },
+      { name: "deadline", type: "uint256", internalType: "uint256" },
+    ],
+    outputs: [],
+    stateMutability: "nonpayable",
+  },
+  {
+    type: "function",
+    name: "vaultBundlesV1Migrate",
+    inputs: [
+      { name: "sourceVault", type: "address", internalType: "address" },
+      { name: "destVault", type: "address", internalType: "address" },
+      {
+        name: "assetsWithdrawn",
+        type: "uint256",
+        internalType: "uint256",
+      },
+      {
+        name: "sharesRedeemed",
+        type: "uint256",
+        internalType: "uint256",
+      },
+      {
+        name: "destMaxSharePriceE27",
+        type: "uint256",
+        internalType: "uint256",
+      },
+      {
+        name: "sharesPermit",
+        type: "tuple",
+        internalType: "struct Permit",
+        components: [
+          { name: "value", type: "uint256", internalType: "uint256" },
+          { name: "nonce", type: "uint256", internalType: "uint256" },
+          { name: "deadline", type: "uint256", internalType: "uint256" },
+          { name: "v", type: "uint8", internalType: "uint8" },
+          { name: "r", type: "bytes32", internalType: "bytes32" },
+          { name: "s", type: "bytes32", internalType: "bytes32" },
+        ],
+      },
+      {
+        name: "referralFeePct",
+        type: "uint256",
+        internalType: "uint256",
+      },
+      {
+        name: "referralFeeRecipient",
+        type: "address",
+        internalType: "address",
+      },
+      { name: "deadline", type: "uint256", internalType: "uint256" },
+    ],
+    outputs: [],
+    stateMutability: "nonpayable",
+  },
+] as const satisfies Abi;
+
 /** ABI for the Bundler3 multicall contract. */
 export const bundler3Abi = [
   {
@@ -1109,7 +1250,7 @@ export const generalAdapter1Abi = [
   },
   {
     type: "function",
-    name: "permit2TransferFrom",
+    name: "permit2SignatureTransfer",
     inputs: [
       {
         name: "token",
--- a/packages/morpho-sdk/src/actions/bundles/common.test.ts
+++ b/packages/morpho-sdk/src/actions/bundles/common.test.ts
@@ -1,0 +1,197 @@
+import fc from "fast-check";
+import { maxUint256, serializeSignature, toHex, zeroHash } from "viem";
+import { describe, expect, test } from "vitest";
+import {
+  type BundlesFundingArgs,
+  BundlesPermitMismatchError,
+  MixedBundlesFundingError,
+  NegativeInputError,
+  NonPositiveInputError,
+  type PermitRequirementSignature,
+} from "../../types/index.js";
+import {
+  getBundlesSharesPermit,
+  resolveBundlesFunding,
+  selectBundlesSharesRequirementSignature,
+  selectBundlesTokenRequirementSignature,
+} from "./common.js";
+
+const owner = "0x0000000000000000000000000000000000000001" as const;
+const vault = "0x0000000000000000000000000000000000000002" as const;
+const spender = "0x0000000000000000000000000000000000000003" as const;
+const signature = serializeSignature({
+  r: toHex(1n, { size: 32 }),
+  s: toHex(2n, { size: 32 }),
+  yParity: 0,
+});
+
+describe("resolveBundlesFunding", () => {
+  test("behavior: preserves exclusive token and native amounts", () => {
+    fc.assert(
+      fc.property(
+        fc.bigInt({ min: 1n, max: maxUint256 }),
+        fc.boolean(),
+        (assets, native) => {
+          const funding: BundlesFundingArgs = native
+            ? { nativeAmount: assets }
+            : { amount: assets };
+          expect(resolveBundlesFunding(funding)).toEqual({
+            assets,
+            value: native ? assets : 0n,
+          });
+        },
+      ),
+      { numRuns: 100, seed: 20_260_909 },
+    );
+  });
+
+  test("error: typed runtime boundaries", () => {
+    expect(() =>
+      resolveBundlesFunding({
+        amount: 1n,
+        nativeAmount: 1n,
+      } as unknown as BundlesFundingArgs),
+    ).toThrow(MixedBundlesFundingError);
+    expect(() => resolveBundlesFunding({ amount: -1n })).toThrow(
+      NegativeInputError,
+    );
+    expect(() => resolveBundlesFunding({ nativeAmount: 0n })).toThrow(
+      NonPositiveInputError,
+    );
+  });
+});
+
+describe("getBundlesSharesPermit", () => {
+  const permit = {
+    args: {
+      owner,
+      asset: vault,
+      amount: 7n,
+      nonce: 9n,
+      deadline: 11n,
+      signature,
+    },
+    action: {
+      type: "permit",
+      args: { spender, amount: 7n, deadline: 11n, nonce: 9n },
+    },
+  } satisfies PermitRequirementSignature;
+
+  test("default", () => {
+    expect(getBundlesSharesPermit({ vault, deadline: 13n })).toEqual({
+      value: 0n,
+      nonce: 0n,
+      deadline: 13n,
+      v: 0,
+      r: zeroHash,
+      s: zeroHash,
+    });
+    expect(
+      getBundlesSharesPermit({
+        vault,
+        owner,
+        spender,
+        amount: 7n,
+        deadline: 13n,
+        requirementSignature: permit,
+      }),
+    ).toMatchObject({ value: 7n, nonce: 9n, deadline: 11n, v: 27 });
+  });
+
+  test("error: BundlesPermitMismatchError", () => {
+    expect(() =>
+      getBundlesSharesPermit({
+        vault,
+        owner,
+        spender,
+        amount: 8n,
+        deadline: 13n,
+        requirementSignature: permit,
+      }),
+    ).toThrow(BundlesPermitMismatchError);
+  });
+});
+
+describe("selectBundlesSharesRequirementSignature", () => {
+  const permit = {
+    args: {
+      owner,
+      asset: vault,
+      amount: 7n,
+      nonce: 9n,
+      deadline: 11n,
+      signature,
+    },
+    action: {
+      type: "permit",
+      args: { spender, amount: 7n, deadline: 11n, nonce: 9n },
+    },
+  } satisfies PermitRequirementSignature;
+
+  test("default", () => {
+    expect(
+      selectBundlesSharesRequirementSignature([permit], {
+        requiredShareAllowance: 7n,
+        expectedRequirement: permit.action,
+      }),
+    ).toEqual(permit);
+  });
+
+  test("error: BundlesPermitMismatchError", () => {
+    expect(() =>
+      selectBundlesSharesRequirementSignature([permit], {
+        requiredShareAllowance: undefined,
+      }),
+    ).toThrow(BundlesPermitMismatchError);
+    expect(() =>
+      selectBundlesSharesRequirementSignature([permit], {
+        requiredShareAllowance: 8n,
+        expectedRequirement: permit.action,
+      }),
+    ).toThrow(BundlesPermitMismatchError);
+    expect(() =>
+      selectBundlesSharesRequirementSignature([permit], {
+        requiredShareAllowance: 7n,
+        expectedRequirement: {
+          ...permit.action,
+          args: { ...permit.action.args, nonce: 10n },
+        },
+      }),
+    ).toThrow(BundlesPermitMismatchError);
+  });
+});
+
+describe("selectBundlesTokenRequirementSignature", () => {
+  const permit = {
+    args: {
+      owner,
+      asset: vault,
+      amount: 7n,
+      nonce: 9n,
+      deadline: 11n,
+      signature,
+    },
+    action: {
+      type: "permit",
+      args: { spender, amount: 7n, deadline: 11n, nonce: 9n },
+    },
+  } satisfies PermitRequirementSignature;
+
+  test("default", () => {
+    expect(
+      selectBundlesTokenRequirementSignature([permit], permit.action),
+    ).toEqual(permit);
+  });
+
+  test("error: rejects signatures outside the prepared requirement", () => {
+    expect(() => selectBundlesTokenRequirementSignature([permit])).toThrow(
+      BundlesPermitMismatchError,
+    );
+    expect(() =>
+      selectBundlesTokenRequirementSignature([permit], {
+        ...permit.action,
+        args: { ...permit.action.args, deadline: 12n },
+      }),
+    ).toThrow(BundlesPermitMismatchError);
+  });
+});
--- a/packages/morpho-sdk/src/actions/bundles/common.test.ts
+++ b/packages/morpho-sdk/src/actions/bundles/common.test.ts
@@ -1,16 +1,39 @@
 import fc from "fast-check";
-import { maxUint256, serializeSignature, toHex, zeroHash } from "viem";
-import { describe, expect, test } from "vitest";
+import {
+  decodeAbiParameters,
+  maxUint256,
+  serializeCompactSignature,
+  serializeSignature,
+  signatureToCompactSignature,
+  toHex,
+  zeroHash,
+} from "viem";
+import { describe, expect, expectTypeOf, test } from "vitest";
+import type {
+  BundleSharesPermit,
+  BundlesSharesPermit,
+  VaultExitBundlesV1PermitStruct,
+} from "../../index.js";
 import {
   type BundlesFundingArgs,
   BundlesPermitMismatchError,
+  BundlesRequirementSignatureMismatchError,
+  DepositAmountMismatchError,
+  DepositAssetMismatchError,
+  DepositOwnerMismatchError,
+  DepositSpenderMismatchError,
+  type Erc2612RequirementSignature,
   MixedBundlesFundingError,
   NegativeInputError,
   NonPositiveInputError,
+  type Permit2AllowanceRequirementSignature,
+  type Permit2SignatureTransferRequirementSignature,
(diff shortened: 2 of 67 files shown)


packages/morpho-sdk/src/helpers/slippage.ts
```diff
@@ -206,3 +213,125 @@ export function computeMinWithdrawSharePrice(params: {
...
+  validateSlippageTolerance(params.slippageTolerance);
+  const accrualTimestamp =
+    params.vaultData instanceof AccrualVaultV2
+      ? MathLib.max(params.deadline, params.vaultData.lastUpdate)
+      : params.deadline;
+  const accruedVault =
+    params.vaultData instanceof AccrualVaultV2
+      ? params.vaultData.accrueInterest(accrualTimestamp).vault
+      : params.vaultData.accrueInterest(accrualTimestamp);
+  const shares = accruedVault.toShares(params.assets);
```
Foulks-Plb: 🟡 **MEDIUM — `computeVaultMaxSharePrice`: implicit `toShares` rounding is wrong for Vault V1**

```ts
const shares = accruedVault.toShares(params.assets); // ⚠️ no rounding direction
```

- `Vault.toShares` (V1 / MetaMorpho) defaults to `"Up"`, whereas `VaultV2.toShares` defaults to `"Down"` — so the bound silently depends on the vault version.
- The on-chain check (`vaultBundlesV1Deposit`, contract L55) is `ceil(toDeposit * 1e27 / shares) <= maxSharePriceE27` with `shares = vault.deposit(...)` → shares rounded **down** (ERC4626). To mirror the contract the preview must round **`"Down"`**.
- On V1, the `"Up"` default over-estimates shares → bound too tight → **false revert** when `slippageTolerance ≈ 0` on small amounts (at slippage 0, `ceil(net*1e27/sharesDown) > ceil(net*1e27/sharesUp)`). Impact: DoS / UX, not fund loss.
- Not covered by tests: `slippage.test.ts:360+` mocks `toShares` to identity, so the real rounding is never exercised.
- The sibling `computeVaultMaxShareAllowance` explicitly passes `"Up"` (L324-325) — the asymmetry confirms the omission here is unintentional.

**Fix:** `accruedVault.toShares(params.assets, "Down")` (+ a V1 test at `slippageTolerance = 0`). Consumed in #998, but worth fixing here since the helper is introduced in this PR.

Foulks-Plb (commented):
### Security / correctness review

Overall a clean, behavior-preserving extraction of the Blue path into a shared bundles core: the calldata encoding is byte-identical to the on-chain `TokenLib.pullToken` (ERC-2612 `(deadline,v,r,s)`, Permit2 `(nonce,deadline,signature)`), the signed EIP-712 messages commit spender/amount/nonce/deadline and the spender is validated against the Morpho address registry, and `grossFromNetAssets` is proven exact by property test. One MEDIUM correctness note is inline on `slippage.ts`; two LOW notes below.

---

🟢 **LOW — Stale JSDoc in a dependent (`wdk-protocol-lending-morpho-evm`)**

`packages/wdk-protocol-lending-morpho-evm/src/morpho-protocol-evm.ts:967-968,1172-1173` reference `MissingPermit2TransferFromNonceError` / `Permit2TransferFromNonceAlreadyUsedError`, which are now `@deprecated` aliases. JSDoc only → no compile break (the aliases still exist), so no wdk changeset is required. Worth cleaning up as part of the §4/§7 dependent audit.

---

🟢 **LOW — ERC-2612 nonce probe swallows all errors**

`getBundlesTokenRequirements` (`packages/morpho-sdk/src/entities/requirements/getBundlesTokenRequirements.ts:136`): `readContract(..., "nonces").catch(() => undefined)` falls back to Permit2 / classic approval even on a transient RPC error, not only on a genuinely non-2612 token. Pre-existing SDK pattern, not introduced by this PR — flagging only (silent failure).

jinmel:
Fixing:
- @devin-ai-integration: review summary — Devin Review** found 1 potential issue.
- @chatgpt-codex-connector: review summary — 💡 Codex Review
- @chatgpt-codex-connector: review summary — 💡 Codex Review
- @Foulks-Plb: review summary — Security / correctness review

Will be in the next commit.

jinmel:
@devin-ai-integration Fixing. Will be in the next commit

packages/morpho-sdk/src/helpers/slippage.ts
```diff
@@ -206,3 +213,126 @@ export function computeMinWithdrawSharePrice(params: {
...
+  validateSlippageTolerance(params.slippageTolerance);
+  const accrualTimestamp =
+    params.vaultData instanceof AccrualVaultV2
+      ? MathLib.max(params.deadline, params.vaultData.lastUpdate)
+      : params.deadline;
+  const accruedVault =
+    params.vaultData instanceof AccrualVaultV2
+      ? params.vaultData.accrueInterest(accrualTimestamp).vault
+      : params.vaultData.accrueInterest(accrualTimestamp);
+  const shares = accruedVault.toShares(params.assets, "Down");
```
Foulks-Plb: 🟡 **LOW — `computeVaultMaxSharePrice` can false-revert a declining-price Vault V2 deposit**

This previews shares only at the deadline-accrued snapshot, implicitly assuming the share price rises monotonically over `[now, deadline]`. That holds for V1 and yield-positive V2, but `AccrualVaultV2.accrueInterest` charges the management fee **regardless of yield** and mints it as shares while `_totalAssets` is unchanged — so an idle / low-yield V2 vault has a *declining* price. The deadline snapshot then yields more shares → a **lower** `maxSharePrice` than the current price, and a deposit included earlier reverts `SlippageExceeded`.

Trigger: V2 + a management-fee vault + a price decline over the window exceeding `slippageTolerance` — reachable with `slippageTolerance: 0n` or a multi-day `deadline` (**not** with the 2h / 0.03% defaults). Liveness/UX only, no fund loss (the user would actually receive *more* shares).

Fix — sample both endpoints and divide by the **minimum** shares (maximum price), mirroring the `max(current, accrued)` the sibling `computeVaultMaxShareAllowance` already uses:

```ts
const currentShares = params.vaultData.toShares(params.assets, "Down");
const accruedShares = accruedVault.toShares(params.assets, "Down");
const shares = MathLib.min(currentShares, accruedShares); // min shares → max price
```

Foulks-Plb (commented):
## Security / correctness review (Claude)

Behavior-preserving extraction of the Blue path into a shared bundles core, plus the VaultBundlesV1 prerequisites. I audited the full transaction flow (funding, referral fee, share price, signature/permit, requirement resolution) against the pinned `vaultBundlesV1Abi` and the on-chain check semantics.

**Verdict: safe for `next`.** No critical/high/medium. One LOW inline below.

**Verified clean**
- Calldata for `vaultBundlesV1Deposit` matches the pinned ABI positionally; permit tuples (`(deadline,v,r,s)` / `(nonce,deadline,signature)`) are byte-identical to the pre-refactor Blue encoder.
- ERC-2612 & Permit2 SignatureTransfer signed messages match what the contract reconstructs (spender = VaultBundlesV1, amount = gross, token/nonce/deadline committed); double-validated in `select*` **and** `getBundlesTokenPermit`; spender allow-listed to `{blueBundlesV1, vaultBundlesV1}`.
- `grossFromNetAssets` is an exact inverse of the on-chain `mulDivDown` fee; since `maxSharePrice` is a price ratio, net-rounding never reverts.
- `ReferralFeePctExceededError extends InputExceedsMaxError` preserves the documented pattern-match.
- `bundles/index.ts` no longer re-exports the `@internal` helpers.

**Pre-existing note (no action required):** `getBundlesTokenRequirements` probes the ERC-2612 nonce with `.catch(() => undefined)`, which also swallows a transient RPC error and silently falls back to Permit2/approval. Non-harmful (a genuine outage still surfaces on the subsequent un-caught allowance / nonce-bitmap reads), but consider narrowing the catch to a real "no `nonces()`" signal.

**Release-gate (tracking):** the `permit2TransferFrom → permit2SignatureTransfer` type/discriminator/encoder rename is a major-only break that skips the one-minor deprecation runway. Fine on `next`; it must not flip to a releasable major until the runway (closed PR #991) lands.

_Automated deep review; findings verified against the code, CI (fork suites green), and the pinned contract semantics._

packages/morpho-sdk/src/actions/bundles/resolveBundlesTokenRequirements.ts
```diff
@@ -0,0 +1,132 @@
...
+  readonly owner: Address;
+  readonly chainId: number;
+  readonly amount: bigint;
+  readonly deadline: bigint;
+  readonly state: BundlesTokenRequirementsState;
+}): readonly (
+  | Readonly<Transaction<ERC20ApprovalAction>>
+  | BundlesTokenSignatureRequirement
+)[] => {
+  if (params.amount < 0n) {
```
Rubilmax: 🟡 **MEDIUM — reject amounts above `uint256` in the standalone resolver**

The outer RPC getter validates this value, but `resolveBundlesTokenRequirements` is root-exported and callable independently. Existing pure encoders such as `encodeErc20Permit2SignatureTransfer` repeat their own uint bounds. Here `maxUint256 + 1n` can enter the approval branch and `encodeErc20Approval` caps the result to `maxUint256`, so the returned prerequisite cannot cover the requested pull. Replace the negative-only check with the shared `validateUint256Field("amount", params.amount)` and cover the upper bound.

packages/wdk-protocol-lending-morpho-evm/src/morpho-protocol-evm.test.ts
```diff
@@ -737,7 +737,7 @@ describe.sequential("MorphoProtocolEvm", () => {
       const requirementSignature = {
         args: { deadline: SIGNATURE_DEADLINE },
         action: { type: "permit2TransferFrom" },
-      } as unknown as BlueBundlesV1TokenRequirementSignature;
+      } as unknown as BundlesTokenRequirementSignature;
```
Rubilmax, line 740: 🟢 **LOW — use the current SignatureTransfer discriminator in this fixture**

The production requirement type and migration guide now use `permit2SignatureTransfer`, while this cast still hides the removed `permit2TransferFrom` shape. Sibling WDK tests intentionally use the same narrow `as unknown as` fixture style, so the matching minimal fix is just to change `action.type` to `permit2SignatureTransfer`; a full requirement fixture is unnecessary.

packages/morpho-sdk/src/actions/bundles/common.ts
```diff
@@ -0,0 +1,656 @@
...
+/** ABI-ready token permit consumed by fixed bundles deposit and funding entrypoints. */
+export interface BundlesTokenPermit {
+  /** TokenLib permit dispatch kind. */
+  readonly kind: BundlesPermitKind;
+  /** ABI-encoded permit payload, or empty bytes for classic allowance funding. */
+  readonly data: Hex;
+}
+
+/** ABI-ready ERC-2612 share permit consumed by vault bundles. */
+export interface BundlesSharesPermit {
```
Rubilmax: 🟡 **MEDIUM — keep one source of truth for the shares-permit tuple**

`BundlesSharesPermit` duplicates the six-field, root-exported `VaultExitBundlesV1PermitStruct` exactly. Existing compatibility surfaces use type aliases, and the package rule requires one interface when domain and ABI shapes are identical. Keep the generic shape as the canonical interface and make the VaultExit name a documented compatibility type alias rather than maintaining two structural definitions.

packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Approval.ts
```diff
@@ -55,6 +55,7 @@ export const encodeErc20Approval = (
       "midnight",
       "midnightBundles",
       "vaultExitBundlesV1",
+      "vaultBundlesV1",
```
Rubilmax, line 58: 🟢 **LOW — document the newly supported VaultBundlesV1 spender**

The allowlist now accepts `vaultBundlesV1`, but `params.spender` still enumerates every prior registry slot without it. Add VaultBundlesV1 to that supported-spender sentence so the generated API docs match the encoder.

packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Permit.ts
```diff
@@ -85,7 +85,12 @@ export const encodeErc20Permit = async (
   validateRequirementSpender({
     chainId,
     spender,
-    allowed: ["generalAdapter1", "midnightBundles", "blueBundlesV1"],
+    allowed: [
+      "generalAdapter1",
+      "midnightBundles",
+      "vaultBundlesV1",
```
Rubilmax, line 91: 🟢 **LOW — update both supported-spender JSDoc lists**

The implementation now accepts `vaultBundlesV1`, but both `params.spender` and the `UnsupportedErc20ApprovalSpenderError` description still list only GeneralAdapter1, MidnightBundles, and BlueBundlesV1. Add VaultBundlesV1 to both lists so the public contract matches the route implemented here.

packages/morpho-sdk/src/abis.ts
```diff
@@ -218,6 +218,147 @@ export const vaultExitBundlesV1Abi = [
   },
 ] as const satisfies Abi;
 
+/**
+ * Pinned ABI for the VaultBundlesV1 entrypoints used by Vault V1 and Vault V2 actions.
+ *
+ * Sourced from `morpho-org/bundles` commit
+ * `f27e7bcf744310303e24faa522b71d702e696686`.
+ */
+export const vaultBundlesV1Abi = [
```
Rubilmax, line 227: 🟡 **MEDIUM — pin the new VaultBundlesV1 selectors in tests**

This pinned ABI is release-critical, and the adjacent `blueBundlesV1Abi` already has a deployed-selector regression test. Add the same test for `vaultBundlesV1Abi` so tuple edits cannot silently change calldata: `initiator` `0x5c39fcc1`, deposit `0x6bbba4e0`, withdraw `0x932084a8`, and migrate `0x8355f776`.

packages/morpho-sdk/src/actions/bundles/common.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  NegativeInputError,
+  NonPositiveInputError,
+  type PermitRequirementSignature,
+} from "../../types/index.js";
+import {
+  getBundlesSharesPermit,
+  resolveBundlesFunding,
+  selectBundlesSharesRequirementSignature,
+  selectBundlesTokenRequirementSignature,
+} from "./common.js";
```
Rubilmax, line 40: 🟢 **LOW — test the exported token-permit converter directly**

`getBundlesTokenPermit` is a new root export with empty, ERC-2612, and Permit2 SignatureTransfer branches plus typed metadata-mismatch failures, but this suite never imports it. Existing signature/action converters have focused unit tests; add direct cases for the three outputs and the owner/token/amount/spender or malformed-signature error classes.

packages/morpho-sdk/src/actions/bundles/common.ts
```diff
@@ -0,0 +1,656 @@
...
+    deadline: params.deadline,
+    referralFeePct,
+    referralFeeRecipient: params.referralFeeRecipient ?? zeroAddress,
+  };
+};
+
+/**
+ * Resolves mutually exclusive ERC-20/native funding into the entrypoint amount and transaction value.
+ *
+ * @param params - Exclusive bundles funding.
```
Rubilmax, line 160: 🟢 **LOW — document the leaves of the new public parameter bags**

The repository's JSDoc convention and sibling action helpers document every options-bag field. `resolveBundlesFunding`, `getBundlesTokenPermit`, and `getBundlesSharesPermit` currently provide only a summary `@param params`. Add the corresponding `@param params.<field>` entries for each root-exported helper.

packages/morpho-sdk/src/actions/bundles/resolveBundlesTokenRequirements.ts
```diff
@@ -0,0 +1,132 @@
...
+      readonly nonceBitmap: bigint;
+    };
+
+/**
+ * Resolves pre-fetched allowance and Permit2 nonce state into ordered bundles requirements.
+ *
+ * The SignatureTransfer branch always places any ERC-20 approval to canonical Permit2 before the
+ * one-time signature requirement.
+ *
+ * @param params - Funding values, expected bundles spender, and pre-fetched state.
```
Rubilmax, line 44: 🟢 **LOW — document this resolver's structured inputs**

Its client-backed sibling `getBundlesTokenRequirements` documents each field, while this independently exported resolver collapses the whole discriminated bag into one summary. Add leaf `@param` entries for token, spender, owner, chainId, amount, deadline, and both `state` variants so generated docs explain which prefetched values each branch requires.

packages/morpho-sdk/src/helpers/referralFee.ts
```diff
@@ -0,0 +1,40 @@
+import { MathLib } from "@morpho-org/blue-sdk";
+import {
+  NegativeInputError,
+  ReferralFeePctExceededError,
+} from "../types/index.js";
+
+/**
+ * Computes the gross amount whose post-referral-fee proceeds equal `netAssets` exactly.
+ *
+ * @param params - Net target and WAD-scaled referral fee.
```
Rubilmax, line 10: 🟢 **LOW — document both referral calculation inputs**

`grossFromNetAssets` is a new root-exported options-bag helper, and established computation helpers document every leaf. Add `@param params.netAssets` and `@param params.referralFeePct` so the units and WAD scaling are visible in generated API docs.

packages/morpho-sdk/src/helpers/slippage.ts
```diff
@@ -206,3 +213,135 @@ export function computeMinWithdrawSharePrice(params: {
...
+
+/**
+ * Computes the RAY-scaled maximum share price for a VaultBundlesV1 deposit leg.
+ *
+ * Previews shares on both the supplied Vault V1 or Vault V2 snapshot and its deadline-accrued
+ * counterpart, dividing by the smaller of the two so the bound covers the highest price the
+ * bundle may execute at. Shares are previewed rounding down, mirroring the ERC-4626 `deposit()`
+ * shares the on-chain `maxSharePrice` check divides by.
+ *
+ * @param params - Vault snapshot, bundles execution deadline, net assets, and slippage.
```
Rubilmax, line 225: 🟢 **LOW — preserve the file's leaf-level parameter docs**

Every pre-existing public computation helper in this file documents its options individually. Add `params.vaultData`, `params.deadline`, `params.assets`, and `params.slippageTolerance` entries to both new VaultBundlesV1 helpers; the summary-only tags omit the accepted snapshot types and units.

packages/morpho-sdk/src/helpers/slippage.ts
```diff
@@ -206,3 +213,135 @@ export function computeMinWithdrawSharePrice(params: {
...
+ * @returns The maximum shares the prepared operation may burn.
+ * @throws {NonPositiveInputError} when `assets` or the computed share cap is not positive.
+ * @throws {NegativeInputError} when `slippageTolerance` is negative.
+ * @throws {ExcessiveSlippageToleranceError} when `slippageTolerance` exceeds the SDK maximum.
+ * @example
+ * ```ts
+ * import { computeVaultMaxShareAllowance } from "@morpho-org/morpho-sdk";
+ *
+ * const shares = computeVaultMaxShareAllowance({
+ *   vaultData,
```
Rubilmax, line 326: 🟢 **LOW — make this public example copy-pasteable**

`vaultData` is never declared, so the example cannot run. Mirror the immediately preceding helper's established setup: import and call `fetchAccrualVault`, create the public client, and obtain the timestamp before invoking `computeVaultMaxShareAllowance`.

packages/morpho-sdk/src/actions/bundles/resolveBundlesTokenRequirements.test.ts
```diff
@@ -0,0 +1,165 @@
...
+        expect(requirement?.action).toMatchObject({
+          type: "permit2SignatureTransfer",
+          args: { nonce: BigInt(nonce) },
+        });
+      }),
+      { numRuns: 100, seed: 20_260_911 },
+    );
+  });
+
+  test("error: Permit2SignatureTransferNonceAlreadyUsedError", () => {
```
Rubilmax, line 256: 🟢 **LOW — cover the resolver's remaining public boundary cases**

This exported resolver advertises several typed failures, but the suite asserts only the used-nonce case. Match sibling action tests by adding class-identity cases for negative and over-`uint256` amounts, an approval target below the spend amount, and negative or over-`uint256` Permit2 nonces, plus the zero-amount early return.

packages/morpho-sdk/src/actions/bundles/common.ts
jinmel: Agreed. Make `BundleSharesPermit` the canonical export interface.

packages/morpho-sdk/src/actions/bundles/resolveBundlesTokenRequirements.ts
jinmel: Addressed in e0e11021. Replaced the negative-only check with `validateUint256Field("amount", params.amount)` before either branch. Added class-identity tests for negative and over-uint256 amounts on both paths, and verified that `maxUint256` remains accepted.

packages/wdk-protocol-lending-morpho-evm/src/morpho-protocol-evm.test.ts
jinmel: Addressed in e0e11021. Changed this fixture to `permit2SignatureTransfer`, retaining the existing narrow fixture style. The WDK unit suite passes.

packages/morpho-sdk/src/actions/bundles/common.ts
jinmel: Addressed in e0e11021. Made `BundleSharesPermit` the canonical root-exported interface, following the naming clarification. `VaultExitBundlesV1PermitStruct` and the existing `BundlesSharesPermit` name now alias that interface. Added type assertions for both aliases; SDK and WDK TypeScript checks pass.

packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Approval.ts
jinmel: Addressed in e0e11021. Added VaultBundlesV1 to the supported-spender description for `params.spender`.

packages/morpho-sdk/src/actions/requirements/encode/encodeErc20Permit.ts
jinmel: Addressed in e0e11021. Added VaultBundlesV1 to both the `params.spender` list and the `UnsupportedErc20ApprovalSpenderError` description.

packages/morpho-sdk/src/abis.ts
jinmel: Addressed in e0e11021. Added a `vaultBundlesV1Abi` selector regression test pinning all four selectors: initiator `0x5c39fcc1`, deposit `0x6bbba4e0`, withdraw `0x932084a8`, and migrate `0x8355f776`. The test passes against the current ABI.

packages/morpho-sdk/src/actions/bundles/common.test.ts
jinmel: Addressed in e0e11021. Added direct tests for the empty permit, serialized/compact ERC-2612 permits, and Permit2 SignatureTransfer payloads. Tests also assert owner, token, both amount fields, spender, deadline, nonce, malformed-signature errors with their cause, and rejection of Permit2 AllowanceTransfer.

packages/morpho-sdk/src/actions/bundles/common.ts
jinmel: Addressed in e0e11021. Added field-level `@param` documentation for all inputs to `resolveBundlesFunding`, `getBundlesTokenPermit`, and `getBundlesSharesPermit`, including units, optional fields, and funding exclusivity.

packages/morpho-sdk/src/actions/bundles/resolveBundlesTokenRequirements.ts
jinmel: Addressed in e0e11021. Documented token, spender, owner, chainId, amount, deadline, and every field of both state variants. The Permit2 bitmap description now specifies the word selected by `permit2Nonce >> 8n`.

packages/morpho-sdk/src/helpers/referralFee.ts
jinmel: Addressed in e0e11021. Added `params.netAssets` and `params.referralFeePct` documentation, specifying asset base units and WAD scaling.

packages/morpho-sdk/src/helpers/slippage.ts
jinmel: Addressed in e0e11021. Added all four field-level parameter entries to both helpers, including the accepted Vault V1/V2 snapshots, Unix-second deadlines, asset units, and WAD-scaled slippage.

packages/morpho-sdk/src/helpers/slippage.ts
jinmel: Addressed in e0e11021. Made the example self-contained with `fetchAccrualVault`, a public client, a fetched vault snapshot, and a block timestamp used to derive the deadline.

packages/morpho-sdk/src/actions/bundles/resolveBundlesTokenRequirements.test.ts
jinmel: Addressed in e0e11021. Added zero-amount and maxUint256 cases for both branches, class-identity assertions for negative/over-uint256 amounts and Permit2 nonces, and an insufficient approval-target case. The resolver suite passes.

packages/morpho-sdk/src/actions/bundles/resolveBundlesTokenRequirements.ts
```diff
@@ -0,0 +1,146 @@
...
+  | {
+      readonly type: "approval";
+      /** Current allowance of the funding token for `spender`. */
+      readonly allowance: bigint;
+      /** Allowance to set when the current allowance cannot cover `amount`. */
+      readonly approvalAmount: bigint;
+    }
+  | {
+      readonly type: "permit2SignatureTransfer";
+      readonly permit2: Address;
```
Rubilmax: Such address should always be taken dynamically from the chain registry, and not be passed as parameter. This avoids being stateful with parameters

packages/morpho-sdk/src/actions/bundles/resolveBundlesTokenRequirements.ts
```diff
@@ -0,0 +1,146 @@
...
+ *   amount: 1_000_000n,
+ *   deadline: 1_900_000_000n,
+ *   state: { type: "approval", allowance: 1_000_000n, approvalAmount: 1_000_000n },
+ * });
+ * // requirements is empty because the allowance already covers the amount.
+ * ```
+ */
+export const resolveBundlesTokenRequirements = (params: {
+  readonly token: Address;
+  readonly spender: Address;
```
Rubilmax, line 88: What we were doing previously was pass a spender as input to the getBundlesTokenPermit but still validate it against a set of allowed spenders taken from the chain's registry of addresses. This way, we are guaranteed to never approve unexpected spenders compared to what's in the chain registry. This seems to not apply anymore, could we restore that?

Rubilmax (changes requested):

Foulks-Plb (approved):

Rubilmax (approved):

PR #1002, opened 2026-09-03
chore(release): pin build/test toolchain via a pnpm catalog (typescript, vitest)

Foulks-Plb:
## What

Introduces a **pnpm catalog** as the single source of truth for the shared build/test toolchain, and pins it exact where it matters:

- `typescript` `^7.0.2` → exact `7.0.2`, referenced via `catalog:` from the root manifest and all 10 package manifests.
- `vitest` `^4.1.10` → moved onto the same `catalog:` reference across its 10 devDependency entries.

A single `catalog:` block in `pnpm-workspace.yaml` holds both versions. Lockfile specifiers are regenerated; resolved versions are unchanged (`typescript@7.0.2`, `vitest@4.1.10`).

## Why

Addresses the concrete, actionable item of Cantina AI finding **SDKS-2** — *"Floating TypeScript compiler can rewrite authority-bearing Morpho artifacts during prepublish."*

The release pipeline builds published `lib` artifacts with `tsc` during `prepublish`, but the compiler was requested as a caret range. Exact-pinning removes the (already narrow) window for the compiler that emits published artifacts to drift within `^7.0.2` at a future dependency refresh, and makes the toolchain consistent with how the repo already pins `viem 2.55.10` and `solc 0.8.36` exactly.

Using a catalog rather than duplicated inline ranges gives one source of truth (§1 of `AGENTS.md`) — a single place to audit and to bump the toolchain, instead of keeping ~10 identical entries per tool in lockstep by hand.

## Severity note

SDKS-2 was AI-rated *Critical*; assessed here as **Low (hardening / defense-in-depth)**. The "floating compiler" premise is already largely defeated in the current setup — the compiler is pinned + integrity-hashed by `pnpm-lock.yaml` and installed with `--frozen-lockfile` at publish, `minimumReleaseAgeStrict` (3-day) gates any bump, and npm provenance + trusted publishing (OIDC) are in place. This PR takes the one cheap, unambiguous win. The remaining checklist items (independent compiler-identity verification, reproducible-build digest comparison of the emitted `lib` tree) are optional hardening, not addressed here.

## Scope / safety

- `devDependency`-only — neither `typescript` nor `vitest` is a runtime dependency anywhere, so this has **no consumer-facing impact**.
- `catalog:` is resolved by pnpm to the concrete range at install/publish; both are devDeps so they do not appear in any published manifest.
- The `@morpho-org/test` `vitest` **peerDependency** range (`>=2.1.2 <5.0.0`) is left inline — it is a deliberately-wide consumer contract, not a dev pin, and must not be collapsed to the catalog.
- Resolved lockfile versions unchanged (verified via `pnpm exec tsc --version` → `7.0.2`, `pnpm exec vitest --version` → `4.1.10`) → no published-package contract change, no behavior change. The pre-existing transitive `typescript@6.1.0` is untouched.
- No changeset required (does not change any published package's contract).

## Note on scope

`viem` is deliberately **not** catalogged: it spans four intentional ranges including consumer-facing `peerDependency` ranges (`^2.0.0`, `^2.21.10`) that must stay loose. Runtime deps (`zod`) and low-count devDeps (`fast-check`) are out of scope for this hardening PR.


0xbulma (commented):
Can't we have a catalog version?

jinmel (approved):

PR #1021, opened 2026-09-07
fix(liquidity-sdk-viem): isolate reallocations by target

Rubilmax:
## What

- [SDKS-92](https://app.notion.com/3cfd69939e6d8154918ed1977e18b8d6) limits each target planner to that target's normalized supplying-vault set while retaining the union only for batched reads. **Example:** Concurrent plans for disjoint markets let target A count or withdraw target B's vault liquidity, overstating A's capacity and starving or reverting B.

## Why

Concurrent target planning must not source withdrawals from vaults that are unrelated to the target being fulfilled.

### DIFF
--- a/.changeset/quiet-targets-plan.md
+++ b/.changeset/quiet-targets-plan.md
@@ -1,0 +1,5 @@
+---
+"@morpho-org/liquidity-sdk-viem": patch
+---
+
+Isolate batched shared-liquidity planning to each target market's supplying vaults.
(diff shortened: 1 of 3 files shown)


Rubilmax:
Closing: this does not address a real concurrency issue. DataLoader only batches reads, and each planner simulates from its own clone. The API defines supplyingVaults as whitelisted vaults with the target enabled and a non-zero cap; the planner already requires that target config and cap headroom, plus positive source liquidity. Therefore a union-only vault cannot contribute. The test constructs inconsistent API/onchain data and only verifies option forwarding because it stubs the planner. Genuine shared-vault transactions can still contend, so this filter does not provide concurrency isolation.

PR #1022, opened 2026-09-07
fix(midnight-sdk): reject incoherent ratifier trees

Rubilmax:
## What

- [SDKS-104](https://app.notion.com/3cfd69939e6d81fcb8a9f3e7f54f3072) accepts cached ratifier trees only when their visible offers, padded offers, leaves, root, and height describe the same tree, while retaining valid extra zero padding. **Example:** A maker UI refreshes visible offers but reuses stale leaves containing an older permissive offer; the wallet signs the root and a counterparty later fills the hidden offer.

## Why

A signer must not approve a root that can contain offers hidden from the visible offer list.


Rubilmax (commented):
LGTM

Foulks-Plb (dismissed):

yum0e (dismissed):

jinmel (dismissed):

Foulks-Plb:
## Review pass — `fix(midnight-sdk): reject incoherent ratifier trees`

Reviewed at `63291f87`. The core guard is correct: the coherence check (height bounds, `paddedOffers.length === leaves.length === 2**height`, each visible offer non-padding and hashing to its leaf with a duplicate guard, empty padding past the visible prefix, recomputed leaves, and a reconstructed root equal to `tree.root`) fully blocks the hidden-real-offer attack, and the invariant tests fail if the guard is removed.

The two items raised by Codex/Devin on the earlier head appear addressed by the latest push:
- recognizing the derived default group when rejecting padding — `offerStructInternal.ts` now treats `EMPTY_OFFER_DEFAULT_GROUP = keccak256(hashStruct(empty))` as padding, which targets exactly that case;
- avoiding a full tree re-validation per ratifier item — the `EcrecoverRatifierUtils` / `SetterRatifierUtils` `ratify` paths were refactored.

No blocking issues from my side.

jinmel (approved):

PR #1023, opened 2026-09-07
fix(blue-sdk-viem): reject oversized Permit2 allowances

Rubilmax:
## What

- [SDKS-128](https://app.notion.com/3cfd69939e6d8173bf88d570701e64fc) rejects finite Permit2 allowances above `MAX_UINT_160`, preserves exact-max as the caller's explicit unlimited value, and exposes the typed overflow failure through `blue-sdk-viem` and both `morpho-sdk` Blue facades. **Example:** A frontend conversion bug requests a finite allowance just above uint160; silent clamping makes it unlimited, letting a compromised spender pull future deposits.

## Why

An overflowing finite allowance must fail explicitly instead of becoming an unlimited approval.

### DIFF
--- a/.changeset/brave-permits-stop.md
+++ b/.changeset/brave-permits-stop.md
@@ -1,0 +1,6 @@
+---
+"@morpho-org/blue-sdk-viem": minor
+"@morpho-org/morpho-sdk": minor
+---
+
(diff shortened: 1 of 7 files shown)


Rubilmax:
Fixed on `origin/next`

PR #1026, opened 2026-09-07
fix(evm-simulation): fail closed on incomplete retention evidence

Rubilmax:
## What

- [SDKS-5](https://app.notion.com/3cfd69939e6d811e8511e36195e83c0b) requires registered Bundler retention metadata before backend execution. **Example:** A Next.js UI registers a custom L2 in its ESM bundle while a CommonJS simulation worker loads another registry instance; the worker otherwise approves a route that leaves USDC in Bundler3.
- [SDKS-41](https://app.notion.com/3cfd69939e6d8136a9e0d95895b41f1a) rejects oversized Tenderly asset-change amounts instead of truncating them to uint256. **Example:** A gateway returns a 65-hex-digit outbound amount that cancels a real Bundler3 inflow in net accounting, so the frontend enables Submit while funds remain stranded.
- [SDKS-49](https://app.notion.com/3cfd69939e6d81388750fca2a0723e10) preserves execution reverts instead of reclassifying them as fallback-eligible service failures. **Example:** Tenderly reports a slippage revert without optional success fields; a fallback at a lagging block succeeds and the UI sends a transaction guaranteed to revert and burn gas.
- [SDKS-73](https://app.notion.com/3cfd69939e6d8187a7d7c19f5fe8e3c6) accepts only canonical uint256 event data. **Example:** A malformed RPC proxy returns 66-character negative decimal log data; length-only parsing treats it as an outflow that cancels real retained tokens.
- [SDKS-76](https://app.notion.com/3cfd69939e6d8198a5bad01b47ce860f) accepts only canonical indexed address topics. **Example:** A token emits a forged outbound Transfer topic with nonzero padding but a Bundler3 address suffix; tail slicing aliases it to Bundler3 and cancels a real inflow.
- [SDKS-80](https://app.notion.com/3cfd69939e6d81699216d5fcd6a19b5e) fails closed when the configured chain has no registered Bundler retention metadata. **Example:** A wallet adds a new L2 RPC but omits deployment registration; simulation otherwise returns green when its route strands funds in that chain's Bundler3.
- [SDKS-111](https://app.notion.com/3cfd69939e6d81fb862cce7e3daa2688) normalizes address and event-signature case before comparisons. **Example:** An RPC proxy uppercases valid hex topics; transfers into Bundler3 disappear from accounting and a retaining route looks safe.
- [SDKS-120](https://app.notion.com/3cfd69939e6d81ceb9fad848fb858268) requires successful Tenderly responses to include asset-change evidence. **Example:** Tenderly returns `status: true` but omits `assetChanges`; retained native ETH emits no ERC-20 log, so a checkout UI incorrectly enables execution.
- [SDKS-147](https://app.notion.com/3cfd69939e6d819c9d71dbf3120d1d1d) recognizes Deposit and Withdrawal events only from the registered wrapped-native token, and ignores those events on registered chains without one. **Example:** A malicious token emits a WETH-shaped Withdrawal beside a real transfer of itself into Bundler3; the fake outflow otherwise nets retention to zero while its balance remains.

## Why

Incomplete registry state, malformed backend evidence, and lookalike token logs must fail closed rather than disable or cancel retained-value detection.

### DIFF
--- a/.changeset/strict-simulations-retain.md
+++ b/.changeset/strict-simulations-retain.md
@@ -1,0 +1,5 @@
+---
+"@morpho-org/evm-simulation": patch
+---
+
+Fail closed when bundler metadata or Tenderly retention evidence is incomplete, validate canonical transfer logs, and trust wrapped-native events only from the registered token.
--- a/packages/evm-simulation/AGENTS.md
+++ b/packages/evm-simulation/AGENTS.md
@@ -5,9 +5,9 @@
 - Let `SimulationRevertedError` propagate; a revert belongs to the bundle, not the backend.
 - Keep backend outputs normalized to `RawSimulationResult`; add new backends under `src/simulate/backends/` with colocated parity tests.
 - Encode signature authorizations as `approve(spender, amount ?? maxUint256)` and prepend them to the simulated bundle.
-- Enforce bundler retention by net `(bundler3 address, token)` balance with `DUST_THRESHOLD = 100n`; skip only unknown blue-sdk chains.
+- Enforce bundler retention by net `(bundler3 address, token)` balance with `DUST_THRESHOLD = 100n`; reject chains without bundler metadata in the active address registry.
 - Keep all thrown domain errors under `SimulationPackageError`; only `ExternalServiceError` is bypassable by callers.
-- Add chains through caller `SimulationConfig.chains`; the per-chain `ChainSimulationConfig` is a discriminated union enforcing at least one of `tenderlyRpc` or `simulateV1Url`. Confirm blue-sdk bundler addresses intentionally.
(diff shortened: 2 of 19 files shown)


Rubilmax:
Closing this as low-value in its current form.

The linked relevance assessment already marks SDKS-49, SDKS-76, SDKS-80, SDKS-111, and SDKS-147 as Not a bug / Nullified; SDKS-5, SDKS-41, and SDKS-73 as Informational low-likelihood; and only SDKS-120 as Low. The diff nevertheless combines all nine into 19 files and 786 changed lines, without a production or Tenderly fixture; the new regressions construct malformed mocked responses.

It also changes established custom-chain behavior: a configured backend that previously simulated while warning and skipping unavailable retention metadata now throws before RPC unless the caller mutates the blue-sdk registry in the same runtime module graph. That does not solve the ESM/CJS split-registry problem, and it is too large a compatibility change for this patch bundle. Several parser checks also do not create a meaningful security boundary because a malicious provider or token can emit canonical false evidence.

If this area becomes a product priority, follow up with a focused PR for one confirmed issue—for example, requiring documented Tenderly success evidence for SDKS-120 or scoping WETH Deposit/Withdrawal parsing to the registered wrapped-native token—backed by a real response or consumer reproduction and an explicit custom-chain API decision.

PR #1029, opened 2026-09-08
fix(evm-simulation): scope wrapped-native events to registry

Rubilmax:
## Motivation

`parseTransfers` currently interprets any contract emitting the WETH9 `Deposit(address,uint256)` or `Withdrawal(address,uint256)` signature as wrapped-native value movement. A real Anvil execution reproduces the issue: a non-`wNative` contract emits `Withdrawal(bundler, amount)`, and `origin/main` includes that lookalike token in both `transfers` and `assetChanges`.

SDKS-120 is intentionally excluded because no current Tenderly response omitting `assetChanges` could be reproduced or found in captured evidence.

## Solution

- Resolve the chain registry `wNative` once in `simulate` and thread it through both transfer-parsing passes.
- Accept WETH9-specific events and apply mint/burn deduplication only for that registered emitter.
- Preserve legacy signature-based parsing for configured custom chains without registry metadata.
- Add a mainnet-fork regression that executes a real WETH withdrawal and an EVM lookalike emitter in the same bundle. It proves registered WETH is retained while the lookalike is absent from both outputs; the same test fails on `origin/main`.
- Add a patch changeset for `@morpho-org/evm-simulation`.

## Testing

- `pnpm --filter @morpho-org/evm-simulation test` (238 tests)
- `pnpm --filter @morpho-org/evm-simulation test:fork` (4 tests)
- `pnpm --filter @morpho-org/evm-simulation build`
- Biome on all 9 changed TypeScript files
- JSDoc self-check and checksum-address lint

### DIFF
--- a/packages/evm-simulation/src/simulate/parsing/transfers.ts
+++ b/packages/evm-simulation/src/simulate/parsing/transfers.ts
@@ -28,10 +28,11 @@ const UINT256_HEX_LENGTH = 66; // "0x" + 32 bytes
  * with the index of the originating call.
  *
  * **Supported event types:** ERC20 `Transfer(from, to, amount)` and WETH9
- * `Deposit(to, amount)` / `Withdrawal(from, amount)`. WETH9 mint/burn Transfer
- * events paired with their Deposit/Withdrawal are deduplicated. ERC721 and
- * ERC1155 transfer events are **not** parsed — consumers with NFT flows will
- * see an incomplete transfer list.
+ * `Deposit(to, amount)` / `Withdrawal(from, amount)` from the registered
+ * wrapped-native token. Chains without registry metadata retain legacy
+ * signature-based parsing. Paired mint/burn Transfer events are deduplicated.
+ * ERC721 and ERC1155 transfer events are **not** parsed — consumers with NFT
+ * flows will see an incomplete transfer list.
  *
  * **Native ETH (`eth_simulateV1` + `traceTransfers`).** When the backend runs
  * `eth_simulateV1` with `traceTransfers` enabled, native-ETH moves — including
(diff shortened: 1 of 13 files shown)


Rubilmax (commented):
LGTM

packages/evm-simulation/src/simulate/simulate.ts
```diff
@@ -80,20 +87,31 @@ export async function simulate(
 ): Promise<SimulationResult> {
   validateInput(params);
 
+  let wNative: Address | undefined;
+  try {
+    wNative = getChainAddresses(params.chainId).wNative;
+  } catch (error) {
+    if (!(error instanceof UnsupportedChainIdError)) throw error;
```
Rubilmax: use `_try`?

packages/evm-simulation/src/simulate/parsing/transfers.ts
```diff
@@ -48,39 +49,46 @@ const UINT256_HEX_LENGTH = 66; // "0x" + 32 bytes
...
-
-  // Tokens that emit `Deposit` or `Withdrawal` somewhere in the bundle look
-  // wnative-shaped. If a zero-address `Transfer` for one of these tokens
-  // misses same-tx dedup, the contract is likely emitting non-canonically
-  // (split across txs) and a phantom transfer would leak into retention.
+  const { logger } = options;
+  const wNative = options.wNative?.toLowerCase();
   const wnativeShapedTokens = collectWnativeShapedTokens(calls);
+  const acceptsWnativeEvent = (address: Address): boolean =>
+    wNative === undefined || address.toLowerCase() === wNative;
```
Rubilmax: isAddressEqual?

packages/evm-simulation/src/simulate/simulate.ts
```diff
@@ -82,20 +89,29 @@ export async function simulate(
 ): Promise<SimulationResult> {
   validateInput(params);
 
+  const wNative = _try(
+    () => getChainAddresses(params.chainId).wNative ?? null,
```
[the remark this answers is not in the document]
Rubilmax: Wanted behavior

Foulks-Plb:
## Review pass — `fix(evm-simulation): scope wrapped-native events to registry`

Reviewed at `3daf41b2`. The fix is sound: `parseTransfers` now only treats WETH9 `Deposit`/`Withdrawal` as wrapped-native for the registered `wNative` emitter, and the added mainnet-fork regression is a genuine Anvil execution (real `eth_simulateV1`, lookalike emitter via `setCode`) that reproduces the bug and fails on `main`.

Re-verified against the current head — the earlier concerns are addressed:
- **Merge conflict** with recent `main` evm-simulation changes: resolved (now mergeable).
- **Tokenless known chains** (Devin's flag): `acceptsWnativeEvent` now distinguishes three states — `undefined` (unknown/custom chain → legacy accept-all), `null` (known chain with no `wNative` → reject all lookalikes), and an address (match only that emitter). `simulate()` resolves `getChainAddresses(chainId).wNative ?? null` via `_try(..., UnsupportedChainIdError)`, so known-but-tokenless chains (e.g. Stable/Celo) correctly reject lookalikes. 👍
- **Codex "unrelated v6 prerelease state" (P1):** no longer applies on this head — `morpho-sdk` is `5.9.0`, there's no `.changeset/pre.json`, and only `filter-wrapped-native-events.md` is added.

No blocking issues from my side. LGTM.

jinmel (approved):

PR #1046, opened 2026-09-09
ci: verify committed deployless artifacts

Rubilmax:
## What

- [SDKS-152](https://linear.app/morpho-labs/issue/SDK-435/sdks-152medium-release-publishes-unverified-deployless-abi-and) verifies that committed deployless ABIs and bytecode match their reviewed Solidity before package builds and releases. **Example:** A reviewer approves a changed Solidity return tuple but stale generated bytecode remains committed; CI now stops a package whose state reader would execute different logic from the reviewed source.

## Why

Published deployless queries must reflect reviewed source because stale or hand-edited artifacts can silently change onchain reads.


Rubilmax (commented):
LGTM

package.json
```diff
@@ -25,7 +25,7 @@
   "scripts": {
     "prepare": "husky",
     "build": "pnpm -r --if-present build",
-    "build:ci": "pnpm -r --if-present --filter './packages/**' build",
+    "build:ci": "$npm_execpath verify:deployless-artifacts && pnpm -r --if-present --filter './packages/**' build",
```
jinmel: i think AI just prefers $npm_execpath over `pnpm`. Can we keep using `pnpm` ?

Foulks-Plb (approved):

PR #1053, opened 2026-09-09
fix(midnight-sdk): validate rounded quote settlement

Rubilmax:
## What

- [SDKS-59](https://linear.app/morpho-labs/issue/SDK-494/sdks-59medium-midnight-quote-guard-ignores-aggregate-integer): Validates `averageWorstPrice` from side-appropriate, per-fill rounded settlement amounts and accepts the current settlement fee for effective buyer or seller pricing. Example: two 1-unit asks priced at 0.5 asset units each settle to 1 asset unit apiece; a frontend now rejects the realized 1.0 average instead of accepting the theoretical 0.5 average.

## Why

Frontends use the quote guard to decide which signed offers to submit. Checking the same integer amounts settlement uses prevents a quote response from passing preflight with an aggregate price worse than the user's bound.

### DIFF
--- a/.changeset/quiet-books-settle.md
+++ b/.changeset/quiet-books-settle.md
@@ -1,0 +1,6 @@
+---
+"@morpho-org/midnight-sdk": minor
+"@morpho-org/morpho-sdk": minor
+---
+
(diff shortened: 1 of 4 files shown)


Rubilmax (commented):
LGTM

Foulks-Plb (approved):

PR #1067, opened 2026-09-14
feat(morpho-sdk)!: expose signature Requirement EIP-712 payload as action.typedData

Foulks-Plb:
## Motivation

Signature `Requirement`s returned by `getRequirements()` only exposed `sign(client, userAddress)`, which builds the EIP-712 payload internally, signs it with a viem `WalletClient`, verifies the recovered signer, and returns a `RequirementSignature`. Integrators who want to **inspect or display the exact payload before signing** — to preview it in a UI, log it, or diff it against expectations — had no way to obtain it.

## Solution

Each signable requirement **action** now carries the EIP-712 payload as `action.typedData`:

```ts
const req = (await output.getRequirements()).find(isRequirementSignature);
if (req == null) return; // nothing to sign (only on-chain approvals, or none)

const typedData = req.action.typedData; // exact EIP-712 payload sign() will sign — no null check
const signed = await req.sign(walletClient, owner);
const tx = output.buildTx([signed]); // Midnight outputs take the single signature instead
```

Covered actions: ERC-2612 permit (`PermitAction`), Permit2 AllowanceTransfer (`Permit2Action`), Permit2 SignatureTransfer (`Permit2SignatureTransferAction`), Morpho authorization (`AuthorizationAction`), and the Midnight offer-root (`MidnightOfferRootSignatureAction`).

- **`typedData`** is built at requirement-creation time and `deepFreeze`d (the permit `domain` is copied first so a caller's `Token.eip5267Domain` is never frozen by reference). On `Requirement.action` it is typed **required** (`RequirementResult["action"] & { readonly typedData: RequirementTypedData }`) so strict TS accepts `req.action.typedData` with no null check; it stays optional on the standalone action interfaces so hand-built metadata (e.g. test fixtures) is unaffected.
- It is the exact `TypedDataDefinition` that `sign()` signs. **Signing itself still goes through `Requirement.sign(client, userAddress)`**, which continues to build nothing new and verify the recovered signer. This PR only *exposes* the payload; it does not add an external-signature intake path.

### Breaking

Because the permit/authorization payload embeds the owner and is now built eagerly, two low-level exported encoders take a new **required `owner`** parameter: `encodeErc20Permit` and `encodeBlueSignatureAuthorization` (forwarded by `getGeneralAdapterRequirementsPermit`). Their `sign()` now also rejects a signer that differs from `owner`. In addition, `Requirement.action` now carries a **required `typedData`** field, which affects any consumer constructing a `Requirement` by hand. All in-SDK requirement resolvers already supply the owner, so high-level flows are unaffected. Marked as a `major` changeset (consistent with the `6.0.0-next` line) and documented in `MIGRATION-v5-to-v6.md`.

**Tests** — each encoder gets an `action.typedData` payload-shape test (including frozenness) and a test asserting that signing `action.typedData` externally recovers the same signer / matches `sign()`. Midnight proves the offer-tree payload is exposed and recovers its maker. Full morpho-sdk unit suite green.

Linear: [SDK-562](https://linear.app/morpho-labs/issue/SDK-562/expose-requirement-eip-712-typeddata-and-accept-external-signatures)

> Targets `next` (this branch is based on the next line).


Foulks-Plb:
Review requested — pinging the bots for a pass on this PR.

@codex review
@devin review
@claude please review

0xbulma (changes requested):
Planification / specification process to be discussed before we move forward.

milanmorpho:
Thanks, this appears to provide the core primitives we need for external EOA signing: expose the exact typed data, accept the resulting signature through `withSignature()`, and let the SDK produce the `RequirementSignature` consumed by `buildTx()`.

One important API integration concern remains. Our intended flow crosses two HTTP requests:

1. prepare the action and return signature requirements
2. collect signatures externally and call finalize

The examples work while the same in-memory `output` and `requirement` objects remain alive, but both objects contain closures and cannot be serialized into Redis or a database. Keeping them behind an in-memory `preparationId` would couple finalization to one process and would break across restarts, deployments, or horizontal instances.

Recreating the action during finalize is also risky. The deadline must be identical, ERC-2612 / Morpho authorization / Permit2 nonce state may have changed, and VaultBundles exact-asset withdrawals derive an exact share allowance from a captured vault snapshot. A freshly prepared requirement may therefore differ from what the user signed.

Could the SDK expose an official stateless boundary, for example one of:

- a JSON-safe serializable prepared-action/requirement snapshot plus a restore/finalize function; or
- a pure helper that accepts an SDK-produced serialized requirement descriptor, signer, and signature and returns the validated `RequirementSignature`, together with enough serialized prepared-action state to build the same transaction?

The goal is to avoid duplicating the SDK signature metadata and consistency checks in the API while also avoiding process-local state.

A few related items that seem worth confirming before we integrate:


- Permit2 SignatureTransfer still requires the caller to provide an unused nonce. It would be useful to confirm whether nonce allocation/idempotency is intentionally an integrator concern.

- `getRequirements()` can return a mixture of calls and multiple signature requirements. We plan to assign API-level requirement IDs and return them separately; please confirm there is no SDK identity/order invariant beyond passing all matching `RequirementSignature` values to the same prepared output.
- The PR and migration-guide examples use `output.buildTx(signed)`. Vault and Blue action outputs appear to expect `readonly RequirementSignature[]`, so I believe those examples should use `output.buildTx([signed])`.

If a serialization/rehydration API is intentionally out of scope for this PR, confirmation of the recommended stateless server pattern would still help us avoid building against an unsupported assumption.

Foulks-Plb:
We don't need any stateful API between the two steps. No sticky sessions, no in-memory `preparationId`, and no server-side store at all. Everything that has to survive between `prepare` and `finalize` is JSON-safe plain data that the **client carries in the round-trip**; the server keeps nothing and rebuilds the transaction on any instance.

### Why it's stateless-friendly

Two properties of the current design make this work:

1. **`buildTx()` is synchronous and does no RPC.** It only forwards its inputs to the **pure, exported** action builders (`blueSupply`, `blueSupplyCollateral`, `vaultV1InKindRedeem`, …) and consumes a plain `readonly RequirementSignature[]` — i.e. `{ args, action }`, deep-frozen plain data, no closures. So `finalize` can skip the live `output` object entirely and call the pure action directly.
2. **The signature boundary is pure data in / pure data out.** `action.typedData` is the exact EIP-712 payload to hand to any signer; `withSignature(signature, userAddress)` verifies it offline (EOA ECDSA recovery) and returns that same plain `RequirementSignature`. You never need the original in-memory `Requirement` to turn a signature into something `buildTx` accepts.

### Architecture — two requests, the client carries the continuation, the server stores nothing

```
Client (browser / EOA / remote signer)              API (any instance — keeps NOTHING between calls)
     │                                                        │
     │  POST /prepare { market, amount, userAddress }         │
     │ ──────────────────────────────────────────────────►   │
     │            output = client.morpho.blue(market).supply({ amount, userAddress })
     │            reqs   = await output.getRequirements()      // the only RPC reads happen here
     │  200 { typedData[] }                                    │
     │ ◄──────────────────────────────────────────────────    │
     │                                                        │
     │  sign each typedData  (walletClient.signTypedData / account / remote KMS/HSM)
     │  → the client now holds: its own inputs + typedData + signatures
     │                                                        │
     │  POST /finalize { market, amount, userAddress,         │
     │                   typedData[], signatures[] }          │  ← may be a DIFFERENT instance;
     │ ──────────────────────────────────────────────────►   │    it received nothing from /prepare
     │            sig = requirement.withSignature(signature, userAddress)   // pure verify, no state
     │            maxSharePrice = <recompute from a fresh RPC read>          // unsigned bound
     │            tx  = blueSupply({ market, args: { ...inputs, maxSharePrice,
     │                                               requirementSignature: sig } })
     │  200 { tx }                                             │
     │ ◄──────────────────────────────────────────────────    │
```

No Redis, no DB row, no `prepId` — the continuation lives in the client round-trip. If you'd rather not trust the client with the inputs, wrap them in an **opaque HMAC-signed token** you return from `/prepare` and validate at `/finalize`: still zero server storage, still any instance. Either way the signed `deadline`/`nonce` travel inside the `typedData`/signature, so the server re-derives the exact requirement from the returned inputs and verifies the signature against it.

### What the client carries (for a permit-based supply)

- **Business inputs:** `marketParams`, `amount`, `userAddress`, `nativeAmount` (the client had these to begin with — it initiated `/prepare`).
- **The signed values**, all inside `typedData` / `RequirementSignature.args`: `owner`, `asset`, `nonce`, `amount`, `deadline`, plus the requirement `type`. Carry the signed data forward **rather than re-running `getRequirements()`/sign on the second instance** — ERC-2612/Permit2 signature deadlines default to `now + 2h` and aren't caller-overridable, so re-preparing would produce a *different* deadline.
- **`maxSharePrice` / `minSharePrice`: not carried — recomputed at `finalize`.** They're unsigned on-chain slippage bounds, not part of any signature, so a fresh value is fine (and safer). Exported helpers: `computeMaxSupplySharePrice`, `computeMinBorrowSharePrice`, `computeMaxRepaySharePrice`, `computeMinWithdrawSharePrice`. (If you want byte-identical calldata to `prepare`, carry them too — your call.)

Path-specific notes:
- **Vault exact-asset `inKindRedeem`:** the vault-snapshot-derived share cap **is the signed permit value**, so it travels in the signature — nothing extra to carry. Just read `deadline` from the signed permit, don't recompute it.
- **Reallocations / shares-mode `repay` transfer amount:** RPC-derived at `prepare` and baked into calldata (the shares-mode transfer amount is in the permit when a permit is used; with a classic approval it isn't) — carry those verbatim rather than recomputing.

### On your three follow-ups

- **Permit2 nonce:** yes, nonce allocation/idempotency is intentionally the integrator's concern. It flows in as `permit2Nonce` and is fully self-contained in the signed `Permit2Args`, so it round-trips cleanly.
- **Mixed calls + multiple signature requirements:** confirmed — no SDK identity/order invariant. `buildTx` takes a `readonly RequirementSignature[]` and selects by type (`isPermitSignature` / `isAuthorizationSignature`); passing all matching signatures to the same prepared output in any order is sufficient. Assigning your own API-level requirement IDs is fine.
- **`buildTx(signed)` vs `buildTx([signed])`:** you're right — the Vault/Blue outputs take `readonly RequirementSignature[]`, so the examples should be `output.buildTx([signed])`. We'll fix the PR + migration-guide snippets.

So a JSON snapshot/rehydrate helper would be nice ergonomics, but it isn't required — the stateless pattern above works with the current surface. Happy to pin down the exact carry-forward field list per action if that helps you wire it up.

0xbulma (dismissed):

milanmorpho (dismissed):

0xbulma (dismissed):

milanmorpho (approved):

0xbulma:
@codex review the last commit

0xbulma (approved):

PR #1077, opened 2026-09-15
fix(morpho-sdk): make Midnight maker offer submission stateless

Foulks-Plb:
## Motivation

Midnight's ecrecover maker flow was the one place in `morpho-sdk` where an entity's `buildTx()` depended on cross-call in-memory state: `sign()` wrote the encoded offer-root payload into a `signedPayloads` Map closed over by the `ActionOutput`, and `buildSubmitOffersTx` looked it up by signature. This broke stateless usage — a requirement signed on one `MorphoMidnight` instance and submitted from another (prepare-on-A → finalize-on-B, or serialize-then-resume) threw `UnpreparedMidnightOfferRootSignatureError`. The SDK's design principle is stateless flows only.

## Solution

`buildSubmitOffersTx` now reads the payload straight off the signature it is handed (`signature.args.payload`, already populated by `sign()` and deep-frozen), so the `Map` is removed from all three maker flows, `getRatifierRequirements`, and the builder. `UnpreparedMidnightOfferRootSignatureError` is kept but now guards the "signature carries no payload" case, and the old `ignores untrusted payload bytes` test — which encoded the removed Map-as-trust-boundary behavior — is updated to assert the payload flows through (consistent with the already-public `mempoolSubmitOffers({ payload })` escape hatch). The stateless-`buildTx` rule is documented in root `AGENTS.md` §1 and `entities/AGENTS.md`, and a patch changeset is included. `tsc`, Biome, and the full `morpho-sdk` unit suite (1030 passing) are green.


Foulks-Plb:
@codex review
@devin review

Rubilmax (approved):

PR #1088, opened 2026-09-16
feat(morpho-ts): register Arc bundles addresses

Foulks-Plb:
## Motivation

The Morpho bundle periphery (VaultExitBundlesV1, VaultBundlesV1, BlueBundlesV1) was deployed on Arc (chain 5042) in [morpho-org/deployments#256](https://github.com/morpho-org/deployments/pull/256), but the SDK address registry did not yet expose them, so `getChainAddress(ChainId.ArcMainnet, "bundles.*")` failed on Arc while resolving on every other supported chain.

## Solution

Register the three bundle addresses in the Arc `bundles` group of `ChainAddresses` (byte-for-byte from `address-book.json`) and record the `VaultExitBundlesV1` creation block (`20_322_300`) in the deployment registry, matching the layout used by the other chains. The block was derived via `eth_getCode` binary-search against the Arc RPC since the address book carries no block numbers. Arc is added to the existing `VaultExitBundlesV1` / `VaultBundlesV1` / `BlueBundlesV1` `test.each` blocks, and a minor changeset covers `morpho-ts`/`blue-sdk`/`morpho-sdk` (patch for `evm-simulation`). Midnight-stack addresses from the same deployment PR are intentionally out of scope.


Foulks-Plb:
@devin verify addresses

Foulks-Plb:
@claude verify from https://github.com/morpho-org/deployments/pull/256 and on-chain.

jinmel (approved):

PR #1098, opened 2026-09-17
fix(release): reject tar entry aliases of package.json (SDK-672, SDK-673)

Foulks-Plb:
## Summary

Fixes [SDK-672](https://linear.app/morpho-labs/issue/SDK-672) and [SDK-673](https://linear.app/morpho-labs/issue/SDK-673) (Apex HIGH, release pipeline).

The privileged publish job in `.github/workflows/publish.yml` validates tarball entries by their *stored* names (`grep` for `package/` prefix, `.`/`..`/`//`, exactly one literal `package/package.json`) and then reads only the literal `package/package.json`. npm's extractor (node-tar 7.x, bundled with npm 11) does not honour those stored names: it strips a leading `\` as a root indicator on every platform, treats `\` as a separator on Windows, and case-insensitive consumer file systems (Windows, macOS default) collapse `package/Package.json` onto `package/package.json`. Both `package/\./package.json` and `package/Package.json` pass the current checks, so a later entry could overwrite the validated manifest at install time while the workflow inspects the benign one.

New tested gate `scripts/ci/verify-tarball-entries.ts` (per `AGENTS.md` §10, decision logic lives in TS, not inline shell), run before the existing checks for every tarball:

```
node scripts/ci/verify-tarball-entries.ts "$tgz"
```

**Input comes from the raw archive, not from GNU `tar -t`.** Review found GNU tar's listing and node-tar's extraction disagree on crafted archives (global PAX `path` records, pax-`path` + GNU `L` longname precedence, lone zero blocks GNU stops at but node-tar reads through). `readTarballEntries(tgz)` gunzips and walks the 512-byte ustar headers with node-tar's rules — ustar `prefix`, one per-entry `x` header whose `path` may override the name — and fails closed on everything else:

- typeflags other than `0`/NUL/`5`/`x` (so `g`, `L`, `K`, links, devices are rejected);
- magic/version bytes other than exactly `ustar\0` + `00` (node-tar only honours `prefix` for that value), an invalid header checksum (node-tar's unsigned sum; a bad header is skipped and re-synced by node-tar), a directory header declaring a non-zero `size` (node-tar forces it to 0), an empty path or a non-empty linkname (node-tar skips such headers one block at a time without consuming the body), non-octal `size`, truncated archives, dangling or consecutive `x` headers, an `x` header larger than node-tar's `maxMetaEntrySize` (1 MiB; node-tar ignores it together with its `path` record). Header strings are decoded with node-tar's exact `decString` (`.replace(/\0.*/, "")`), so bytes hidden behind NUL + newline are seen and rejected rather than truncated away;
- PAX records outside `BENIGN_PAX_KEYS` (notably `size`, which would desynchronise parsers), and any PAX record containing a newline (node-tar's `parseKV` splits the body on `\n` rather than framing by length, so a `comment` value can smuggle a `path=` line past a length-framed parser);
- a zero block followed by any non-zero data, and an archive with no end-of-archive zero block at all;
- a regular file whose name ends in `/`.

`verifyTarballEntries(entries)` then rejects (fail-closed):
- any entry containing `\`; `.` / `..` / empty segments; a first segment other than `package`;
- any segment ending in `.` or a space (Win32 trims), longer than 255 chars, any entry path of 1024+ chars (macOS `PATH_MAX`), containing `~` (8.3 alias), a Win32-invalid character (`<>:"|?*`), or a DOS device basename (`CON`, `NUL`, `COM1`, …);
- any character outside printable ASCII (rather than approximating each FS's Unicode folding);
- any two entries whose `canonicalEntryPath` (ASCII lower-case + no trailing `/`) collide, and any regular file that is an ancestor of another entry in any order — including a `.gitignore` under the `.npmignore` name pacote renames it to on install (only the literally spelled sibling `.npmignore` file is tolerated, in either order);
- a manifest aliasing `package/package.json` without being stored literally, or no manifest at all (a duplicate is caught by the collision check).

The existing inline GNU-listing checks are kept unchanged (defence in depth); the gate runs first so they only ever see names it approved. Verified against real `pnpm pack` output (passes) and against crafted raw-header archives in the colocated tests (48 tests), including the global-PAX, `L`-longname and lone-zero-block layouts from review. Workflow-only change, no package behaviour affected, so no changeset.

Link to Devin session: https://app.devin.ai/sessions/dc1c8b3e20304d43a362ab7164d31e79
Open in Devin Desktop: https://app.devin.ai/desktop/session/dc1c8b3e20304d43a362ab7164d31e79?variant=devin
Requested by: @Foulks-Plb

---

<a href="https://app.devin.ai/review/morpho-org/sdks/pull/1098" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


.github/workflows/publish.yml
```diff
@@ -256,7 +256,26 @@ jobs:
           declare -A received_tarball
           shopt -s nullglob
           for tgz in "${RUNNER_TEMP}/tarballs/"*.tgz; do
-            entries="$(tar -tzf "$tgz")"
+            # npm's extractor (node-tar) does not extract the stored names the
```
jinmel: remove comments

